PR #3, opened 2020-02-29
WIP: various changes

Jutho:
Hi Maarten,

very much work in progress, my first attempt at various changes, mostly in FiniteMPS code. This currently breaks DMRG etc and is anyway very much work in progress. But the commit log should provide a rather complete history of everything I have been doing and changing :-).

I will also need to rebase for your latest changes.


Jutho:
Oh great. But I am not sure if this was completely ready. I will anyway continue making updates :-)

maartenvd:
Ah, I was going to modify mpscomoving so that the orthonormalization behaved in the same way, and then get dmrg back working - I don't know how to accept only a subset of commits

Jutho:
That's perfect, it's probably good not too let these branches diverge too much.

PR #6, opened 2020-03-05
Finitemps view

maartenvd:
not really ready to merge - would this kind of thing work? So we store the original data in the state and never really touch that (optionally renaming setindex! to unsafe_setindex!), and then the 'view' objects manipulate the state and move the center?

(only findgroundstate in the examples works)


src/states/orthoview.jl
```diff
@@ -0,0 +1,73 @@
...
+
+    v.target[i] = vec
+end
+
+struct ALView{S}
+    target::S
+end
+
+function Base.getindex(v::ALView,i::Int)
+    i == length(v.target) && throw(ArgumentError("out of bounds"))
```
Jutho: `throw(BoundsError(v, i))`?

src/algorithms/groundstate/dmrg.jl
```diff
@@ -20,30 +20,25 @@ end
 
         for pos=1:(length(state)-1)
             (eigvals,vecs) =  let state = state,parameters = parameters
```
Jutho: This is a completely independent remark: I don't think the `let` construction has much effect on type stability here. If I recall correctly, when using arguments of the function (here `state` and `parameters`) in closures, their type is fixed and therefore the `let` block is not necessary. If there would be a reason to use a `let` block, it would be for the variable `pos` which is used in the closure. Ideally, of course, no `let` block is needed altogether.

maartenvd: I recall this being needed on an older version of julia - I haven't really tested it recently. I'll check it again because if we can get rid of those let blocks, the code would look much cleaner :)

maartenvd:
Ok great, then I will go ahead and fix all algorithms to make use of this new convention. Do we disable setindex! for finitemps and rename it to unsafe_setindex! ? Or do we still want that option? getindex is still useful because it is not inconceivable to want to access mps data without unnecessary qr steps.

Jutho:
> Do we disable setindex! for finitemps and rename it to unsafe_setindex! ? Or do we still want that option? getindex is still useful because it is not inconceivable to want to access mps data without unnecessary qr steps.

You know the code better than me; I think disabling `setindex!` is certainly good, given that this would otherwise invalidate the `centerpos` field. `getindex` could indeed be useful, if a user just wants to access the tensors without caring about their gauge. Or we could have an `AView` for that (i.e. `psi.A[n]`) which only supports `getindex` and no `setindex!`. Seems overkill, but `InfiniteMPS` also don't support directly indexing into the state. 

Unrelated remark: a more traditional name for `target` field in the view types would be `parent`.

Jutho:
Yes I see your point. Did you already take a look at TDVP (1 or 2 site); this might also bring along additional problems, that we should also take into account when trying to find a more robust solution.

As a temporary fix for now, maybe don't caring about the extra QR step is fine.

maartenvd:
I suppose most is solved when we switch back to some kind of gauge flag per tensor. Then only allow setindex! with additionally specified gauge flags (and nice ac/al/ar wrappers around that). Tdvp2 would then use the uglier setindex!, while a lot of other code can already use the nice wrappers.

The only thing that isn't really solved is the .C view which probably - unless we store more information in the state - cannot be written without superfluous ortho-steps.

Jutho:
Another solution is to change the semantics of what `setindex!` on `ALView` means. It could also mean, make the chain left canonical up to (but not including) site `i`, assign a new tensor to site `n` with the guarantee that it is also left canonical, and now modify `centerpos` such that it starts to the right of site `n`, i.e. something like
```julia
function Base.setindex!(v::ALView, vec, i::Int)
     i == length(v.target) && throw(ArgumentError("out of bounds"))
     leftorth!(v.target, i, normalize = false);
     v.target[i] = vec
     v.centerpos = max(i+1, first(v.centerpos)):max(i+1, last(v.centerpos))
 end
```
and similar for `AR`.
I think this solves the issue for `DMRG` but it's maybe a bit cheating. Although I think that if you are updating two sequential tensors in the MPS, and you want to use some nice interface for changing individual tensors for that, you will always have a somewhat finicky state in between.

maartenvd:
Ah yes, that works - but feels a bit cheaty

I was thinking about making the state object more involved; storing both tensors and bond tensors. getindex would be disabled and .A[] would return the "regular" array of tensor view.

If we then query AL, we can leftorth .A[] and store the resulting C on the bond. This would allow us to also write onesite tdvp without unnecessary qr steps.

The downside is that it is complicated and looks a bit overwhelming but the actual interface with .AL/AR/... would be easy to work with. We can disable get-set index and guarantee correctness of the gauges everywhere.

Jutho:
Such an idea has also crossed my mind. This then also allows to store an MPS in the Vidal format that people sometimes like. For certain aspects of spatial symmetries it is also beneficial to have explicit bond tensors; say if you want to explicitly parameterise trivial reflection invariance, that it is useful if `space(A,1) = space(A,3)` , and then you have some bond tensor in between with `space(C,1)=space(C,2) = dual(space(A,1))` which glues the tensors together. 

I see a lot of advantages to this, but this of course has the potential to break a lot of your code no? Would you use a Union{..,Nothing} or Union{..,Missing} for absent bond tensors?

However, I don't immediately see how this allows to circumvent your original DMRG concern?

maartenvd:
>I see a lot of advantages to this, but this of course has the potential to break a lot of your code no? 

I don't think the code breakage will be that big of a problem, as soon as we decide on a nice code interface then it shouldn't take more then a day to get everything working again.

I would use ** Union{..,Missing}** because that's a bit more consistent with the mpo hamiltonian case

It doesn't really adress the 2site dmrg problem, but maybe we can allow setindex!(ACView,(AL,C),index) and setindex!(ACView,(C,AR)) (or we introduce a new view?) Arguably a bit clunky but together with the bond-tensor storage this would completely solve the problem?

Jutho:
Sounds great; I am looking forward to your implementation.

maartenvd:
In case I'm going too much off track I have pushed my most recent commits. I still use your finitemps implementation but have written the orthoviews to include setindex!( (AL,C)) so - providing we write a different finitemps implementation - we can do without any unnecessary qr's.

I have not gone over everything but the tests (without the finitempo ones) are running. FiniteMPO is posing a new problem; either we duplicate the orthoview code, or we change the finitempo to have the layout (virtual - physical - physical' - virtual'). Maybe changing finitempo makes sense? I'm not sure.

PR #7, opened 2020-06-04
Adding Grassmann gradient optimisation

mhauru:


src/algorithms/groundstate/gradient_grassmann.jl
```diff
@@ -0,0 +1,165 @@
+"""
+GradientGrassmann is an optimisation methdod that keeps the MPS in left-canonical form, and
+treats the tensors as points on Grassmann manifolds. It then applies one of the well-known
+gradient optimisation methods, e.g. conjugate gradient, to the MPS, making use of the
+Riemannian manifold structure. A so called preconditioner is used, so that effectively the
```
Jutho, line 5: I would remove "so called"  :-)

src/algorithms/groundstate/gradient_grassmann.jl
```diff
@@ -0,0 +1,165 @@
...
+"""
+Take the L2 Tikhonov regularised inverse of a matrix `m`.
+
+The regularisation parameter is the larger of `delta` (the optional argument that defaults
+to zero) and square root of machine epsilon. The inverse is done using an SVD.
+"""
+function reginv(m, delta=zero(eltype(m)))
+    delta = max(delta, sqrt(eps(real(float(one(eltype(m)))))))
+    U, S, Vdg = tsvd(m)
+    Sinv = inv(real(sqrt(S^2 + delta^2*id(domain(S)))))
```
Jutho, line 145: I think `one(S)` is slightly easier than `id(domain(S))`, and also more future robust, e.g. if `S` at some point will be a specific `DiagonalTensorMap`, or if the storage of `S` would be a `CuArray`, then `one(S)` can easily inherit these properties.

src/algorithms/groundstate/gradient_grassmann.jl
```diff
@@ -0,0 +1,165 @@
...
+
+"""
+Precondition a given Grassmann tangent `g` at state `x` by the Hilbert space inner product.
+
+This requires inverting the right MPS transfer matrix. This is done using `reginv`, with a
+regularisation parameter that is the norm of the tangent `g`.
+"""
+function precondition(x, g)
+    (state, pars) = x
+    delta = min(real(one(eltype(state))), sqrt(inner(x, g, g)))
```
Jutho, line 158: This is probably ok, but just as a warning. Typically the preconditioner is (the inverse of) a linear operator acting on `g` (although there is indeed this paper on nonlinear preconditioners). Now, in CG, the preconditioner is only applied to `g` being the gradient at point `x`, and so `sqrt(inner(x, g, g))` could just as well be interpreted as a function of `x`, and indeed, the preconditioner can have arbitrary `x` dependence. So just to say that strictly speaking, this definition of `precondition` is not a linear function of `g`, but I assume it will not cause problems.

Jutho:
Look's great. I will probably also just register OptimKit in its current state and don't care about the absence of tests.

mhauru:
Thanks for all the feedback! I'll fix the small things you pointed out. If you are about to register OptimKit, then let's wait for that before merging. It'll allow to drop Manifest from version control, like it was before. I don't know if you talked about this with Maarten already, but we spent some time last night trying to get the docs to compile, and the conclusion was that trying to have an unregistered package as a dependency is a fragile mess (and was the only reason to include Manifest in the repo).

Jutho:
OptimKit was registered Friday night/Saturday morning, but with a new package, there is a 3 day waiting period. So on Tuesday morning OptimKit.jl should be a registered package.

src/algorithms/groundstate/gradient_grassmann.jl
mhauru: Yes, done.

src/algorithms/groundstate/gradient_grassmann.jl
mhauru: Done.

mhauru:
I just removed Manifest.toml, now that OptimKit is registered. The docs build, and assuming the tests will pass (they passed for me locally, the nightlies probably just fail because their package registry is a day out of date or something like that), then as far as I'm concerned this is ready to merge.

PR #8, opened 2020-06-10
Improve the design of the GradientGrassmann struct and its constructor

mhauru:
As discussed. Tests pass locally.


maartenvd:
I really like this design :)

PR #11, opened 2020-08-06
Make dependency versions less restrictive

Jutho:
Somewhat annoying, but because of how SEMVER works, there can be breaking changes between 0.x and 0.y versions (not between 1.x and 1.y), so that `KrylovKit = "0.4"` does not allow to use `v"0.5"`, leading to dependency conflicts. I've also added a few non-existing (yet) versions of my packages, to be somewhat future proof, including a version 1, which means this should be good to go up to version 2 (whenever that will ever happen).

Another food for thought: if you want less dependencies, you could remove OptimKit and TensorManifolds, and only make this ground state method available if these packages are separately loaded using the Require.jl mechanism. Not sure if that is worth the trouble though.


mhauru:
If you want more flexibility in how many <1 versions will precede 1.0, you could also consider using the hypen specifiers: https://julialang.github.io/Pkg.jl/v1/compatibility/#Compatibility-1

Requires julia >= 1.4 though.

Jutho:
Yes many possibilities, the only thing important for releasing Julia packages is that all dependencies have some upper bound in the version, so things like ">=0.4" are no good.

maartenvd:
There is even this caret/tilde specifier, which I guess is even nicer. I don't particularly mind depending on optimkit and tensorkitmanifolds

Jutho:
Yes, but if I read correctly, not specifying any prefix is equivalent to the caret specification. So I think the only alternative would be an inequality specification, e.g. "<2"

mhauru:
Or e.g. `0.4 - 1`, which should match everything starting from 0.4.0 up to but *not* including 2.0.0.

Jutho:
Yes, I guess that's nice. I will change to that format.

PR #13, opened 2020-09-09
Added a function that returns sigma and pi order parameters for CGN 

Gertian:


Jutho:
Such things should not really go into the main source code right?

They do not contribute to the general purpose of an MPS library. Or maybe the different models in `src/models/*.jl` could go into a submodule `Models` (for now), such that they only come into scope after doing `using MPSKit.Models`.

Gertian:
Yes that would be cleaner.
Would be handy to accept the pull request for now and then move the models folder to a new module.

maartenvd:
Indeed, we can accept this now. But also register a packet with all kinds of mpskit models. Once that is registered, we can remove the entire models folder from mpskit

Jutho:
Yes that would perhaps be useful. I still have an initial attempt of a `ManyBodyKit` package somewhere, though the purpose of that was to facilitate defining your own lattice models (both quantum and stat mech) using a new DSL

Jutho:
So an MPSKitModels package sounds like a good idea.

mhauru:
Why MPSKitModels and not MPSKit.Models? I don't see where else MPSKitModels would get used, other than as a dependency of MPSKit. Unless we would make it significantly more general, provide both local term representations and MPOs for each model, etc., in which case it should probably be called something else too.

Jutho:
Well, MPSKitModels could be a separate repository, and would not be a dependency of MPSKit (but the other way around). So you can use MPSKit without needing the MPSKitModels package. If you just make it a submodule, it is still loaded whenever you do `use MPSKit`; it's just no longer exported by default into global scope, but it still contributes to the load time.

mhauru:
The effect on load time should be pretty negligible, no? I was thinking the main effect would be code cleanliness/complexity and ease of use. You could argue it's more sanitary to have MPSKit only care about the actual algorithms and not what you apply them to, but ease of use takes a hit when somebody wants to test drive MPSKit on a simple example and the answer starts with "well first you need to install this other package...".

Another solution would be having a second module in the same repo. This works nicely as long as you assume people clone your repo, and is what I'm doing with MERA at the moment, where there's the MERA.jl module in `src`, and then another folder called `demo` that includes usage examples and things you need for them, like model Hamiltonians. Not sure how this is going to work once I register the package though.

mhauru:
Maarten currently has the `examples` folder. Would that then be in MPSKitModels? Or would it be some kind of subpackage that has MPSKitModels as a dependency?

maartenvd:
examples would still be in mpskit, and would use mpskitmodels.

Now I have to artificially add new tests whenever I add a new model (otherwise coverage takes a hit). To do it properly I would have to test both the algorithms, and then also all the exported hamiltonians. I think it'd be cleaner to seperate this out, and I can then become really sloppy adding new models to this mpskitmodels package without having to worry about the mpskit coverage.

It's also what the machine learning community appears to do, where they have a modelzoo package, and the infrastructure is seperated out in different packages

Jutho:
But examples and demos are typically scripts, they should not really be modules right? Many people actually use notebooks for this, though Pluto.jl offerst the perspective that it could at the same time be a notebook and a plain source code text file

mhauru:
For coverage, I think you can ignore specific files or folders if you want: https://docs.codecov.io/docs/ignoring-paths

mhauru:
> But examples and demos are typically scripts, they should not really be modules right? Many people actually use notebooks for this, though Pluto.jl offerst the perspective that it could at the same time be a notebook and a plain source code text file

Instead of "module", I should probably have said "package". A separate folder, outside of `src`, that has the example scripts and maybe modules like the stuff in MPSKitModels that the scripts need that are not part of the main package. Could maybe have it's own Project.toml too.

Jutho:
> Instead of "module", I should probably have said "package". A separate folder, outside of src, that has the example scripts and maybe modules like the stuff in MPSKitModels that the scripts need that are not part of the main package. Could maybe have it's own Project.toml too.

Did you then not intend to say just a Julia project? From the manual:
"A package is a project with a name, uuid and version entry in the Project.toml file, and a src/PackageName.jl file that defines the module PackageName. This file is executed when the package is loaded."

maartenvd:
If we had some package agnostic format usable by itensors,montecarlo,mpskit, then we would also have models in a seperate package, so I have no problem with at least removing them from mpskit

mhauru:
Well I wasn't thinking that there would be a `demo/src` or `demo/src/Demo.jl`. But whatever the details of project/package/module, the point is that it would be in the same git repo, but outside the main package, i.e. not in `src`. Something that is intimately tied to the package MPSKit, but not part of the module MPSKit that one imports, kinda like `docs` or `test`. In contrast, I understood that you were proposing an entirely separate package and git repo for MPSKitModels.

maartenvd:
well maybe you're right,  it makes more sense to bundle it together with mpskit in a seperate module, within the same repository

mhauru:
> If we had some package agnostic format usable by itensors,montecarlo,mpskit, then we would also have models in a seperate package, so I have no problem with at least removing them from mpskit

If it would be usable by things other than MPSKit, and not super specific to conventions and design choices of it, then I agree, it should definitely be entirely separate, because there indeed is a clear separation between the algorithms and the models.

Jutho:
> In contrast, I understood that you were proposing an entirely separate package and git repo for MPSKitModels.

Yes indeed, and I don't fully understand the argument against that? The name MPSKitModels is still reasonably short yet sufficiently specific not to clutter the package namespace. And installing a Julia package is not a big deal, right? it's just entering one line in the REPL. In fact, I think if you've installed MPSKit, it's way easier to also install MPSKitModels, then to actually go and find the examples folder in whatever path Julia decided to install MPSKit in the first place. So after installing MPSKit as a regular user, you would still end up to go and find the examples on the online git repository, as that's easier then finding it on your local hard drive.

At least that's my impression? I do this all the time for packages which are not my own. I don't bother to go and try to find the source code or git repository, as Julia gave it some crazy unrecognisable name? 

Furthermore, it seems to be what several other Julia ecosystems are doing no?

An alternative is indeed a more general package (like ManyBodyKit) that encodes Hamiltonians in some background agnostic way to be used by several other packages. And then there would be some glue code (using e.g. Requires.jl) in MPSKit that, if you also load ManyBodyKit, there are functions for converting those generic Hamiltonian's into e.g. the MPO format of MPSKit. And ITensor could have something similar, and Yao could have something similar, and so forth.

Jutho:
Though actually ManyBodyKit.jl the way I started building it (only exists on my harddrive, or actually dropbox, not on github) was also kind of built on TensorKit.jl to define local Hilbert spaces etc.

mhauru:
>And installing a Julia package is not a big deal, right?
Well, yes and no. It's not a big deal if you know what you are doing. If you are new user, possibly not only to MPSKit but to Julia too, then every additional non-obvious step that you need to take to actually see DMRG or VUMPS crunching numbers can be an obstacle and a frustration. If MPSKit itself comes with examples, and these examples use MPSKitModels, then I think at the very least it should then be a dependency that's installed automatically. And if it's a dependency, and that's it's only purpose (it has no other use case other than running examples for MPSKit), then why not bundle it in a single repo?

The point you raise about the trouble of finding the examples that come with MPSKit is a good one, and something I haven't found a good solution to yet. I'll have to figure this out for MERA.jl too. I would really like a new user to just be able to `]add MPSKit` and have everything they need at hand, but I don't know if there's a good way of achieving that without bloating the main module with stuff related to models and examples, which I agree is not ideal.

maartenvd:
It is true that every non-obvious step can be an obstacle, and I should really write a small quickstart guide in readme.md which can then also describe the ]add MPSKitModels step.

I have no strong opinion one way or the other, but slightly prefer a seperate git repo entirely, because I like the idea of small modular packages.

mhauru:
Sure, I'm not terribly opposed to that. I find it not to be ideal, but it may be optimal, in that what I was proposing also has downsides to it. The `examples` from MPSKit would then be moved to MPSKitModels?

maartenvd:
I would keep examples in mpskit, they also illustrate how the mpskit algorithms work. As it stands, they often depend anyway on other packages (or maybe I got rid of it, but some examples really need to use plots to show what is happening). I would also change the readme to explain exactly how everything works

Jutho:
Note the specific comment of mine that started this discussion also was aimed at the Gross Neveu model, and then in particular at helper functions for observables in the chiral gross neveu model.

The idiomatic models like Ising, XXZ, Heisenberg , Hubbard, could certainly live in MPSKit.jl for now, though just as well as in MPSKitModels.jl, where then more models could easily be added. The first three are all part of one family

```H = (1-gamma)/2 X X + (1+gamma)/2  Y Y + Delta ZZ + hx X + hz Z```

(I am reluctant to add `hy Y` as this makes the Hamiltonian complex). 

though imposing symmetries will constrain the parameters.

PR #14, opened 2020-09-09
TM-exctations

Gertian:
Code that calculates the leading eigenvalues of the TM

### DIFF
--- a/src/algorithms/excitation/TM_excitations.jl
+++ b/src/algorithms/excitation/TM_excitations.jl
@@ -1,0 +1,26 @@
+function correlation_length(state::InfiniteMPS; otherstate=state, tol = 1e-8, exi_space = ComplexSpace(1), tol_angle = 0.1, triv=true)
+    ALs1 = state.AL
+    ALs2 = otherstate.AL
+
+   #the first eigenvalue tells us what angle to search at for the leading nontrivial eigenvalue
+    init = TensorMap(rand, ComplexF64, space(ALs2[1],1)*exi_space,space(ALs1[1],1))
+
+    eigenvals, _ = eigsolve(init, 20, :LM, tol=tol) do x
+		return transfer_left(x, ALs1, ALs2)
+    end
+
+	triv && (eigenvals = eigenvals[2:end])
+
+    best_angle = mod1(angle(eigenvals[1]), 2*pi)
+    ind_at_angle = findall(x->x<0.1 || abs(x-2*pi)<0.1, abs.(mod1.(angle.(eigenvals), 2*pi).-best_angle))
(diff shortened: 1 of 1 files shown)


src/algorithms/excitation/TM_excitations.jl
```diff
@@ -0,0 +1,26 @@
+function correlation_length(state::InfiniteMPS; otherstate=state, tol = 1e-8, exi_space = ComplexSpace(1), tol_angle = 0.1, triv=true)
+    ALs1 = state.AL
+    ALs2 = otherstate.AL
+
+   #the first eigenvalue tells us what angle to search at for the leading nontrivial eigenvalue
+    init = TensorMap(rand, ComplexF64, space(ALs2[1],1)*exi_space,space(ALs1[1],1))
+
+    eigenvals, _ = eigsolve(init, 20, :LM, tol=tol) do x
```
maartenvd, line 8: this can be problematic, when the transfer matrix has dimension < 20. I would also maybe move the 20 to the keyword arguments?

Gertian: good point. This is a bit arbitrary we just need to ensure that the second singular value in the requested symmetry sector is in there.

src/algorithms/excitation/TM_excitations.jl
```diff
@@ -0,0 +1,26 @@
+function correlation_length(state::InfiniteMPS; otherstate=state, tol = 1e-8, exi_space = ComplexSpace(1), tol_angle = 0.1, triv=true, max_sinvals = 20)
```
Jutho, line 1: So why is this argument called `max_sinvals`? It's not singular values right? It's eigenvalues of the transfer matrix. Maybe more generic name like `howmany` would be more appropriate? 

Instead of `exi_space`, I would rather specify a sector (and call the argument `sector::Sector = Trivial()`). The corresponding exi_space to be used is then `exi_space = ℂ[sector=>1]`. This also works for `sector = Trivial()` and you can even use the trivial case `exi_space = ℂ^1` when the spacetype is not `ComplexSpace`, by replacing `space(ALs2[1],1)*exi_space` with `space(ALs2[1],1) ⊠ exi_space` (the deligne tensor product at work). 

The reason not to directly give `exi_space` is that, if e.g. a user gives  `exi_space =  ℂ^2`, this will just duplicate every eigenvalue, cause you have a transfer matrix acting on index 1 and 3 of your eigenvector, and and identity map acting on the second index (the `exi_space` index), so essentially you are computing the eigenvalues of transfer matrix kronecker product with an identity matrix.

maartenvd: this would also be better in the quasiparticle excitation code

PR #33, opened 2021-11-09
Sparsempo

maartenvd:


maartenvd:
I wanted operators to be more consistent. There are 2 base types; densempo and sparsempo. Code should preferably not care what kind of mpo it is, hopefully allowing one to use densempo to represent a hamiltonian.

SparseMPO can then be used as the storagetype for mpohamiltonian, and can represent time evolution mpos. By fixing some convention, we can now easily use time evolution mpos for finite mps (in the past you had to use densempo, and then it wasn't clear what leg was supposed to live at the edge).

I tried to merge some environment caches together, and made the infinite mpo environment not care if it used dense or sparse mpos, so we can use sparse mpos for infinite time evolution as well.

Sparsempo and sparsemposlice are both abstractarrays

PR #34, opened 2022-01-18
Typocalypse

maartenvd:
This can serve as a discussion place? Changes I remember making:

- Transfer matrix is an operator now. The one site operator exists, as does a sequence of those one-site operators (a different type). This transfer matrix can also be regularized (E -> E - |L> < R| ), which is yet another type.

- There are two new operators (operators in the sense of an mpo or a hamiltonian; things that can act on an MPS). There is a projection operator, which is really only used behind the scenes in code for finding excited states with DMRG. There is also the "LazyLinco" operator, which represents the linear combination of different operators. I don't overload sum, as the sum of 2 hamiltonians should really be another hamiltonian, and not a weird object that represents the sum in a lazy way. LazyLinco is also really only used behind the scenes and therefore both new operators are really "stub" implementations.

- The effective zero,one,two-site derivatives are all linear operators. DMRG/VUMPS call AC_eff(position,state,operator,environments), and this returns a linear operator that they in turn pas to eigsolve. 

- I did not introduce a new type for the sparsempo-environment vector. I started writing that, but the implementation felt a bit too forced.

There will probably be further changes, but these are all the ones that concern new types


src/transfermatrix/transfermatrix.jl
```diff
@@ -0,0 +1,69 @@
...
+
+# constructors
+TransferMatrix(a) = TransferMatrix(a,nothing,a);
+TransferMatrix(a,b) = TransferMatrix(a,nothing,b);
+TransferMatrix(a,b,c) = TransferMatrix(a,b,c,false);
+TransferMatrix(a::AbstractTensorMap,b,c::AbstractTensorMap,isflipped) = SingleTransferMatrix(a,b,c,isflipped);
+function TransferMatrix(a::AbstractVector,b,c::AbstractVector,isflipped)
+    tot = ProductTransferMatrix(TransferMatrix.(a,b,c));
+    isflipped ? flip(tot) : tot
+end
```
Jutho, line 51: There seems to be some inconsistency over when `isflipped` is an extra required argument, and when it is not an argument at all. Maybe make it an argument with default value or a keyword argument for all constructors?

src/transfermatrix/transfermatrix.jl
```diff
@@ -0,0 +1,69 @@
...
+(d::SingleTransferMatrix)(vec) = d.isflipped  ? transfer_left(vec,d.middle,d.above,d.below) : transfer_right(vec,d.middle,d.above,d.below);
+(d::RegTransferMatrix)(vec) = regularize!(d.tm*vec,d.lvec,d.rvec);
+
+
+# constructors
+TransferMatrix(a) = TransferMatrix(a,nothing,a);
+TransferMatrix(a,b) = TransferMatrix(a,nothing,b);
+TransferMatrix(a,b,c) = TransferMatrix(a,b,c,false);
+TransferMatrix(a::AbstractTensorMap,b,c::AbstractTensorMap,isflipped) = SingleTransferMatrix(a,b,c,isflipped);
+function TransferMatrix(a::AbstractVector,b,c::AbstractVector,isflipped)
```
Jutho, line 48: `ProductTransferMatrix` only accepts `Vector`; I assume the conversion will be implicit or a consequence of broadcasting. I keep wondering whether something like `StaticVector` should be allowed, which does encode length information (and probably preserves it in broadcasting). But for now the choice seems to be only `Vector`, so I would also enforce this here, unless there was some other reason.

maartenvd: this would be tricky, as mpskit often deals with periodicarrays, for which transfermatrix would then fail. It currently works indeed because of the broadcasting, which turns it into a vector.

I can also make producttransfermatrix accept any AbstractVector subtype?

PR #42, opened 2022-08-23
Diskarray

maartenvd:
stabilize storagetype and slightly lower memory usage. Some parts are rather ugly, maybe we should add some convenience functions to tensorkit?


maartenvd:
@Jutho perhaps it would be nice to have something like similar_from_storagetype? Currently I often have to take one instance with this storagetype (somewhat at random), and then call similar(tensor,new_domain,new_codomain). 

I also need the ability to create both zero/random tensors, with the same storagetype as some given tensors. Currently I do this using logic defined in utility/utility.jl:96 , but maybe something like this belongs in tensorkit?

Lastly, should mpskit even preserve the storagetype? Given a state living on the gpu, it may not be advisable to also store all environment tensors there... I've done this now to make it work with DiskArrays, but this may have been the wrong call...

PR #45, opened 2022-09-22
fix <a|mpo|b> for InfiniteMPS

tangwei94:
fix the function `dot` for <a|mpo|b>, where |a> |b> are `InfiniteMPS`s


maartenvd:
thanks!

PR #55, opened 2023-06-21
WindowMPS

DaanMaertens:
Renaming "MPScomoving" to "WindowMPS" and adding a "Window" struct to carry operators/environments for windows. This way WindowMPS could be an instance of Window, but this does not allow WindowMPS to be a subtype of AbstractFiniteMPS (easily), so this was not done for now.


maartensandbox:
I like the renaming, because MPSComoving is a bad name. 

I don't understand the idea behind "window". If it's just as a structure that operators/environments for windowmps can hold, then I don't think this will save you much and might even complicate things slightly.

Another change I was hoping to once get around to is to make the windowMPS movable, so that it does not have to describe the tensors at locations 1:length(window). And yet another change that would be useful is the possibility of having a utility leg dangling from the windowmps. We should definitely just start with the rename, but maybe in future those options could be nice to add

PR #77, opened 2023-10-09
Add dynamical tolerance scaling for `VUMPS`

leburgel:
Adds some optional dynamical tolerance scaling to `VUMPS` which sets the tolerance of the local eigensolver updates, environment recalculation and infinite MPS gauging based on the current galerkin error. I've enabled this by default here since I don't see any clear downsides to having this. We could also keep the default behavior as is and have users switch on tolerance scaling manually. Is this preferable?

In addition, I have set `eager=true` in the default eigenvalue solver to check for early convergence. It seems to me that this is really what you would want to do since basically all eigensolver subroutines get the result of the previous iteration as an initial guess, and this is generally pretty good. For my current use cases (variational PEPS optimization) this gives a speedup by over a factor of 2. If we don't want to have this eager behavior by default, it would at the very least be good to have better access to the actual eigensolver settings when initializing a `VUMPS` instance.

### DIFF
--- a/src/environments/mpohaminfenv.jl
+++ b/src/environments/mpohaminfenv.jl
@@ -53,19 +53,22 @@ function rightenv(envs::MPOHamInfEnv, pos::Int, state)
     return envs.rw[:, pos]
 end
 
-function recalculate!(envs::MPOHamInfEnv, nstate)
+function recalculate!(envs::MPOHamInfEnv, nstate; tol=envs.solver.tol)
     sameDspace = reduce(&, _lastspace.(envs.lw[1, :]) .== _firstspace.(nstate.CR))
 
     if !sameDspace
         (envs.lw, envs.rw) = gen_lw_rw(nstate, envs.opp)
     end
 
+    solver = envs.solver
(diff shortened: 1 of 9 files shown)


src/environments/mpohaminfenv.jl
```diff
@@ -53,19 +53,22 @@ function rightenv(envs::MPOHamInfEnv, pos::Int, state)
...
 
-function recalculate!(envs::MPOHamInfEnv, nstate)
+function recalculate!(envs::MPOHamInfEnv, nstate; tol=envs.solver.tol)
     sameDspace = reduce(&, _lastspace.(envs.lw[1, :]) .== _firstspace.(nstate.CR))
 
     if !sameDspace
         (envs.lw, envs.rw) = gen_lw_rw(nstate, envs.opp)
     end
 
+    solver = envs.solver
```
lkdvos, line 63: Can you only create a new solver if tolerance changed?
```julia
solver = solver.tol == tol ? solver : @set solver.tol = tol
```

lkdvos (approved):

PR #78, opened 2023-10-09
Normalize in multiline galerkin

leburgel:
Normalizes the `AC´` tensor after application of the effective Hamiltonian before using it to compute the galerkin. This seems more fair for stat mech problems with very large leading boundary eigenvalues, the current implementation was giving me quite a bit of trouble.

### DIFF
--- a/src/algorithms/toolbox.jl
+++ b/src/algorithms/toolbox.jl
@@ -25,14 +25,13 @@ calc_galerkin(state::Union{InfiniteMPS,FiniteMPS,WindowMPS}, envs)::Float64 =
 function calc_galerkin(state::MPSMultiline, envs::PerMPOInfEnv)::Float64
     above = isnothing(envs.above) ? state : envs.above
 
-    return maximum(
-        [
-            norm(
-                leftnull(state.AC[row + 1, col])' *
-                (∂∂AC(row, col, state, envs.opp, envs) * above.AC[row, col]),
-            ) for (row, col) in product(1:size(state, 1), 1:size(state, 2))
-        ][:],
-    )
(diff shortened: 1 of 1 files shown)


lkdvos:
We should probably do the same for the other cases, while less likely, it is still possible to have a pathological Hamiltonian with energy close to zero, or very large energies.

Small note, in principle the left nullspace of AC is given by (1 - AL AL*) which is slightly faster. (See above). It might be a good idea to get all of these cases to be the same

Jutho:
Doesn't `normalize!(AC´)` work instead of `AC´/norm(AC´)`? The latter gives one extra allocation. The former should work (and amount to `scale!(AC´,1/norm(AC´))`). If it doesn't, that's a bug.

src/algorithms/toolbox.jl
```diff
@@ -14,10 +14,15 @@ function infinite_temperature(ham::MPOHamiltonian)
...
+    calc_galerkin(state, envs)
+
+Calculate the galerkin error.
+"""
 function calc_galerkin(state::Union{InfiniteMPS,FiniteMPS,WindowMPS}, loc, envs)::Float64
-    out = ∂∂AC(loc, state, envs.opp, envs) * state.AC[loc]
-    out -= state.AL[loc] * state.AL[loc]' * out
+    AC´ = ∂∂AC(loc, state, envs.opp, envs) * state.AC[loc]
+    normalize!(AC´)
+    out = add!(AC´, state.AL[loc] * state.AL[loc]' * AC´, -1)
```
Jutho, line 25: One less allocation could have been obtained by writing
`out = mul!(AC´, state.AL[loc], state.AL[loc]' * AC´, -1, +1)`

PR #91, opened 2023-11-29
LazySum: Functionality to perform TDVP and find_groundstate with a lazy sum

DaanMaertens:
Adds a new struct `LazySum` for a lazy sum of operators, together with a `MultipleEnvironments` cache for the corresponding environments. This lazy sum only gets evaluated when it is needed e.g. in expectation_value
 or when the action of derivative operator is calculated. Thanks to support for the latter `LazySum` works out of the box for VUMPS, GradientGrassmann or TDVP algorithms. For iDMRG some minor changes were needed in order to make it work with the iDMRG environments.


DaanMaertens:
With the addition of time dependent operators in mind for the future, some support to differentiate between time dependent and time-independent operators with Holy traits, is already added. I'm not sure my implementation is satisfactory and if this system should be extended to derivative operators or leave it as is.

src/operators/lazysum.jl
```diff
@@ -0,0 +1,76 @@
...
+#Base.convert(::Type{<:LazySum},x::O) where {O} = LazySum(x)
+#Base.convert(::Type{T}, x::T) where {T<:LazySum} = x
+
+#Base.:+(op1::LazySum, op2) = +(promote(op1,op2)...)
+
+Base.:+(op1::LazySum,op2) = op1 + LazySum(op2)
+Base.:+(op1,op2::LazySum) = LazySum(op1) + op2
+# removes ambiguity for +(A::AbstractArray, B::AbstractArray)
+Base.:+(op1::LazySum, op2::AbstractArray) = op1 + LazySum(op2)
+Base.:+(op1::AbstractArray, op2::LazySum) = LazySum(op1) + op2
```
lkdvos, line 74: Do we want to allow users to add `AbstractArray`s? It might be more robust to keep it a little verbose, and ask them to explicitly call `sum1 + LazySum(vecofoperators)` which I think is fair because `LazySum` is exported and not too long anyways

src/operators/lazysum.jl
```diff
@@ -0,0 +1,76 @@
...
+
+Base.repeat(x::LazySum, args...) = LazySum(repeat.(x,args...))
```
lkdvos, line 76: I am not sure, but I think if you implement the full `AbstractArray` interface this might follow automatically.
https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array
In particular, I guess this needs `similar` to be defined, but this might need some checking

src/utility/lazysum.jl
lkdvos, line 1: I think this file is obsolete and not included?

src/operators/lazysum.jl
```diff
@@ -0,0 +1,76 @@
...
+eval_at(::TimeDependent,x::LazySum,t::Number) = LazySum(O -> _eval_at(O,t),x) 
+eval_at(::TimeDependent,x::LazySum) = throw(ArgumentError("attempting to evaluate time-dependent LazySum without specifiying a time"))
+eval_at(::NotTimeDependent,x::LazySum) = sum(O -> _eval_at(O),x)
+eval_at(::NotTimeDependent,x::LazySum,t::Number) = throw(ArgumentError("attempting to evaluate time-independent LazySum at time"))
+
+# For users
+(x::LazySum)() = eval_at(x)
+ConvertOperator(x::LazySum) = eval_at(x) # using ConvertOperator should do explicit multiplication
+(x::LazySum)(t::Number) = eval_at(x,t) # using (t) should return NotTimeDependent LazySum
+ConvertOperator(x::LazySum,t) = x(t)()
```
lkdvos, line 57: I am not a big fan of having `lazysum()` automatically sum up the elements. This does not really feel like intuitive syntax. I also think the capitalized `ConvertOperator` is misleading, this seems to suggest that it is a special type you are constructing. Is it an option to just replace this with `Base.sum`?

src/MPSKit.jl
```diff
@@ -27,7 +27,8 @@ export entanglementplot, transferplot
 
 # hamiltonian things
 export Cache
-export SparseMPO, MPOHamiltonian, DenseMPO, MPOMultiline
+export SparseMPO,
+    MPOHamiltonian, DenseMPO, MPOMultiline, LazySum, ConvertOperator
```
lkdvos, line 31: See remark about `ConvertOperator` in lazysum.jl

lkdvos (changes requested):
Great PR, looks clean! Can we resolve the minor comments and then merge?

src/algorithms/groundstate/idmrg.jl
```diff
@@ -47,8 +46,7 @@ function find_groundstate(ost::InfiniteMPS, ham, alg::IDMRG1, oenvs=environments
             Ψ.CR[pos - 1], temp = rightorth(_transpose_tail(vecs[1]))
             Ψ.AR[pos] = _transpose_front(temp)
 
-            tm = TransferMatrix(Ψ.AR[pos], ham[pos], Ψ.AR[pos])
-            setrightenv!(envs, pos - 1, tm * rightenv(envs, pos))
+            update_rightenv!(envs, Ψ, ham, pos)
```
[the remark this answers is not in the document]
DaanMaertens: yep that was a mistake. Fixed with eca2cd03e976fedce7ce35dfddd24c1eee404352.I also changed update_leftenv so the pos in the arguments is always the position where you set the env. The code in imdrg should have been changed accordingly.

src/operators/lazysum.jl
DaanMaertens: I agree. Removed with 1c737774023c400f75659481a26d9109e1f261e8.

src/operators/lazysum.jl
DaanMaertens: I don't thinks this can be done this way because the repeat for a `LazySum{<:MPOHamiltonian}` for example needs to repeated the elements themselves, not the `Vector{<:MPOHamiltonian}` inside LazySum.

PR #104, opened 2023-12-13
MultipliedOperator: Functionality for time-evolution of time-dependent Hamiltonians

DaanMaertens:
Adds a new struct `MultipliedOperator `that represents an operator that gets multiplied lazily with another object. Particular subtypes are `TimedOperator` and `UntimedOperator` where the operator gets multiplied by a time-dependent function or a number respectively. The actual multiplication is only done when it is needed e.g. in expectation_value
or when the action of derivative operator is calculated. Works out of the box for VUMPS, GradientGrassmann or TDVP algorithms and with only minor additions also in combination with `LazySum`. A `time_evolve` function was also added that does multiple `timestep`'s and has a convenient finalize that allows for example the bond dimension to be expanded. Comments and suggestions?

### DIFF
--- a/src/MPSKit.jl
+++ b/src/MPSKit.jl
@@ -79,6 +79,7 @@ include("operators/sparsempo/sparsempo.jl")
 include("operators/mpohamiltonian.jl") # the mpohamiltonian objects
 include("operators/mpomultiline.jl")
 include("operators/projection.jl")
+include("operators/timedependence.jl")
 include("operators/lazysum.jl")
 
 include("transfermatrix/transfermatrix.jl")
--- a/src/MPSKit.jl
+++ b/src/MPSKit.jl
@@ -27,7 +27,7 @@ export entanglementplot, transferplot
 
 # hamiltonian things
 export Cache
-export SparseMPO, MPOHamiltonian, DenseMPO, MPOMultiline, LazySum
+export SparseMPO, MPOHamiltonian, DenseMPO, MPOMultiline, UntimedOperator, TimedOperator, LazySum
 export ∂C, ∂AC, ∂AC2, environments, expectation_value, effective_excitation_hamiltonian
 export leftenv, rightenv
 
@@ -81,6 +81,7 @@ include("operators/mpomultiline.jl")
 include("operators/projection.jl")
 include("operators/timedependence.jl")
 include("operators/lazysum.jl")
+include("operators/multipliedoperator.jl")
 
 include("transfermatrix/transfermatrix.jl")
 include("transfermatrix/transfer.jl")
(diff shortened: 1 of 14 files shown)


src/operators/multipliedoperator.jl
```diff
@@ -0,0 +1,56 @@
...
+"""
+    Structure representing a time-independent operator that will be multiplied with a constant coefficient. Consists of
+        - An operator (MPO, Hamiltonian, ...)
+        - A number f that gets multiplied with the operator
+"""
+const UntimedOperator{O} = MultipliedOperator{O,<:Union{Real,One}}
+
+#constructors for (un)TimedOperator
+TimedOperator(x::O, f::F) where {F<:Function,O} = MultipliedOperator(x, f)
+UntimedOperator(x::O, c::C) where {C<:Union{Real,One},O} = MultipliedOperator(x, c)
```
lkdvos, line 27: Do we need explicit constructors for Timed and Untimed? Can we not always use MultipliedOperator, which automatically infers if it is Timed or Untimed based on the input arguments?

src/MPSKit.jl
```diff
@@ -27,7 +27,9 @@ export entanglementplot, transferplot
 
 # hamiltonian things
 export Cache
-export SparseMPO, MPOHamiltonian, DenseMPO, MPOMultiline, LazySum
+export SparseMPO, MPOHamiltonian, DenseMPO, MPOMultiline
+export UntimedOperator, TimedOperator, LazySum
```
lkdvos, line 31: Do we need to export timed and untimed separately? Can we just expose `MultipliedOperator`?

src/algorithms/derivatives.jl
```diff
@@ -28,6 +28,7 @@ Base.:*(h::Union{MPO_∂∂C,MPO_∂∂AC,MPO_∂∂AC2}, v) = h(v);
 (h::MPO_∂∂C)(x) = ∂C(x, h.leftenv, h.rightenv);
 (h::MPO_∂∂AC)(x) = ∂AC(x, h.o, h.leftenv, h.rightenv);
 (h::MPO_∂∂AC2)(x) = ∂AC2(x, h.o1, h.o2, h.leftenv, h.rightenv);
+(h::DerivativeOperator)(v, ::Number) = h(v)
```
lkdvos, line 31: I am completely fine with this as default behaviour, but it does kind of circumvent the goal of not allowing you to use a time argument when the hamiltonian is not time dependent?

DaanMaertens: yes it does, but the idea is that the user should not need to fiddle around with the derivative operators so it is "safe" (read: not confusing) to allow a time argument that gets ignored. The reason to define this is so that it plays nice with the `integrate` function which expects a two argument function.

src/MPSKit.jl
DaanMaertens: Yes we could. At the moment I've only written code for `TimedOperator` and `UntimedOperator` so the user could try to put something in MultipliedOperator which is not (yet) supported and encounter errors down the line. That is why only `TimedOperator` and `UntimedOperator` are exported.

src/operators/multipliedoperator.jl
DaanMaertens: though they are not strictly necessary, I found these constructors handy and intuitive.

Issue #111, opened 2024-02-07
Need (?) for warning when computing multiple excited states per momentum


Gertian:
When computing quasi-particle excitations a krylov algorithm is used to solve the eigenvalue problem.

According to the documentation of KrylovKit it is therefore quite risky to ask for multiple excitatations per momentum in the case that the spectrum is degenerate. 

>Degenerate eigenvalues
>
>From a theoretical point of view, Krylov methods can at most find a single eigenvector associated with a targetted eigenvalue, even if the latter is degenerate. In the case of a degenerate eigenvalue, the specific eigenvector that is returned is determined by the starting vector x₀. For large problems, this turns out to be less of an issue in practice, as often a second linearly independent eigenvector is generated out of the numerical noise resulting from the orthogonalisation steps in the Lanczos or Arnoldi iteration. Nonetheless, it is important to take this into account and to try not to depend on this potentially fragile behaviour, especially for smaller problems.

As far as I know this is not mentioned in the documentation of MPSKit and on top of this it might be useful to use @warn to inform the user about this.

A more practical question : If I'd be asking for the two lowest quasiparticles which I suspect to have degenerate eigenvalues. How would this affect performance ? Could this be incredibly slow ? What exactly is ment with
>generated out of numerical noise ?

and when is this expected to fail ? What happens if it does ?

Gertian:
It also seems as if the verbosity argument of KrylovKit isn't propagated into MPSKit right ?

maartenvd:
Failure means that it will not find the second degenerate eigenvector, and
instead find the one above it. If you suspect the system to have a
degenerate eigenvalue then you could probably run the eigensolver, take a
random vector and project out the already found eigenvector, and run the
eigensolver again with this new initial state.

If the system is indeed degenerate, then you expect both eigensolver calls
to find the same eigenvalue but with two different associated eigenvectors.
If the system is not degenerate, then the second eigensolver call should
either find the same eigenvalue with the same eigenvector, or it could even
fail to find that eigenvalue (since you projected out the exact
eigenvector).

Op wo 7 feb 2024 om 18:11 schreef Gertian ***@***.***>:

Gertian:
I'm sorry but I don't understand why you mean. 

If I run : 
`excitations(H::MPOMultiline, alg::QuasiparticleAnsatz, V₀::Multiline{<:InfiniteQP}, lenvs, renvs; num=my_num, solver=Defaults.linearsolver)`
it's going to call : 
[...]

PR #113, opened 2024-02-11
Quasiparticle optimizations

lkdvos:
Some attempts at speeding up the quasiparticle algorithms:

- removes the computation of renormalization energy from the effective excitation hamiltonian, as this does not need to be recomputed
- adds multithreading over sites, similar to how groundstate algorithms work

### DIFF
--- a/docs/src/man/algorithms.md
+++ b/docs/src/man/algorithms.md
@@ -1,3 +1,9 @@
+```@meta
+DocTestSetup = quote
+    using MPSKit, MPSKitModels, TensorKit
+end
+```
+
 # [Algorithms](@id um_algorithms)
 
 ## Minimizing the energy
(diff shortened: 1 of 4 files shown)


lkdvos:
Might help for #111

maartenvd:
have you experimented with exci_transfers? I recall doing quite a few unnecessary computations there. There is also the way I construct the excitation environments, which goes row by row through the MPO. In practice this can often be parallelized, but it requires being smarter. (A similar thing can be done for the groundstate environments)

lkdvos:
I haven't been focusing too much on parallelization over the different blocks at all, as I mostly wanted to move that into BlockTensorKit, and just experiment with different backends. I must also say that our recent experimentations with multithreading TensorKit have somewhat discouraged me from trying to multithread at that level, as it seems like as soon as the load is not balanced, it is quite hard to outperform just sending all threads to BLAS. One of the things I've been menaing to try is actually the opposite direction, to make the hamiltonian dense for applications of the transfers etc (fully dense for effective hamiltonians or row-wise/column-wise for still being able to do the linear problems in the environments). I have the feeling that at the very least for semi-local hamiltonians, the gain from using BLAS's level of optimisation is so high that the sparseness of the hamiltonian does not really achieve too much

lkdvos:
I think this is actually exactly what is prohibiting making mkl+Julia competitive. One of the optimizations that BLAS relies on quite heavily is making effective use of the cpu caches, which makes them effectively do less work -- they need to load entries from actual ram less often. This is known to be quite a dominant factor for `mul!`, especially when the matrices aren't too large. In that sense, trying to do multiple matrix multiplications on different cores invalidates some of that work (at the very least for the L3 cache which is typically shared). 

I must admit that I am also not too happy about this, as this is really a "if you can't be smart, just be strong" kind of solution, but if it just works than it's hard to avoid :grin:

lkdvos:
I will merge this for now, it seems like this is strictly an improvement, and we can make further improvements later

Gertian:
It's definitely an improvement. After experimenting a bit it's definitely not clear that using MKL over Openblas is an improvement though. This might obviously be due to bad choices in num_threads

PR #114, opened 2024-02-13
Update README.md

maartenvd:


lkdvos:
I somehow missed this, my apologies. I made the readme use a table, similar to tensorkit and tensoroperations, so if you're okay with this I'll just leave that?

maartenvd:
You are forgiven

PR #118, opened 2024-02-20
refactor dynamic tolerances using `ThrottledTol` 

lkdvos:
This cleans up some of the keyword arguments of VUMPS that are not necessary when you do not want to use dynamic tolerances. Additionally, it removes `between`, as this is a copy of `Base.clamp`, and streamlines the creation of default subalgorithms. Finally, `dynamical` is replaced with `dynamic`.

Note that this is quite breaking, as many of the original keywords for VUMPS no longer exist. I think this is warranted, and want to include most of these breaking changes in the transition to v0.11.0. Do you think explicit deprecation warnings are necessary/warranted, or can we just make the changes breaking with the argument v0.x can change the interface?

### DIFF
--- a/src/utility/throttledtol.jl
+++ b/src/utility/throttledtol.jl
@@ -1,3 +1,10 @@
+module ThrottledTols
+
+import ..MPSKit: Algorithm
+using Accessors
+
+export updatetol, ThrottledTol
+
 @doc """
     updatetol(alg, iter, ϵ)
 
@@ -48,4 +55,6 @@ end
 # default implementation with Accessors.jl, but can be hooked into
 function _updatetol(alg, tol::Real)
     return @set alg.tol = tol
+end
+
 end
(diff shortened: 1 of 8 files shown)


src/utility/throttledtol.jl
DaanMaertens, line 1: Maybe we can rename the module to `DynamicTols`? I find this a more intuitive name.

DaanMaertens (approved):

src/utility/throttledtol.jl
lkdvos: I had it like this originally, but it's a bit annoying to distinguish these cases:
```julia
alg = DynamicalDMRG{...}
alg = DynamicTol(DMRG)
```
Basically, dynamic means changing, while dynamical means concerning dynamics, and I didn't like how close together these are. I don't have that strong of an opinion on it though, so I can definitely change it

PR #119, opened 2024-02-20
Refactor tests into modules

### LINKED ISSUE #115: Tests complain about duplicate includes of `setup.jl`
Because the setup.jl file is included for each of the test groups, but none of these actually use modules to separate their namespaces, julia throws a bunch of complaints about re-including a file. This could easily be fixed by making either/both setup.jl and the test groups their own module.

lkdvos:
This removes the warning messages that the tests were throwing about multiple inclusions. Fixes #115 .

Additionally, changes some of the tolerances and bonddimensions around in the test algorithms for groundstate search to make them a bit more stable

### DIFF
--- a/test/algorithms.jl
+++ b/test/algorithms.jl
@@ -13,7 +13,7 @@ using TensorKit: ℙ
 
 @testset "find_groundstate" verbose = true begin
     tol = 1e-8
-    verbosity = 2
+    verbosity = 0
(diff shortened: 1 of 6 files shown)


DaanMaertens:
Looks good. No comments

DaanMaertens (approved):

PR #122, opened 2024-02-22
Functionality to time evolve `WindowMPS`

DaanMaertens:
Adds `timestep`/`timestep!` functions to time evolve a `WindowMPS` with a window Hamiltonian. `WindowMPS` now has two extra type parameters to indicate whether the left and/or right infinite part are variable with regards to time evolution. When both are fixed the time evolution is equivalent to the one already available for `WindowMPS` in MPSKit.

A new struct `WindowEnv`has been added that keeps track of the environments for a `WindowMPS`. Whenever `leftenv`/`rightenv` is called, there is an automated check that the left and rightstart of the finite part are correct (i.e. from the infinite part).


DaanMaertens:
@Gertian Time evolving a finite window in a bath of the groundstate is already possible in MPSKit. This PR aims to extend the time evolution so that the infinite states can be evolved as well. As long as the tensors at the boundary of the finite window are the same as the ones from the infinite part (i.e. the dynamics have not yet reached the boundary of the finite window) there should not be any problem with phases.

Gertian:
@lkdvos @DaanMaertens I remember that I was playing around with this years ago and that it was not entirely trivial to get these phases to work (even though it should be since you know exactly which tensor will pick up which phase, i.e. I was probably being a bit of a nooby...)

Either way as Lukas said, surely there will be phases since even an exact eigenstate evolves as |psi> -> exp( i E t) |psi> and this should be accounted for. If not at least there should be a good reason why not to do it...

For example, I could see that it drops out in all expectation values that one might compute but I'm not sure that this is enough.

DaanMaertens:
I would think that it drops out everywhere (in expectation values, in effective hamiltonians like `∂∂AC`), but I could be mistaken.

lkdvos:
> I would think that it drops out everywhere (in expectation values, in effective hamiltonians like `∂∂AC`), but I could be mistaken.

It definitely does not drop out if you do things like lochsmidt echo, where you take overlaps at different times

lkdvos:
Also Daan, van you run the formatter so we don't hit the GitHub limit ?

PR #130, opened 2024-03-15
KrylovKit v0.7

lkdvos:
Adds support for KrylovKit v0.7

Mostly required VectorInterface implementation for QuasiParticle states.


lkdvos:
Replaces #129

PR #131, opened 2024-03-15
Update README.md

v1j4y:
Updated the `verbosity=0` keyword argument to `verbose=false` to make the example run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,14 +105,14 @@ using MPSKit, MPSKitModels, TensorKit
 using ProgressMeter, Plots # for demonstration purposes
 
 D = 4 # bonddimension
-init_state = InfiniteMPS(ℂ^2, ℂ^D)
+init_state = MPSKit.InfiniteMPS(ℂ^2, ℂ^D)
 
 g_values = 0.1:0.1:2
 Z = @mpoham sum(σᶻ(){i} for i in vertices(InfiniteChain()))
 
 M = @showprogress map(g_values) do g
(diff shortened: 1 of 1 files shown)


lkdvos:
Thanks for noticing and fixing this!
I am a bit confused by the `InfiniteMPS` not being exposed, as it really should be exported. [see this line](https://github.com/maartenvd/MPSKit.jl/blob/af3d3523983ba3814c338c0584e03e953e126265/src/MPSKit.jl#L16)
Am I missing something here?
In any case, I really appreciate the help!

lkdvos:
I looked into this a bit more, and my apologies for the initial confusion.

I've been making some breaking changes to streamline the interface of MPSKit. In particular, I normalized the `verbosity` and `verbose` keywords to all be `verbosity` (See #125 ), as this is more in line with `KrylovKit`. However, there still are some breaking changes that I want to make before releasing a new version v0.11, in order to bundle all breaking changes in a single release. Thus, the master branch is a bit out of sync with the latest released version, so you'll have to bear with me for a bit.
(Thus, the readme changes really are on purpose)

In any case, I apologize for the confusion and inconvenience, and hope to get this sorted ASAP.

v1j4y:
Hello !

Thanks for the detailed reply.

I installed using `] add MPSKit` (i.e. the latest version) which explains everything :slightly_smiling_face: 

I'll close the PR since the issue was simply due to a change in the API which as you said will be updated soon.

All the best,
 Vijay

PR #147, opened 2024-06-05
Add Chepiga ansatz

lkdvos:
Co-authored-by: Atsushi Ueda

### DIFF
--- a/docs/src/man/algorithms.md
+++ b/docs/src/man/algorithms.md
@@ -221,6 +221,49 @@ ts = FiniteMPS(10,ℂ^2,ℂ^12);
 (energies,Bs) = excitations(th,FiniteExcited(),ts,envs);
 ```
 
+### "Chepiga Ansatz"
+
+Sometimes, it might be sufficient to find excited states by only locally perturbing the groundstate.
+To this end, you might consider selecting a single site, and allowing only variations for these tensors, keeping all other tensors fixed.
+This algorithm is especially well-suited whenever the system is sufficiently gapped.
+This is supported via the following syntax:
+
+```jldoctest; output=false
+g = 10.0
+L = 16
+H = transverse_field_ising(; g)
+ψ₀ = FiniteMPS(L, ComplexSpace(2), ComplexSpace(32))
+ψ, envs, = find_groundstate(ψ₀, H; verbosity=0)
+E₀ = real(sum(expectation_value(ψ, H, envs)))
+Es, ϕs = excitations(H, ChepigaAnsatz(), ψ, envs; num=1)
+isapprox(Es[1] - E₀,  2(g - 1); rtol=1e-2) # infinite analytical result
+
+# output
+
+true
+```
+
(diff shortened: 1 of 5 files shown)


Gertian:
@lkdvos could add a reference for this Ansatz ?

lkdvos:
It's already in the docstrings and the documentation page, not sure where else you want the reference?

Gertian:
> It's already in the docstrings and the documentation page, not sure where else you want the reference?

I ment what Jutho just mentioned above. But I also agree that this is not easy. From my part this was just out of curiosity, I didn't notice you already mentioned the source in the docstring.

Jutho:
Something like `LocalizedExcitationAnsatz` (maybe even without the `Ansatz` part)?

Also, I am not sure whether the documentation advice is correct that this is good if the system is "sufficiently" gapped. Even in the gapped case, you need the excitation to delocalise. My impression was that this ansatz actually works well in the case of critical systems, because there the virtual space (of an MPS that well approximates the ground state) contains sufficient degrees of freedom to change the global state by only modifying a single local tensor. I.e. it is the much slower decay of correlations in such states that makes this ansatz work. If you do this is a strongly gapped state like AKLT, it would be pretty bad I think (one could easily check). There, even with the momentum space ansatz (I still don't like the quasiparticle name either :smile:), modifying a single tensor is not sufficient to reach great precision, and this should be strictly worse.

lkdvos:
Ah, I seem to have read over the abstract too quickly. In any case, it seems like this ansatz is probably contained within the manifold of the quasiparticle ansatz, so I would guess in general this is strictly "worse" than that, however it is also a whole lot cheaper. I'll try and adapt the doc page a bit, it's not too easy to write things for an algorithm I know very little about. I think it's mostly added for easy comparison, and since Atsushi sent me the code I just put it in.

Considering the name, we already have a couple of these in MPSKit, as for example `marek_gap` is presumably not known beyond our office, and VOMPS also gives no hits on google scholar.  In that sense, I actually quite like the name, as it is A. short and easy to remember B. has already been used colloquially by some people in Benasque and in our office and C. has some explicit form of credit/citation in the code, so there is some benefit to naming it that way (I did not come up with this myself).  I agree that it is a bit personal, so maybe I'll send an email to the authors asking for their permission, or if they use a different name themselves?

Jutho:
`marek_gap` definitely needs to go, that's not even using a family name. The problem with this is that you only credit the first author, which is often not a fair representation of reality.

Jutho:
For `marek_gap`, it anyway seems the documentation is very limited and the interface is not at all specified. The first return value is not the "Marek" gap but the actual (inverse) correlation length, and only the second return value is the gap. Personally, I would think that interface in the form of

`(inverse_)correlation_length_spectrum(state, howmany, some options such as which symmetry sector or which angle)`

would be more useful. Then there could be a `first_gap` utility, to find the first gap in a list of values that represent a `spectrum` of some kind. But this should probably not be discussed here.

PR #149, opened 2024-06-06
Add chepiga comment

dartsushi:
I wrote some additional comments on Chepiga ansatz. This code, in principle, shall be used near the criticality. Thus, I propose to change g to 1.0 instead too if possible.


docs/src/man/algorithms.md
```diff
@@ -222,9 +222,7 @@ ts = FiniteMPS(10,ℂ^2,ℂ^12);
 ```
 
 ### "Chepiga Ansatz"
-
-Sometimes, it might be sufficient to find excited states by only locally perturbing the groundstate.
-To this end, you might consider selecting a single site, and allowing only variations for these tensors, keeping all other tensors fixed.
+Computing excitations in critical systems poses a significant challenge due to the diverging correlation length, which requires very large bond dimensions. However, we can leverage this long-range correlation to effectively identify excitations. In this context, the left/right gauged MPS, serving as isometries, are effectively projecting the Hamiltonian into the low-energy sector. This projection method is particularly effective in long-range systems, where excitations are distributed throughout the entire system. Consequently, the low-lying energy spectrum can be extracted by diagonalizing the effective Hamiltonian(without any additional DMRG costs!). The states of these excitations are then represented by the ground state MPS, with one site substituted by the corresponding eigenvector. This approach is often referred to as the 'Chepiga ansatz,' named after the author of this paper.
```
lkdvos, line 225:
```suggestion
Computing excitations in critical systems poses a significant challenge due to the diverging correlation length, which requires very large bond dimensions. However, we can leverage this long-range correlation to effectively identify excitations. In this context, the left/right gauged MPS, serving as isometries, are effectively projecting the Hamiltonian into the low-energy sector. This projection method is particularly effective in long-range systems, where excitations are distributed throughout the entire system. Consequently, the low-lying energy spectrum can be extracted by diagonalizing the effective Hamiltonian (without any additional DMRG costs!). The states of these excitations are then represented by the ground state MPS, with one site substituted by the corresponding eigenvector. This approach is often referred to as the 'Chepiga ansatz', named after one of the authors of this paper.
```

lkdvos: Can you maybe also redistribute the lines such that it's not a single long line? I think we currently use a 92 character line wrapping

lkdvos:
Failing tests are unrelated, so I'll merge this

PR #158, opened 2024-07-08
Fix hard hexagon example

leburgel:
Fixes the hard hexagon example by using a non-Hermitian default eigensolver.


lkdvos (approved):
Looks good, thanks!

PR #166, opened 2024-08-06
Update docs

Gertian:
I made some changes to the docs for changebonds


lkdvos:
Can you format the docstrings with the maximal line length, and maybe also add something about the behaviour of the `trscheme` kwarg?

Gertian:
Will do, and also, I'll try to get rid of the extra stuff in the PR :p

PR #168, opened 2024-08-07
Documentation update: `changebonds` improvements

Gertian:
The pull request that I ment to make yesterday. 

@lkdvos, I still need to fix the formatting but I'm not sure how long the lines should be and if I should do this manually or using some tool ?

### DIFF
--- a/docs/src/man/algorithms.md
+++ b/docs/src/man/algorithms.md
@@ -354,8 +354,7 @@ disadvantages:
   Because of the two-site update, this can again become expensive, but the algorithm has the
   option of both expanding as well as truncating the bond dimension. 
 
-  Here the trscheme kwarg truncates the full state s.t. for example trscheme=truncdim(d)
-  will return a state with bonddimension smaller than d.
+  Here, `trscheme` controls the truncation of the state after the two-site update.
 
 ## leading boundary
 
--- a/docs/src/man/algorithms.md
+++ b/docs/src/man/algorithms.md
@@ -8,8 +8,8 @@ end
 
 ## Minimizing the energy
 
-There are a number of different possible energy-minimization algorithms,
-depending on the system size. Exclusive to finite systems are
+There are a number of different possible energy-minimization algorithms, depending on the
+system size. Exclusive to finite systems are
 
     - DMRG
 
@@ -23,12 +23,11 @@ Exclusive to infinite systems are
 
     - VUMPS
 
-with a last algorithm - GradientGrassmann - implemented for both finite and
-infinite systems.
+with a last algorithm - GradientGrassmann - implemented for both finite and infinite
+systems.
 
-WindowMPS, which is a finite patch of mutable tensors embedded in an infinite
-MPS, is handled as a finite system where we only optimize over the patch of
-mutable tensors.
+WindowMPS, which is a finite patch of mutable tensors embedded in an infinite MPS, is
+handled as a finite system where we only optimize over the patch of mutable tensors.
 
 ### DMRG
 
@@ -38,12 +37,12 @@ operator = nonsym_ising_ham();
 (groundstate,environments,delta) = find_groundstate!(state,operator,DMRG())
 ```
 
-The DMRG algorithm sweeps through the system, optimizing every site. Because of
-its single-site behaviour, this will always keep the bond dimension fixed. If
-you do want to increase the bond dimension dynamically, then there are two
-options. Either you use the two-site variant of DMRG (DMRG2()), or you make use
-of the finalize option. Finalize is a function that gets called at the end of
-every DMRG iteration. Within that function call one can modify the state.
+The DMRG algorithm sweeps through the system, optimizing every site. Because of its
+single-site behaviour, this will always keep the bond dimension fixed. If you do want to
+increase the bond dimension dynamically, then there are two options. Either you use the
+two-site variant of DMRG (DMRG2()), or you make use of the finalize option. Finalize is a
+function that gets called at the end of every DMRG iteration. Within that function call one
+can modify the state.
 
 ```julia
 function my_finalize(iter,state,H,envs)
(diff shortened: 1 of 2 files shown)


docs/src/man/algorithms.md
lkdvos, line 1: It looks like you have to change a bit the setting on the linewidth...
I think we used [Julia's default](https://docs.julialang.org/en/v1/manual/documentation/#Syntax-Guide) (92 lines)

Gertian: Okay, thanks for pointing this out. I'll redo do tomorrow or tonight :)

docs/src/man/algorithms.md
```diff
@@ -287,40 +316,62 @@ disadvantages:
   locally truncating the state using an SVD decomposition. This yields a (locally) optimal
   truncation, but clearly cannot be used to increase the bond dimension. Note that a
   globally optimal truncation can be obtained by using the [`SvdCut`](@ref) algorithm in
-  combination with [`approximate`](@ref). The state will remain largely unchanged.
+  combination with [`approximate`](@ref). Since the output of this method might have a
+  truncated bonddimension, the new state might not be identical to the input state.
+
+  The truncation is implemented through the kwarg trscheme which truncates the singular
+  values of the inputted state.
```
lkdvos, line 323:
```suggestion
  The truncation is controlled through `trscheme`, which dictates how the singular values
  of the original state are truncated.
```

docs/src/man/algorithms.md
```diff
@@ -287,40 +316,62 @@ disadvantages:
...
 * [`OptimalExpand`](@ref): This algorithm is based on the idea of expanding the bond
   dimension by investigating the two-site derivative, and adding the most important blocks
   which are orthogonal to the current state. From the point of view of a local two-site
   update, this procedure is *optimal*, but it requires to evaluate a two-site derivative,
-  which can be costly when the physical space is large. The state will remain unchanged, but
-  a one-site scheme will now be able to push the optimization further.
+  which can be costly when the physical space is large. 
+
+  Since the new sectors are added with singular values 0 the state will remain unchanged.
+  Afterwards a one-site scheme will be able to push the optimization further.
```
lkdvos, line 333: I'm not sure if this change really does anything, and it also adds singular vectors in sectors that already appear, so I would honestly just keep the old wording here.
```suggestion
  which can be costly when the physical space is large. The state will remain unchanged, but
  a one-site scheme will now be able to push the optimization further.
```

docs/src/man/algorithms.md
```diff
@@ -287,40 +316,62 @@ disadvantages:
...
+
+  For optimalexpand the kwarg trscheme only truncates the added sectors. E.g. with
+  trscheme=truncdim(d_extra) the bonddimension increases with no more than d_extra.
```
lkdvos, line 336:
```suggestion
  The subspace used for expansion can be controlled through `trscheme`, which dictates how many singular values will be added.
```
(+ format, doing this on github)

docs/src/man/algorithms.md
```diff
@@ -287,40 +316,62 @@ disadvantages:
...
 
 * [`RandExpand`](@ref): This algorithm similarly adds blocks orthogonal to the current
   state, but does not attempt to select the most important ones, and rather just selects
   them at random. The advantage here is that this is much cheaper than the optimal expand,
   and if the bond dimension is grown slow enough, this still obtains a very good expansion
-  scheme. Again, the state will remain unchanged, and a one-site scheme will be able to push
-  the optimization further.
+  scheme.
```
lkdvos, line 342: same comment here

docs/src/man/algorithms.md
```diff
@@ -287,40 +316,62 @@ disadvantages:
...
   them at random. The advantage here is that this is much cheaper than the optimal expand,
   and if the bond dimension is grown slow enough, this still obtains a very good expansion
-  scheme. Again, the state will remain unchanged, and a one-site scheme will be able to push
-  the optimization further.
+  scheme. 
+  
+  Again, the new sectors are added with singular value 0 so that the state remains
+  unchanged, and a one-site scheme will be able to push the optimization further.
+
+  Similarly to the Optimalexpand algortihm the trscheme kwarg only truncates added sectors.
```
lkdvos, line 347: I would also copy the sentence from above

docs/src/man/algorithms.md
```diff
@@ -287,40 +316,62 @@ disadvantages:
...
-  of both expanding as well as truncating the bond dimension. Note that this will change the
-  state itself, as it consists of an update step.
+  consists of performing a two-site update (which introduced new sectors), and then
+  truncating the state back down using some truncation algorithm. In contrast with the
+  previous methods the new sectors are not added with singular values 0 (but rather with
+  singular values optimal to lower the energy) so that the outputted state is, in general,
+  not equal to the input state. 
+
+  Because of the two-site update, this can again become expensive, but the algorithm has the
+  option of both expanding as well as truncating the bond dimension.
```
lkdvos, line 357: I'm also not sure if this contains more information, the note on new sectors is not necessarily true, and the singular values being 0 really is an implementation detail that only explains something if you already know what is going on...

docs/src/man/algorithms.md
```diff
@@ -287,40 +316,62 @@ disadvantages:
...
+  truncating the state back down using some truncation algorithm. In contrast with the
+  previous methods the new sectors are not added with singular values 0 (but rather with
+  singular values optimal to lower the energy) so that the outputted state is, in general,
+  not equal to the input state. 
+
+  Because of the two-site update, this can again become expensive, but the algorithm has the
+  option of both expanding as well as truncating the bond dimension. 
 
+  Here the trscheme kwarg truncates the full state s.t. for example trscheme=truncdim(d)
+  will return a state with bonddimension smaller than d.
```
lkdvos, line 360:
```suggestion
  Here, `trscheme` controls the truncation of the state after the two-site update.
```

docs/src/man/algorithms.md
Gertian: For me this is still not entirely clear.


I think : The subspace used for expansion can be truncation using `trscheme` .

Might be clearer ?

docs/src/man/algorithms.md
Gertian: I agree that the part on sinvals being zero can be dropped.


Regarding the rest, doesn't the two site update add sectors that lower the energy ?

lkdvos: I don't know if it is guaranteed to lower the energy actually, but it definitely tries to :)
The thing is, it adds singular values optimal (in a twosite sense) to lower the energy, which you will then use for the onesite energy, and there's a whole can of worms I don't want to touch :p 
Mostly what I meant is, it does not necessarily add new sectors (for example in Z2 it would only add degeneracy), and while I know what you mean with that statement, that might be confusing.
This is also only applicable to hamiltonians, while you could also use this for transferMPO stuff or PEPS things (although I am not sure if that's currently implemented)

Gertian:
@lkdvos , I think this now has all your suggestions in it.

PR #172, opened 2024-08-09
Some `MPOHamiltonian` fixes

leburgel:
There were two examples in the new `MPOHamiltonian` documentation that failed during the docs build, which are fixed here.

One was just a typo here
https://github.com/QuantumKitHub/MPSKit.jl/blob/965999ad5d5dd808c2802d59a18557db0a44cb53/docs/src/man/operators.md?plain=1#L234
where `identity` should have been `id`. I also had to specify the storage type since it could not convert to `Matrix{ComplexF64}` the default `Matrix{Float64}` here
https://github.com/QuantumKitHub/MPSKit.jl/blob/965999ad5d5dd808c2802d59a18557db0a44cb53/src/operators/sparsempo/sparsempo.jl#L72
which threw
```julia
ERROR: MethodError: Cannot `convert` an object of type 
  TrivialTensorMap{ComplexSpace, 2, 2, Matrix{ComplexF64}} to an object of type 
  Union{Float64, TrivialTensorMap{ComplexSpace, 2, 2, Matrix{Float64}}}
```

The second was an issue with iterating over a `CartesianIndex` here
https://github.com/QuantumKitHub/MPSKit.jl/blob/965999ad5d5dd808c2802d59a18557db0a44cb53/src/operators/mpohamiltonian.jl#L124
which caused this example to fail:
https://github.com/QuantumKitHub/MPSKit.jl/blob/965999ad5d5dd808c2802d59a18557db0a44cb53/docs/src/man/operators.md?plain=1#L124-L153
I just replaced the `I...` with `Tuple(I)...` which fixed the example. This should not break for `I::Tuple{Vararg{Int}}` which is good, but I don't know if there are other allowed types for specifying vertices for which this fix won't work?

### DIFF
--- a/docs/src/man/operators.md
+++ b/docs/src/man/operators.md
@@ -231,7 +231,7 @@ matrix, and the third dimension specifies the column of the matrix.
 
 ```@example operators
 data = Array{Any,3}(missing, 1, 3, 3) # missing is interpreted as zero
-data[1, 1, 1] = identity(ℂ^2)
+data[1, 1, 1] = id(Matrix{ComplexF64}, ℂ^2)
 data[1, 3, 3] = 1 # regular numbers are interpreted as identity operators
 data[1, 1, 2] = -J * S_x
 data[1, 2, 3] = S_x
--- a/docs/src/man/operators.md
+++ b/docs/src/man/operators.md
@@ -132,17 +132,17 @@ end
 
 # horizontal and vertical interactions are easier using Cartesian indices
 horizontal_operators = Dict()
-I_horizontal = CartesianIndex(1, 0)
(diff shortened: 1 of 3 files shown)


lkdvos:
Looks like great fixes to me! Any chance you could add something that uses the CartesianIndex syntax to one of the tests, such that it would detect this?

leburgel:
One last maybe less relevant comment: shouldn't 'horizontal' and 'vertical' be switched in the example linked above? Thinking of indexing the square lattice as a conventional matrix using `CartesianIndices`, wouldn't horizontal interactions correspond to a shift by `CartesianIndex(0, 1)` while vertical interactions correspond to a shift by `CartesianIndex(1, 0)`?

PR #174, opened 2024-08-12
Fix issue where SVDCut produces InfiniteMPS that is not full rank

lkdvos:
`changebonds` with `SVDCut` used the `InfiniteMPS(AL, C0)` constructor, and this constructor did not check that all tensors are full rank, or try to correct for it. `InfiniteMPS(AL)` does this, so this fixes the issue.

### DIFF
--- a/src/algorithms/changebonds/svdcut.jl
+++ b/src/algorithms/changebonds/svdcut.jl
@@ -83,7 +83,16 @@ function changebonds(ψ::InfiniteMPS, alg::SvdCut)
         copied[i + 1] = _transpose_front(U' * _transpose_tail(copied[i + 1]))
     end
 
-    return normalize!(InfiniteMPS(copied, complex(ncr)))
+    # make sure everything is full rank:
+    makefullrank!(copied)
+
+    # if the bond dimension is not changed, we can keep the same center, otherwise recompute
+    ψ = if space(ncr, 1) != space(copied[1], 1)
+        InfiniteMPS(copied)
+    else
+        InfiniteMPS(copied, complex(ncr))
+    end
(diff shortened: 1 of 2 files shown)


Gertian:
Thanks for looking into this !

This is probably a stupid question but tensors not being full rank corresponds to them not having a block for certain sectors ?

lkdvos:
Not necessarily, this also happens for non-symmetric tensors:

An mpstensor of dim 2x2x5 can never be full rank from left to right, which basically means that there is "one column too much". This can be transformed away without changing the state, and some of our algorithms fail because of this.

The exact condition is really just that both R <= P * L, and L <= P * R (with appropriate arrows), which can happen when you cut drastically or with strange truncation schemes.

Note that a sector not appearing really is not a problem, it is completely fine to have spin 1/2 on the left virtual space and the physical space, and only have spin 0 on the right.
Conversely, it would not be fine to only have spin 2 on the right, because then the condition is violated again.

As a final remark to go fully in depth: the error you had in VUMPS originates from the fact that TensorKit has a QR decomposition that makes Q square, it does not add zero rows because then it loses it's meaning as orthogonal basis.
A 4x5 matrix would thus have a 4x4 Q and a 4x5 R, where the last row of the R is zero. If you then multiply the R to the other side of the bond, this bond now suddenly becomes dimension 4 instead of 5, hence the spacemismatch. I wrongly assumed that our constructor automatically converts everything to full rank this way, because it is built into the constructor, but apparently not all constructors, thus the issue

On the other hand, it is perfectly fine to have a sector not appearing,

Gertian:
Perfect, thank you for this great explanation :)

PR #175, opened 2024-08-13
Incorporated an alg argument for GD which can be set to SDD(fast, unstable) or SVD(slow, stable)

Gertian:
Together with https://github.com/Jutho/TensorKitManifolds.jl/pull/11 this implements functionality that allows the user to choose which SVD algorithm is being used internally in the gradient descent gradient calculation. 

The options are SDD(fast but potentially unstable) vs SVD(slow but stable).

These two PR should fix the issue : https://github.com/QuantumKitHub/MPSKit.jl/issues/109

### DIFF
--- a/src/algorithms/groundstate/gradient_grassmann.jl
+++ b/src/algorithms/groundstate/gradient_grassmann.jl
@@ -23,8 +23,7 @@ with a preconditioner to induce the metric from the Hilbert space inner product.
 - `maxiter::Int`: maximum amount of iterations
 - `verbosity::Int`: level of information display
 - `alg::OrthogonalFactorizationAlgorithm` : specifies which algorithm is used for singular
-  value decompositions in the retract and transport functions. Choices are SDD() which is faster but
-  less stable and SVD() which is slower but more stable.
+  value decompositions in the retract and transport functions.
 """
 struct GradientGrassmann <: Algorithm
     method::OptimKit.OptimizationAlgorithm
--- a/src/algorithms/groundstate/gradient_grassmann.jl
+++ b/src/algorithms/groundstate/gradient_grassmann.jl
@@ -30,7 +30,8 @@ struct GradientGrassmann <: Algorithm
     finalize!::Function
     svd_alg::OrthogonalFactorizationAlgorithm
     function GradientGrassmann(; method=ConjugateGradient, (finalize!)=OptimKit._finalize!,
-                               tol=Defaults.tol, maxiter=Defaults.maxiter, verbosity=2, svd_alg=TensorKit.SDD()) 
+                               tol=Defaults.tol, maxiter=Defaults.maxiter, verbosity=2,
+                               svd_alg=TensorKit.SDD())
         if isa(method, OptimKit.OptimizationAlgorithm)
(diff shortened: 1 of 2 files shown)


lkdvos:
I think you swapped SDD (fast, unstable) with SVD (slower, stable).
Can you try and run the formatter as well?

src/algorithms/groundstate/gradient_grassmann.jl
```diff
@@ -22,13 +22,14 @@ with a preconditioner to induce the metric from the Hilbert space inner product.
 - `tol::Float64`: tolerance for convergence criterium
 - `maxiter::Int`: maximum amount of iterations
 - `verbosity::Int`: level of information display
+- `alg::OrthogonalFactorizationAlgorithm` : algorithm to perform the orthogonal factorizations during the automatic differentiation
```
lkdvos, line 25: automatic differentiation?

Gertian: I ran the formatter and changed this line to : 
`alg::OrthogonalFactorizationAlgorithm` : specifies which algorithm is used for singular
  value decompositions in the retract and transport functions. Choices are SDD() which is faster but
  less stable and SVD() which is slower but more stable.

lkdvos: I think if the type specifyer is there, it's probably not a great idea to also list the options, as this can change in the future (think `SafeSVD`), in which case we have to simply remember to update the docstring, which is unlikely to happen.

Gertian: That's a good point. I'll resolve this tomorrow.

Gertian:
@lkdvos I updated everything according to your input.
Somehow the Format suggestion test keeps failing. Despite me formatting all files I touched.

Gertian:
> @lkdvos I updated everything according to your input. Somehow the Format suggestion test keeps failing. Despite me formatting all files I touched.

Ok, this has been resolved. For later reference, I had to make sure VSCode had the correct workspace loaded (so that .JuliaFormatter.toml was in my working directory) and then just run the normal formatter :)

lkdvos:
If you want, you can always manually run the formatter on the entire repository:
```julia
using JuliaFormatter
format("src")
```

Gertian:
@Jutho , @lkdvos what is the status here ?
Can this and the corresponding TensorKitManifolds PR https://github.com/Jutho/TensorKitManifolds.jl/pull/11 be merged ? Or is there still work to be done ?

Jutho:
Again, apologies for the late response here too. With my suggestions for the TensorKitManifolds.jl, I don't think this is necessary anymore. I agree this approach is nicer, and so maybe this PR can stay open, but this only makes sense after a larger TensorKitManifolds.jl overhaul. As mentioned TensorKitManifolds.jl, there are also different retraction schemes possible for Grassmann, and the `alg` keyword of `retract` should be used for that, and might then be configurable in the `GrassmannMPS` algorithm.

Gertian:
Ok. I personally don't mind the specific implementation as long as I have access to the stable SVD through some option in either MPSKit or TensorKitManifolds :+1:

PR #178, opened 2024-10-15
This PR makes GD paralellize over sites

Gertian:
To remarks :

1) I failed to paralelize 
```julia
function retract(x::ManifoldPoint{<:FiniteMPS}, g, alpha)
    #TODO : support parralelize_sites.
    state = x.state
    envs = x.envs

    y = copy(state)  # The end-point
    h = similar(g)  # The tangent at the end-point
    for i in 1:length(g)
        yal, th = Grassmann.retract(state.AL[i], g[i].Pg, alpha)
        h[i] = PrecGrad(th)
        y.AC[i] = (yal, state.CR[i])
    end
    normalize!(y)

    n_y = ManifoldPoint(y, envs)

    return n_y, h
end
```

2)I don't think there are explicit tests for multi-site unit cells. Is this something that should be added ?

### DIFF
--- a/src/algorithms/grassmann.jl
+++ b/src/algorithms/grassmann.jl
@@ -11,6 +11,7 @@ The module exports nothing, and all references to it should be qualified, e.g.
 module GrassmannMPS
 
 using ..MPSKit
+using ..Defaults
 using TensorKit
 import TensorKitManifolds.Grassmann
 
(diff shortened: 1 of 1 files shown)


Gertian:
Ok I'm sorry for the mess.
I'm trying to do another PR at the same time so I need to give this one a better name.

In fact I should have done that in the first place...

Remaking this with better name  :)

PR #179, opened 2024-10-15
This PR makes Gradient Descent parallelized using Threads.@spawn

Gertian:
As the title suggest this PR makes Gradient descent parallel...

Two remarks are : 
1) I failed to parallelize : 
```julia
function retract(x::ManifoldPoint{<:FiniteMPS}, g, alpha)
    #TODO : support parralelize_sites.
    state = x.state
    envs = x.envs

    y = copy(state)  # The end-point
    h = similar(g)  # The tangent at the end-point
    for i in 1:length(g)
        yal, th = Grassmann.retract(state.AL[i], g[i].Pg, alpha)
        h[i] = PrecGrad(th)
        y.AC[i] = (yal, state.CR[i])
    end
    normalize!(y)

    n_y = ManifoldPoint(y, envs)

    return n_y, h
end
```

2) the tests currently don't consider n-site unit cells. To cure this I'll make some new tests in another PR which do 3 site unit cells.

### DIFF
--- a/src/algorithms/grassmann.jl
+++ b/src/algorithms/grassmann.jl
@@ -76,9 +76,9 @@ function ManifoldPoint(state::Union{InfiniteMPS,FiniteMPS}, envs)
                 al_d[i] = MPSKit.∂∂AC(i, state, envs.opp, envs) * state.AC[i]
             end
         end
-        g = map(CartesianIndices(state.AL)) do I
-            return Threads.@spawn Grassmann.project(al_d[I], state.AL[I])  
-        end .|> fetch
+        g = fetch.(map(CartesianIndices(state.AL)) do I
+                       return Threads.@spawn Grassmann.project(al_d[I], state.AL[I])
+                   end)
     else
         al_d = similar(state.AL)
         for i in 1:length(state)
--- a/src/algorithms/grassmann.jl
+++ b/src/algorithms/grassmann.jl
@@ -69,7 +69,7 @@ struct ManifoldPoint{T,E,G,C}
 end
 
 function ManifoldPoint(state::Union{InfiniteMPS,FiniteMPS}, envs)
-    g = similar(state.AL)
+    g = Vector{  Grassmann.GrassmannTangent  }(undef, length(state.Al))
     @static if Defaults.parallelize_sites
         @sync for i in 1:length(state)
             Threads.@spawn begin
--- a/src/algorithms/grassmann.jl
+++ b/src/algorithms/grassmann.jl
@@ -11,6 +11,7 @@ The module exports nothing, and all references to it should be qualified, e.g.
 module GrassmannMPS
 
 using ..MPSKit
+using ..Defaults
 using TensorKit
 import TensorKitManifolds.Grassmann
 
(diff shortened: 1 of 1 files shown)


Gertian:
I made the extra tests which work for the non-modified code.
I'm now running them locally on my machine on this updated version.

Gertian:
@lkdvos I don't understand how to make codecov happy.

The new lines are/are not being tested depending on MPSKit.Defaults aren't they ? 

I understand that we might want to test both scenarios but again, this is currently not implemented in any tests as far as I understand ?

Gertian:
The failing macOS test seems to be unrelated to the code I've added.

src/algorithms/grassmann.jl
```diff
@@ -69,22 +70,38 @@ end
...
+                al_d[i] = MPSKit.∂∂AC(i, state, envs.opp, envs) * state.AC[i]
+            end
+        end
+    else
+        for i in 1:length(state)
+            al_d[i] = MPSKit.∂∂AC(i, state, envs.opp, envs) * state.AC[i]
+        end
     end
-
     g = Grassmann.project.(al_d, state.AL)
```
lkdvos, line 85: I feel like this projection step should also be inside the parallel loop above, as far as I can see they are also completely independent, and they contain svds so that might actually make a big difference

src/algorithms/grassmann.jl
```diff
@@ -69,22 +70,38 @@ end
...
 
     Rhoreg = Vector{eltype(state.CR)}(undef, length(state))
     δmin = sqrt(eps(real(scalartype(state))))
-    for i in 1:length(state)
-        Rhoreg[i] = regularize(state.CR[i], max(norm(g[i]) / 10, δmin))
+    @static if Defaults.parallelize_sites
```
lkdvos, line 88: A similar comment here, can we just make the entire thing run in one single parallel loop over the sites? The regularisation cut-off is the only thing that depends on the other iterations if I'm not mistaken, and maybe we could just take the local norm of the gradient instead of the max

src/algorithms/grassmann.jl
```diff
@@ -147,6 +172,7 @@ end
 Retract a left-canonical MPSMultiline along Grassmann tangent `g` by distance `alpha`.
 """
 function retract(x::ManifoldPoint{<:MPSMultiline}, tg, alpha)
+    #TODO : support parralelize_sites
```
lkdvos, line 175: Isn't this already parallelized now?

src/algorithms/grassmann.jl
```diff
@@ -71,17 +71,19 @@ end
 function ManifoldPoint(state::Union{InfiniteMPS,FiniteMPS}, envs)
     al_d = similar(state.AL)
     @static if Defaults.parallelize_sites
+        g = fill(nothing, length(state))  #not typstable but this won't be a performance issue for now :)
```
lkdvos: Does this even work?
Also, can you not preallocate with a `g = similar(state.AL)`? I think these things should have the same type?

lkdvos: come to think about it, do we even need this al_d separately? I think you can avoid allocating that vector and only keep the projected gradient

Gertian:
It seems like this PR works, and does parallelize the code. 

However, it does re-introduce out-of-memory crashes that julia 1.10 had solved before. I tried to get the cluster managers to update to 1.11 to see if this fixes the problem :)

Apart from that, perhaps the issue might be solved through ``GC.safepoint()`` but I'm not sure how that works.

Gertian:
@lkdvos what is your opinion on re-doing this on the new branch ?
Would it still add something or has this become obsolete ?

PR #181, opened 2024-10-15
Parallel vumpssvd

Gertian:
This PR also makes vumpSVD parallel.

Again tests have been implemented in another PR and I've tested the combined PR locally on my machine :)

### DIFF
--- a/src/algorithms/changebonds/vumpssvd.jl
+++ b/src/algorithms/changebonds/vumpssvd.jl
@@ -77,32 +77,36 @@ function changebonds_n(state::InfiniteMPS, H, alg::VUMPSSvdCut, envs=environment
     return state, envs
 end
 
-function changebonds_parallel_n(state::InfiniteMPS, H, alg::VUMPSSvdCut, envs=environments(state, H))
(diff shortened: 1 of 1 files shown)


lkdvos (commented):
Is this ever a bottleneck? Because this definitely nontrivially changes the algorithm by doing it in parallel, because every bond update is assuming the other sites to be constant, and this implementation breaks that assumption. It might indeed be the case that this does not matter, you're interested in getting the spaces approximately right, but I'm somewhat less confident about this

Gertian:
> Is this ever a bottleneck? Because this definitely nontrivially changes the algorithm by doing it in parallel, because every bond update is assuming the other sites to be constant, and this implementation breaks that assumption. It might indeed be the case that this does not matter, you're interested in getting the spaces approximately right, but I'm somewhat less confident about this

I just had some code running where this took about half a day. And bond dimensions are still growing.

I wouldn't say its the biggest bottleneck but I was parallelizing stuff anyways so I thought, why not...

I agree that this changes the algorithm,  hence my choice to explicitly retain the old algorithm in the non parallel case. 
I don't think the change is bad though, parallel VUMPS also breaks this assumption and still converges eventually. So I don't see any reason why this wouldn't give something reasonable.

PR #182, opened 2024-10-15
Combine pr

Gertian:
In case it's more convenient for the main devs (I guess @lkdvos ), this PR merges my three prior PRs into one.

So in this PR I've combined 
1) parallel GD
2) parallel VumpsSvd
3) tests for n-site unit cells 

Most notably the new tests seemed important for me since the parallel GD and VumpsSvd might only fail when multiple sites are present and the parallel code kicks in...

At the very least this PR allows the reviewer to see that the new code works with the new tests :)

### DIFF
--- a/src/algorithms/changebonds/vumpssvd.jl
+++ b/src/algorithms/changebonds/vumpssvd.jl
@@ -77,32 +77,36 @@ function changebonds_n(state::InfiniteMPS, H, alg::VUMPSSvdCut, envs=environment
     return state, envs
 end
 
-function changebonds_parallel_n(state::InfiniteMPS, H, alg::VUMPSSvdCut, envs=environments(state, H))
+function changebonds_parallel_n(state::InfiniteMPS, H, alg::VUMPSSvdCut,
+                                envs=environments(state, H))
     #prepare the loops we'll have to run over, this is different depending on wether the length of the state is odd or even ! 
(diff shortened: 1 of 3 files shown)


lkdvos:
I'm gonna do my best to get to all of this and look at everything as soon as possible, I'm a bit short on time these days. (Moving houses, traveling and preparing presentations seem to be taken up a lot of time...)
As a small side-note, I wanted to ask how urgent this is. I do agree we want to add is, but we have an upcoming major rewrite in the branch blocktensor2, which is more or less working but requires an updated version of tensorkit first, which can then release blocktensorkit, which then makes that pr mergable. I am a bit hesitant to make many intermediate changes to avoid having to port over too much of the code, but I also see that it is not reasonable to completely stop development in the meantime, so it's all about striking a balance.
Anyways, if I still didn't get to this by end of next week, feel free to ping me again!

Gertian:
Hi,

Thanks for the quick reply.

I wouldn't say its super urgent, but on the other hand, for my usecase with unit cell 32 this should amount to a HUGE speed increase.

I guess I can always use my local branch though. I don't like doing this though.

Gertian:
@lkdvos seems like you've already reviewed the other PRs so this is irrelevant now.

As you suggested the best way forward is probably to first merge the tests and then the other two PRs !

PR #189, opened 2024-11-13
excise `RecursiveVec`

lkdvos:


Gertian:
@lkdvos, out of interest, what used to be the motivation behind the RecursiveVec ? 
And why did you get rid of it ?

lkdvos:
It's just a vector of vectors, that you want to think about as the direct sum of these components, such that addition is handled recursively, scalar multiplication distributes and the inner product is the sum of the inner products. Since KrylovKit moved to supporting VectorInterface, this is actually automatically handled for arrays, so the extra layer is redundant and you can just use a vector or tuple of vectors.

Gertian:
Thank you for explaining !

Issue #103, opened 2023-12-05
`virtualspace` conventions can be confusing


lkdvos:
Currently, the use of `virtualspace` across the package can be a bit confusing, and there are contradicting choices that have been made which further add to the confusion.

The current state of affairs:
- Creating `FiniteMPS` objects, the constructor can take in a vector of virtual spaces which is one shorter than the length of the physical spaces, with optional kwargs for a left and right starting space. Thus, the `i`th element of the input vector corresponds to the virtual space to the **right** of site `i`.
- Creating `InfiniteMPS` objects, there is no trivial left and right starting space, thus the vector of virtual spaces is the same length as the vector fo physical spaces. In this case, the `i`th element of the vector corresponds to the virtual space to the **left** of site `i`.
- Querying the virtual space of an MPS object, we have `left_virtualspace(state, i)` and `right_virtualspace(state, i)`, which return the left (right) virtual space of the bond tensor to the **right** of site `i`.
- `entanglement_spectrum` yields the singular values of the bond matrix to the **right** of the physical site.

Clearly, there are some inconsistencies, as well as `left_virtualspace` and `right_virtualspace` being a bit misleading.

One solution might be to:
1. Update the constructor for the `InfiniteMPS` to reflect that we most often refer to virtual spaces as being to the **right** of a physical site.
2. Introduce a new (exported) method `virtualspace`, which asserts that left- and right- virtualspace to the **right** of a site, for given state, are equal and then returns this. This should only fail if the user updated a (finite?) MPS manually with a changed bond dimension, in which case the (unexported) internal methods `left_virtualspace` and `right_virtualspace` can be used as a  more *expert-user-case*.
3. Reflect everything clearly in the documentation and check if there are any more cases where this is relevant. In all cases, we adopt the convention that virtualspaces are defined to the **right** of a site index.

maartenvd:
I think I started with virtualspace referring to the right virtualspace. However in hindsight, it is more natural to make this refer to the left virtualspace instead. If you write down a finite mps, then the first virtual space you encounter is the left one - but this would require rewriting quite a bit of code, with very little gain...

Jutho:
If you also want the left and rightmost virtual spaces, there are N+1 bonds for N sites. In that case, I would rather enumerate the bonds from 0 to N than from 1 to N+1. 

Another factor is to have it consistent with how C is labeled? I assume it is also `AL(n) - C(n) - AR(n+1)`?

[...]

PR #205, opened 2024-12-12
update `left` and `right_virtualspace`

lkdvos:
This is a fix of the long-standing issue #103, which rewrites the package to consistently use `left_virtualspace` to denote the space to the left of a lattice site, and `right_virtualspace` to denote the space to the right of a lattice site. Additionally, they are both returned "non-dualised", i.e. `left_virtualspace(mps, i + 1) == right_virtualspace(mps, i)`.

This is obviously very breaking, but should be more consistent in the future.

### DIFF
--- a/src/MPSKit.jl
+++ b/src/MPSKit.jl
@@ -60,8 +60,6 @@ export exact_diagonalization
 export TransferMatrix
 export transfer_left, transfer_right
 
-@deprecate virtualspace left_virtualspace # there is a possible ambiguity when C isn't square, necessitating specifying left or right virtualspace
-
 # Abstract type defs
 abstract type Algorithm end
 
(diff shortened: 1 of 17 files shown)


maartenvd:
how do you garantuee left_virtualspace(mps, i + 1) == right_virtualspace(mps, i) during twosite algorithms?

lkdvos:
Sorry, my comment above was a bit misleading. What I meant is `isdual(left_virtualspace(mps, i+1)) == isdual(right_virtualspace(mps, i))`, and in most cases (infinite and onesite finite), the equality is strict. In other cases, `left_virtualspace(mps, i + 1)` will give you the space to the right of the bond between `i` and `i+1`, while `right_virtualspace(mps, i)` gives you the space to the left of that bond. In other words, `i` now always refers to sites.
As a small remark, it turns out you actually were not really using these functions anyways, and simply calling `space` on the relevant tensors most of the time, so this has less big of an effect than I initially expected (although I honestly won't be surprised if some bugs still pop up).

Issue #214, opened 2024-12-17
Incorrect indexing of Hamiltonian


Gertian:
Hi,

I think I've ran into another bug. When calling ``environments(state, H)`` for an ``InfiniteMPOHamiltonian`` with non constant bond dimension. 
```julia
julia> environments(initial_state, Ham)



ERROR: TaskFailedException

    nested task error: BoundsError: attempt to access 23×1×1×20 SumSpaceIndices{GradedSpace{FermionParity, Tuple{Int64, Int64}},2,2} at index [1:21, 1, 1, 21]
    Stacktrace:
     [1] throw_boundserror(A::SumSpaceIndices{GradedSpace{…},2,2}, I::Tuple{UnitRange{…}, Int64, Int64, Int64})
       @ Base ./essentials.jl:14
     [2] checkbounds
       @ ./abstractarray.jl:699 [inlined]
     [3] _getindex
       @ ~/.julia/packages/BlockTensorKit/k4PdQ/src/vectorspaces/sumspaceindices.jl:60 [inlined]
     [4] getindex
       @ ./abstractarray.jl:1312 [inlined]
     [5] getindex
       @ ~/.julia/packages/BlockTensorKit/k4PdQ/src/tensors/abstractblocktensor/abstractarray.jl:92 [inlined]
     [6] left_cyclethrough!(index::Int64, GL::PeriodicVector{…}, H::InfiniteMPOHamiltonian{…}, state::InfiniteMPS{…})
       @ MPSKit ~/.julia/packages/MPSKit/Tq56r/src/environments/infinitempohamiltonian_envs.jl:139
     [7] compute_leftenv!(envs::MPSKit.InfiniteMPOHamiltonianEnvironments{…})
       @ MPSKit ~/.julia/packages/MPSKit/Tq56r/src/environments/infinitempohamiltonian_envs.jl:103
     [8] (::MPSKit.var"#197#199"{MPSKit.InfiniteMPOHamiltonianEnvironments{…}})()
       @ MPSKit ~/.julia/packages/MPSKit/Tq56r/src/environments/abstract_envs.jl:101

...and 1 more exception.

Stacktrace:
 [1] sync_end(c::Channel{Any})
   @ Base ./task.jl:466
 [2] macro expansion
   @ ./task.jl:499 [inlined]
 [3] recalculate!(envs::MPSKit.InfiniteMPOHamiltonianEnvironments{…}, state::InfiniteMPS{…}; tol::Float64)
[...]

PR #215, opened 2024-12-18
Enforce InfiniteMPOHamiltonian to be square

lkdvos:
After further consideration, for now the easiest solution is to enforce `InfiniteMPOHamiltonian` to be square. This addresses #214 .


lkdvos:
With the current implementation, it probably makes no difference.

PR #216, opened 2024-12-19
Rewrite MPS types to use `C` instead of `CR` (and `CL`) and add ` ψ.center` functionality.

VictorVanthilt:
This PR changes the name of MPS bondtensors. The bond tensor for all MPS types is now called `C` and is equivalent to what was previously called `CR`.

A new property "center" of `FiniteMPS` is introduced. It can be accessed through `ψ.center`.

It will return a `HalfInteger` `i`:
- `isinteger(i)` → `i` is a whole number and indicates the location of the first `AC` tensor present in `ψ.ACs`
- `ishalfodd(i)` → `i` is a half-odd-integer (e.g. 3/2), meaning that there are no `AC` tensors, and indicating between which sites the bond tensor lives.
For example: `i` = 3/2 (= 1.5) means that there lives a `C` between the `AL` on site 1 and the `AR` on site 2

The goal is to use this new `center` property to avoid unnecessary gaugefixing in methods like `norm(::FiniteMPS)` where before, the center gauge was brought to the first site, after which the norm was calculated. After this PR the norm is calculated using the (bond) tensor at `ψ.center`, without gaugefixing.

An important note is that `ψ.C[center(ψ)] `does not work, `ψ.C` should be indexed using integers and `ψ.C[i]` returns the bond tensor between sites i and i+1


Gertian:
It surely was confusing. Essentially I looked into the code to see what it returned whenever relevant...
Anyways, good docs are indeed sufficient.

For the front end user half indexing might be very handy and user friendly though. If you and @lkdvos are open to this I can make a PR. Imo both can just coexist :)

src/states/orthoview.jl
```diff
@@ -155,31 +155,39 @@ function Base.setindex!(v::ACView{<:Multiline}, vec, i::Int, j::Int)
...
 end
 
+# function Base.getindex(A::AbstractArray, hi::HalfInteger)
+#     if isinteger(hi)
+#         return getindex(A, Int(hi))
+#     else
+#         throw(ArgumentError("Indexing a $(typeof(A)) with half-odd-integer indices is not supported"))
+#     end
+# end
+
```
lkdvos, line 165: This can be removed :)

src/states/finitemps.jl
```diff
@@ -129,13 +133,44 @@ function Base.getproperty(ψ::FiniteMPS, prop::Symbol)
...
-    elseif prop == :CR
-        return CRView(ψ)
+    elseif prop == :C
+        return CView(ψ)
+    elseif prop == :center
+        return _gaugecenter(ψ)
     else
         return getfield(ψ, prop)
     end
 end
```
lkdvos, line 143: Probably it also makes sense to add `Base.propertynames` to reflect our implementation:
https://docs.julialang.org/en/v1/base/base/#Base.propertynames

lkdvos: One other thing is that it might be reasonable to not actually pass on the `ACs` etc to `getfield`, and instead require that these are called manually through `getfield`, in order to reflect that they are considered internal

VictorVanthilt: `Base.propertynames` is now implemented.
Right now the docstring for `FiniteMPS` only shows the properties, not the fields. If people want to access the underlying `.ACs`, they know they exist and therefore probably know what they are doing.

To be honest, only being able to access \`.ACs\` via `getfield` will make for lots of interal changes to the code and will make it less readable imo.

lkdvos (approved):

PR #218, opened 2024-12-20
Add `FiniteMPS` `getindex` 

VictorVanthilt:
This PR implements `getindex` for `FiniteMPS` using both `Integer`s and `AbstractUnitRange`s.

### DIFF
--- a/src/states/finitemps.jl
+++ b/src/states/finitemps.jl
@@ -286,7 +286,7 @@ Base.@propagate_inbounds function Base.getindex(ψ::FiniteMPS, i::Int)
     c = ψ.center
 
     @boundscheck begin
-        (1 <= i <= length(ψ)) || throw(BoundsError(ψ, i))
+        checkbounds(ψ, i)
     end
 
     if isinteger(c)
--- a/src/states/finitemps.jl
+++ b/src/states/finitemps.jl
@@ -307,6 +307,11 @@ Base.@propagate_inbounds function Base.getindex(ψ::FiniteMPS, i::Int)
     end
 end
 
+Base.@propagate_inbounds function Base.getindex(ψ::FiniteMPS, I::AbstractUnitRange)
+    return Base.getindex.(Ref(ψ), I)
+end
+
+
 function Base.convert(::Type{TensorMap}, ψ::FiniteMPS)
     T = foldl(ψ.AR[2:end]; init=first(ψ.AC)) do x, y
         return _transpose_front(x * _transpose_tail(y))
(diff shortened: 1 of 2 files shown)


VictorVanthilt:
@lkdvos can you look at the implementation as well as if the `Base.@propagate_inbounds` magic would work with this? I haven't worked with this before.

src/states/finitemps.jl
```diff
@@ -282,6 +282,35 @@ end
 
 Base.checkbounds(::Type{Bool}, ψ::FiniteMPS, i::Integer) = 1 <= i <= length(ψ)
 
+Base.@propagate_inbounds function Base.getindex(ψ::FiniteMPS, i::Int)
+    c = ψ.center
+
+    @boundscheck begin
+        (1 <= i <= length(ψ)) || throw(BoundsError(ψ, i))
```
lkdvos, line 289:
```suggestion
        checkbounds(ψ, i)
```
(this is defined just above here)

src/states/finitemps.jl
```diff
@@ -282,6 +282,35 @@ end
...
+    end
+
+    if isinteger(c)
+        return if i > Int(c)
+            ψ.AR[i]
```
lkdvos, line 294: Because you are checking the bounds above, in principle you should be allowed to put an `@inbounds` here. Honestly, I wouldn't bother with it, since we are not guaranteed that these arrays are the actual sizes they should be, in which case this would lead to segfaults. This should never really be called in super tight loops, so this is fine as is.

src/states/finitemps.jl
```diff
@@ -282,6 +282,35 @@ end
...
+        c = Int(c - 1 / 2)
+        return if i > Int(c)
+            ψ.AR[i]
+        else
+            ψ.AL[i]
+        end
+    end
+end
+
+Base.@propagate_inbounds function Base.getindex(ψ::FiniteMPS, I::AbstractUnitRange)
```
lkdvos, line 310:
```suggestion
@inline function Base.getindex(ψ::FiniteMPS, I::AbstractUnitRange)
```

lkdvos (changes requested):
Looks okay to me!

For the `boundscheck` business: the long story short (ish) is that `@inbounds` will bypass code that is marked within a `@boundscheck` block, on the condition that it is inlined, or passed through a function barrier with `@propagate_inbounds`.

Boundschecking itself happens by the function `checkbounds`, which has two signatures: `checkbounds(Bool, args...)` and `checkbounds(args...)`. The first outputs a bool to check the bounds, the second calls the first and throws if necessary.

In general, the boundschecking really only matters for performance-critical sections, or expensive boundschecks. I like putting an extra check in the function, as this gives clearer error messages, but putting `@inbounds` should only be used as a last resort, since that can lead to segfaults.

Could you try and add some tests for this, testing inferred, proper error throwing etc? Nothing too major, but just a few lines to ensure everything is (and keeps) functioning property would be great.

lkdvos (approved):

PR #220, opened 2025-01-03
Revert TaylorCluster Optimizations of #199

VictorVanthilt:


lkdvos (changes requested):
Looks good to me, I would say good to go if we add some tests?

lkdvos (approved):

VictorVanthilt:
Any idea why the julia 1 test on macos is giving LAPACK errors?

PR #221, opened 2025-01-06
Pass appropriate `tol_min` in default dynamic tolerances

leburgel:
I randomly noticed that all the `DynamicTol` calls in the `Defaults` module passed the algorithm `tol` as the `tol_min` of the `DynamicTol` struct instead of the actual `tol_min` kwarg. Looked like a typo to me, so thought I should submit a fix in case it is.


lkdvos (approved):
Good catch, thanks!

PR #222, opened 2025-01-07
make `tol` a kwarg in `remove_orphans!(::FiniteMPO)`

VictorVanthilt:
`tol` was a kwarg in `remove_orphans!` for `InfiniteMPO`s and a regular argument with a default value in `remove_orphans!` for `FiniteMPO`s. This PR unifies the calling signature of the two methods.

### DIFF
--- a/src/operators/mpo.jl
+++ b/src/operators/mpo.jl
@@ -81,7 +81,7 @@ function remove_orphans!(mpo::InfiniteMPO; tol=eps(real(scalartype(mpo)))^(3 / 4
     return mpo
 end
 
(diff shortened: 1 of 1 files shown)


lkdvos:
I'll concede on the name, it's fair to say that `droptol!` typically does not change the spaces/sizes.

The failing tests are actually unrelated to this, we updated the compat on TensorKitManifolds, such that now the tests are actually using TensorKit v0.14 (which they previously weren't). I'll look into fixing that hopefully this afternoon

lkdvos:
Can be merged as soon as #223 is merged.

lkdvos (approved):

PR #224, opened 2025-01-08
Refactor `InfiniteEnvironments` and `GradientGrassmann`

lkdvos:
This PR consists of a number of changes that ultimately boil down to removing the automated recomputation of the infinite environments.

In order to achieve this, a number of additional changes are also included:
- most notably, I finally got around to cleaning up the grassmann implementation, which now uses the preconditioner in OptimKit. This *GREATLY* improves the clarity, and should hopefully be a nice change for future development.
- all infinite environments are now unified
- various utility changes
- various preparatory changes for having `above` and `below` different.

PR #225, opened 2025-01-12
Minor improvements

lkdvos:
Some small changes and improvements, mostly to remove unused code and make Aqua happy (happier at least).

There is still a large source of ambiguities related to the following:
- `environments` -> hopefully improved with #224 but should be checked afterwards
- `LazySum` -> having it a subtype of `AbstractVector` and defining `*(::LazySum, ::Any)` and related is a problem
- `FiniteMPS` constructors seem to be a mess

And piracies related to defining methods for type aliases such as `MPSTensor`


VictorVanthilt:
These look good to me, can you explain the reasoning for adding the whole `_not_missing_type` thing?

VictorVanthilt (commented):
These look good to me, can you explain the reasoning for adding the whole `_not_missing_type` thing?

lkdvos:
It's an unbound type arguments problem: since `Vector{Missing} <: Vector{T} where {T<:Union{Missing,A}}` would also satisfy the signature, `A` would then not be defined. Aqua complains about that, even though realistically you would never call this with a vector of missing type. In order to get around this, I need to distill `A` from the eltype, but it's not easy since you cannot dispatch really dispatch on union types too well.

PR #227, opened 2025-01-16
Set default `tol_max` in gauging eigsolver

leburgel:
Sets the default maximal tolerance in the solver used in the gauging eigsolve steps. I was having stability issues, where for a slowly converging orthogonalization the eigsolve step used a tolerance so high that it was messing up convergence altogether since the error kept increasing after each inaccurate eigsolve.

I don't have a minimal working example, but I think restricting the `tol_max` here can't hurt either way. Setting it at `Inf` seems a bit dangerous...

### DIFF
--- a/src/states/ortho.jl
+++ b/src/states/ortho.jl
@@ -2,7 +2,7 @@
 # ----------
 
 const _GAUGE_ALG_EIGSOLVE = DynamicTol(Arnoldi(; krylovdim=30, eager=true),
-                                       Defaults.tolgauge / 10, Inf, 1)
+                                       Defaults.tolgauge / 10, Defaults.tol_max, 1)
 
 """
     struct LeftCanonical <: Algorithm


lkdvos (approved):
Yeah, that's probably way more sensible.

I'm okay with merging this as is, but if you have the time it would be even better to actually move this to the `Defaults` section.

leburgel:
> I'm okay with merging this as is, but if you have the time it would be even better to actually move this to the `Defaults` section.

I could just move the `_GAUGE_ALG_EIGSOLVE` definition to the `Defaults` module, rename it and then call that. But I think it's actually easier to just call `Defaults.alg_eigsolve` with the appropriate kwargs. I'll give that a go.

leburgel:
I think it makes sense to still define a constant within `ortho.jl` but use `Defaults.alg_eigsolve` to construct it. In the end the only non-default part was `ishermitian=false` and `tol_factor=1`, so it's a bit cleaner.

It didn't make sense to me to move this specific definition to `Defaults` and then call it from there. I don't think this would be cleaner, and I don't see a more general use case for these specific settings.

lkdvos (approved):

PR #229, opened 2025-01-20
Avoid adding adding exact zeros in `TaylorCluster` implementation

leburgel:
The addition of `SparseBlockTensorMap` slices in the `TaylorCluster` implementation lead to the explicit instantiation and subsequent purging of explicit zero blocks due to the `getindex`-`setindex!` interplay in expressions like `slice[linds[s_c], 1, 1, :] += slice[c_lin, 1, 1, :]`. This PR circumvents this by explicitly restricting to the nonzero keys of the right hand side. It's not the cleanest solution, but it gives a big performance increase for Hamiltonians with a lot of channels.


lkdvos (approved):
Test failures seem unrelated, merging this.

PR #230, opened 2025-01-21
Last updates before v0.12 tag

lkdvos:
I tried to give the docs a final push to try and get this released tomorrow.
It can locally be run using https://github.com/QuantumKitHub/MPSKitModels.jl/pull/42, which cannot be released before tagging this so we will have to do some post-release fixing of the docs.

While it's obvious that the docs are far from finished, I don't think it is worth it to wait for them to release this. For me, this is thus good to go.


VictorVanthilt (approved):
Big fan of the `IDMRG1` rename.
The rest of the changes look good.
Having documentation on environments is something that is very necessary in my opinion as this is something I've heard users complain about not being clear at all. It is however, not essential for the release of 0.12 so all is good to go.

leburgel (approved):
Looks great! I just added a few minor things:
- The `makedocs` call threw an error, seemed like a type in the doctest kwarg and was fixed by changing `doctests -> doctest`
- There was a missing finite lattice in the Hamiltonian construction in the  `ChepigaAnsatz` example, causing it to fail. I added it so it runs and the (local) build is fixed, but it seems it doesn't actually give the expected result?
- There was a missing docstring for `fidelity_susceptibility`, so I added this to the best of my abilities
- The `exact_diagonalization` docstring was not very useful, so I tried to expand it a bit
- Fixed some typos and made some minor cosmetic tweaks
- Added the return arguments to the function signature in the docstrings of the main algorithms. This is something that I keep forgetting, so I thought might as well add it now

PR #244, opened 2025-02-03
Relax MPO element type restrictions

leburgel:
Attempt at relaxing the type restrictions on MPO elements, in order to be able to use generic effective MPOs. This should work as long as the element type adheres to the (until now implicit) interface for `MPOTensor` which requires:

- `left_virtualspace`
- `right_virtualspace`
- `_conj_mpo`
- `add_physical_charge`
- `_fuse_mpo_mpo`
- `_fuse_mpo_mps`
- `_mpo_to_mps`
- ...?

I also added explicit type annotations for internal methods where an element type `<:MPOTensor` is really assumed.

See also QuantumKitHub/PEPSKit.jl#125 for the intended use case.

### DIFF
--- a/src/operators/mpo.jl
+++ b/src/operators/mpo.jl
@@ -1,24 +1,24 @@
 """
-    struct MPO{O<:MPOTensor,V<:AbstractVector{O}} <: AbstractMPO{O}
+    struct MPO{O,V<:AbstractVector{O}} <: AbstractMPO{O}
 
 Matrix Product Operator (MPO) acting on a tensor product space with a linear order.
 
 See also: [`FiniteMPO`](@ref), [`InfiniteMPO`](@ref)
 """
-struct MPO{TO<:MPOTensor,V<:AbstractVector{TO}} <: AbstractMPO{TO}
+struct MPO{TO,V<:AbstractVector{TO}} <: AbstractMPO{TO}
     O::V
 end
 
 """
-    FiniteMPO(Os::Vector{<:MPOTensor}) -> FiniteMPO
-    FiniteMPO(O::AbstractTensorMap{S,N,N}) where {S,N} -> FiniteMPO
+    FiniteMPO(Os::Vector{O}) -> FiniteMPO{O}
+    FiniteMPO(O::AbstractTensorMap{S,N,N}) where {S,N} -> FiniteMPO{O<:MPOTensor}
 
 Matrix Product Operator (MPO) acting on a finite tensor product space with a linear order.
 """
-const FiniteMPO{O<:MPOTensor} = MPO{O,Vector{O}}
+const FiniteMPO{O} = MPO{O,Vector{O}}
 Base.isfinite(::Type{<:FiniteMPO}) = true
 
-function FiniteMPO(Os::AbstractVector{O}) where {O<:MPOTensor}
+function FiniteMPO(Os::AbstractVector{O}) where {O}
     for i in eachindex(Os)[1:(end - 1)]
         right_virtualspace(Os[i]) == left_virtualspace(Os[i + 1]) ||
             throw(SpaceMismatch("unmatching virtual spaces at site $i"))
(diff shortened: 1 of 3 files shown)


src/operators/mpo.jl
```diff
@@ -277,25 +277,18 @@ function Base.:*(mpo1::InfiniteMPO, mpo2::InfiniteMPO)
...
-                                        O2[1 2; -3 3] *
-                                        O1[4 -2; 2 5] *
-                                        conj(Fᵣ[-4; 3 5])
-end
-
 # TODO: I think the fastest order is to start from both ends, and take the overlap at the
 # largest virtual space cut, but it might be better to just multithread both sides and meet
 # in the middle
-function TensorKit.dot(bra::FiniteMPS{T}, mpo::FiniteMPO, ket::FiniteMPS{T}) where {T}
+function TensorKit.dot(bra::FiniteMPS{T}, mpo::FiniteMPO{<:MPOTensor},
```
lkdvos, line 290: I guess if you replace the `similar(...)` calls with the correct `allocate_GL`/`allocate_GR`, this is again agnostic to the eltype?

src/operators/multilinempo.jl
```diff
@@ -13,10 +13,10 @@ See also: [`Multiline`](@ref), [`AbstractMPO`](@ref)
 """
 const MultilineMPO = Multiline{<:AbstractMPO}
 
-function MultilineMPO(Os::AbstractMatrix{T}) where {T<:MPOTensor}
+function MultilineMPO(Os::AbstractMatrix{T}) where {T}
     return MultilineMPO(map(FiniteMPO, eachrow(Os)))
 end
-function MultilineMPO(Os::PeriodicMatrix{T}) where {T<:MPOTensor}
+function MultilineMPO(Os::PeriodicMatrix{T}) where {T}
```
lkdvos, line 19:
```suggestion
function MultilineMPO(Os::PeriodicMatrix)
```

src/operators/multilinempo.jl
```diff
@@ -13,10 +13,10 @@ See also: [`Multiline`](@ref), [`AbstractMPO`](@ref)
 """
 const MultilineMPO = Multiline{<:AbstractMPO}
 
-function MultilineMPO(Os::AbstractMatrix{T}) where {T<:MPOTensor}
+function MultilineMPO(Os::AbstractMatrix{T}) where {T}
```
lkdvos, line 16:
```suggestion
function MultilineMPO(Os::AbstractMatrix)
```

lkdvos (commented):
Overall, looks clean to me. Can you confirm that the PEPSKit changes you want have passing tests before I merge and tag any of this?

src/operators/mpo.jl
```diff
@@ -207,7 +207,7 @@ function VectorInterface.scale!(mpo::MPO, α::Number)
     return mpo
 end
 
-function Base.:*(mpo1::FiniteMPO{TO}, mpo2::FiniteMPO{TO}) where {TO}
+function Base.:*(mpo1::FiniteMPO{TO}, mpo2::FiniteMPO{TO}) where {TO<:MPOTensor}
```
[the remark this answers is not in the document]
leburgel: Added a TODO so I don't forget. I think it is the same map, but the warnings when the auxiliary legs at the start and end are not the `oneunit` are specific to the finite case. I can replace the loop in the finite implementation by the same `fuse_mul_mpo` map, but did you mean really fully merging them? Then we would need to add some kind of edge space check that does nothing for an infinite MPO and checks for oneunits in the finite case, which also feels a bit hacky.

src/operators/mpo.jl
leburgel: I only found one `similar` I could replace (I also think the `bra` and `ket` were switched in the `similar` call there, so I think that's a fortunate accidental fix). Did you have any others in mind that I maybe missed?

lkdvos: Yeah sorry my comment was badly placed. You got the one I meant, sadly we don't currently have a `isomorphism!` implementation so the other `rho` instantiations will have to remain as they are.

lkdvos (approved):

PR #245, opened 2025-02-05
change default site value for `FiniteMPS` in `entanglement_spectrum`

VictorVanthilt:
As pointed out by @borisdevos, calling `entanglement_spectrum(::FiniteMPS)` does not work as the default value for `site` is 0, which would call the C tensor left of site 0, which does exist, it is just a trivial 1, but throws a `BoundsError` due to the checkbounds in `entanglement_spectrum`.

Currently I just split off the `entanglement_spectrum` method that dispatched on `FiniteMPS` to have a default `site` value of 1. This works but now the docstring is not consistent as it shows a default value of 0 for both methods.

In my opinion having a default value for the finite case really does not make sense and the user should just choose the site according to their needs.

Any opinions?

### DIFF
--- a/src/algorithms/toolbox.jl
+++ b/src/algorithms/toolbox.jl
@@ -92,7 +92,11 @@ Compute the entanglement spectrum at a given site, i.e. the singular values of t
 matrix to the right of a given site. This is a dictionary mapping the charge to the singular
 values.
 """
-function entanglement_spectrum(st::Union{InfiniteMPS,FiniteMPS,WindowMPS}, site::Int=0)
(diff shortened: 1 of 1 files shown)


lkdvos:
This is definitely ok for me, and removing the default value for the FiniteMPS also seems reasonable.
It might also be a good idea to check if we are using `site` as a kwarg or a regular argument in other parts of the code, at least for me I know that this often confuses me

VictorVanthilt:
@lkdvos What do you think of this implementation? (The one I just pushed.)

lkdvos (approved):

Issue #246, opened 2025-02-05
`entanglementplot` and `transferplot` give `MethodError` for `InfiniteMPS` and `FiniteMPS`


borisdevos:
`entanglementplot` and `transferplot` of an `InfiniteMPS` and `FiniteMPS` give a `MethodError`. Here's a MWE for `InfiniteMPS`:

```julia
julia> using TensorKit, MPSKit

julia> test = InfiniteMPS(rand, ComplexF64, ComplexSpace(2), ComplexSpace(16))

julia> entanglementplot(test)
ERROR: MethodError: no method matching plot(::MPSKit.EntanglementPlot)
Stacktrace:
 [1] entanglementplot(args::InfiniteMPS{TensorMap{…}, TensorMap{…}}; kw::@Kwargs{})
   @ MPSKit C:\Users\bodvos\.julia\packages\RecipesBase\BRe07\src\RecipesBase.jl:380
 [2] top-level scope
   @ REPL[3]:1
```
Similarly:

```julia
julia> transferplot(test)
ERROR: MethodError: no method matching plot(::MPSKit.TransferPlot)
Stacktrace:
 [1] transferplot(args::InfiniteMPS{TensorMap{…}, TensorMap{…}}; kw::@Kwargs{})
   @ MPSKit C:\Users\bodvos\.julia\packages\RecipesBase\BRe07\src\RecipesBase.jl:380
 [2] top-level scope
   @ REPL[4]:1
```
The first error is the exact same when constructing a `FiniteMPS`. I'm using TensorKit@v0.14.3 and MPSKit@v0.12.3 for clarity.

leburgel:
I think just adding `using Plots` should fix this. MPSKit relies on [RecipesBase.jl](https://github.com/JuliaPlots/Plots.jl/tree/v2/RecipesBase) to define how to make these kind of plots, but doesn't actually depend on any package that can *produce* these plots. For that, you need to load Plots.jl.

I do agree it's very confusing to have an exported method that you can't run without importing another package, but it was probably a very deliberate choice to do it this way. Is there a clean way to fix this, or at least get a better error if users forget to import Plots.JL @lkdvos?

borisdevos:
You're right! I've moved on from Plots.jl and started using CairoMakie.jl, so this didn't occur to me. Maybe as a longer-term thing, we could consider moving over to Makie, since this seems to be the more popular plotting package nowadays. I'm willing to look into this if this is wanted!

[...]

PR #247, opened 2025-02-06
Update plotting docstrings

leburgel:
Updates the plotting docstrings to state that users need to add `using Plots` themselves before being able to produce any plots.

See #246.


lkdvos (commented):
Would it be possible to make a slightly more clear separation between arguments and keyword arguments? With them listed like this, you need to have noticed that there is a semicolon in the signature, which seems a bit easy to miss

leburgel:
> Would it be possible to make a slightly more clear separation between arguments and keyword arguments?

Done, and also updated the signature itself (the `transferplot` one didn't even have the semicolon...).

lkdvos (approved):

PR #248, opened 2025-02-06
Compute `expval` using precomputed data

lkdvos:
Instead of computing the expectation value in the middle of the chain, we now attempt to be a bit more smart and actually figure out where the precomputed data is.


leburgel (approved):
Looks good to me!

PR #252, opened 2025-02-18
Implement cluster expansion MPOs

VictorVanthilt:
This is my initial implementation for the "Symmetric cluster expansions" introduced in  https://doi.org/10.1103/PhysRevA.103.L020402
They are currently only implemented for infinite systems.

The current implementation is still undergoing testing and suffers from a rather large compilation time.

Things that still need to be done:
- [ ] Find the reason for and potentially improve compilation time.
- [ ] Thoroughly test their performance
- [ ] Implement tests
- [ ] Finite MPOHamiltonian compatibility


VictorVanthilt:
Perhaps also using @generated in the making of the apply functions could help speed up the first compilation ...

src/algorithms/timestep/clusterexpansion.jl
```diff
@@ -0,0 +1,164 @@
...
+    N₁ = numin(Tr)
+    N = numind(Tr)
+    t_out = tensorexpr(:new_e_r, -(1:(N₁ + 1)), -((N₁ + 2):(N + 2)))
+    t_left = tensorexpr(:Or, -(1:2), (-(N₁ + 2), 1))
+    t_right = tensorexpr(:Tr, (1, (-(3:(N₁ + 1))...)), -((N₁ + 3):(N + 2)))
+    return macroexpand(@__MODULE__,
+                       :(return @tensor $t_out := $t_left * $t_right))
+end
+
+function _make_apply_functions(O::SparseBlockTensorMap, n::Int)
```
lkdvos, line 115: I'm not entirely sure returning an anonymous function is a great idea here. If you could refactor this into a callable struct, or any struct that you make work with KrylovKit.jl would avoid that you have to recompile these functions every time you call this.

src/algorithms/timestep/clusterexpansion.jl
```diff
@@ -0,0 +1,164 @@
...
+    nT = (n - 1) ÷ 2
+    Al, Ar = make_envs(O, n)
+
+    fun = if iseven(n)
```
lkdvos, line 119: Splitting on the value of `n` is again very type-unstable...

src/algorithms/timestep/clusterexpansion.jl
```diff
@@ -0,0 +1,164 @@
+struct ClusterExpansion
+    N::Int
+end
+
+function make_time_mpo(H::MPOHamiltonian, dt::Number, alg::ClusterExpansion)
+    N = alg.N
+    lmax = N ÷ 2 # largest level
+    τ = -im * dt
+    # spaces
+    P = physicalspace(H)[1]
```
lkdvos, line 10:
```suggestion
    P = physicalspace(H, 1)
```

src/algorithms/timestep/clusterexpansion.jl
```diff
@@ -0,0 +1,164 @@
...
+    D = dim(physicalspace(H)[1]) # physical dimension
```
lkdvos, line 11:
```suggestion
    D = dim(P) # physical dimension
```

src/algorithms/timestep/clusterexpansion.jl
```diff
@@ -0,0 +1,164 @@
...
+    V = BlockTensorKit.oplus([ℂ^(D^2l) for l in 0:lmax]...)
+
+    TT = tensormaptype(ComplexSpace, 2, 2, ComplexF64)
```
lkdvos, line 14:
```suggestion
    TT = tensormaptype(ComplexSpace, 2, 2, complex(scalartype(H)))
```

src/algorithms/timestep/clusterexpansion.jl
```diff
@@ -0,0 +1,164 @@
...
+            (info.converged) != 1 &&
+                @warn "Did not converge when constucting the $n cluster"
+
+            if isodd(n) # linear problem
+                # assign the new level to O
+                O[jnl, 1, 1, jnl] = Onew
+            elseif iseven(n) # linear problem + svd
+                U, S, V = tsvd(Onew, (1, 2, 4), (3, 5, 6))
+                # assign the new levels to O
+                O[jnl - 1, 1, 1, jnl] = permute(U * sqrt(S), ((1, 2), (3, 4)))
```
lkdvos, line 40:
```suggestion
           @plansor O[jnl - 1, 1, 1, jnl][-1 -2; -3 -4] := U[-1 -2 -3; 1] * sqrt(S)[1, -4]
```
Writing this with `@plansor` calls makes it compatible with fermions, and can help with allocations in the future.

src/algorithms/timestep/clusterexpansion.jl
```diff
@@ -0,0 +1,164 @@
...
+                             Tuple(vcat([2n + 2], collect((n + 2):(2n + 1)))))),
+                 Tuple(1:(n + 1)), Tuple(vcat(collect((n + 3):(2n + 2)), [n + 2])))
+    bo = _fully_contract_O(O, n)
+    return be - bo
+end
+
+function _fully_contract_O(O::SparseBlockTensorMap, n::Int)
+    # Make projector onto the 0 subspace of the virtual level
+    Pl = zeros(ComplexF64, ℂ^1 ← space(O, 1))
+    Pl[1] = 1
```
lkdvos, line 67: This pattern seems wrong (and probably will break once I actually get to fixing that), and probably only works for non-symmetric tensors... If I get this right, you want a trivial unit tensor as the first element, so you should probably just instantiate that tensor. There are some examples of this scattered throughout MPSKit 😉

lkdvos (commented):
Keep in mind that `@generated` tends to **increase** compilation time, since it has to compile new functions for all types.
Having a quick glance at this, it seems to me that the compile time is this large mostly because a lot of the functions are very type-unstable. One thing that might help is to use `Val(N)` whenever you are referring to the lengths of the finite mpos. Since you are instantiating them as contracted tensors, this yields `AbstractTensorMap{T,S,N,N}` or similar objects, and becomes a problem when `N` is not known at compile time.
(which coincidentally also requires more `@generated` functions since `ncon` is also not type stable 😁 )

This being said, am I right in looking at this and thinking the main missing components in terms of contractions is the instantiation or application of `FiniteMPO` with static lengths?
If this is the case, we might just create `SFiniteMPO` in analogy to `SVector` (StaticArrays.jl), and implement this generically? This has additional benefits as well, since the derivatives would fit in this same framework.

PR #253, opened 2025-02-19
Allow for non-square / uniform `MPOHamiltonian`s in `make_time_mpo(..., ::TaylorCluster)`

VictorVanthilt:


src/algorithms/timestep/timeevmpo.jl
```diff
@@ -59,38 +59,17 @@ function make_time_mpo(H::MPOHamiltonian, dt::Number, alg::TaylorCluster;
...
-    end
-
-    # Check if mpo has the same size everywhere. This is assumed in the following.
-    @assert allequal(size.(H′)) "make_time_mpo assumes all mpo tensors to have equal size. A fix for this is yet to be implemented"
-
     # start with H^N
-    H_n = H′^N
-    V = size(H′[1], 1)
+    H_n = H^N
+    V = size(H[1], 1)
```
lkdvos, line 64: I don't think this is enough: for non-square things we need to account for things that are considered level "1" and "3" differently for each site

lkdvos:
I made an attempt at fixing this, I have something that seems to be running, and at least locally produces the correct scaling behavior. Comparing the 2nd order infinite result though, it seems like we obtain a 6x6 matrix for a nearest-neighbour model, which does not match the 3x3 that the paper shows (appendix A).
Honestly, I would love to just write down and compare to these exact expressions as a testcase, I also don't think the tests you added are working...

lkdvos (approved):

PR #262, opened 2025-03-14
[New] Finite temperature utilities + docs example

lkdvos:
Here, I added some additional support for working with DMRG and TDVP for density matrices. I added a rudimentary example to show off the abilities, which could do with a lot of cleanup but I am running out of reasonable time to spend on this.

I'd be okay with merging this as is and finishing it up later, unless someone wants to go for it now.


lkdvos:
https://quantumkithub.github.io/MPSKit.jl/previews/PR262/

examples/quantum1d/7.xy-finiteT/main.jl
```diff
@@ -0,0 +1,471 @@
...
+Z(\beta) =
+    \text{Tr} \left( e^{-\beta H} \right) =
+    \text{Tr} \left( e^{-\beta H / 2} e^{-\beta H / 2} \right) =
+    \left\langle e^{-\beta H^\dagger / 2}, e^{-\beta H / 2} \right\rangle
+```
+
+In other words, we can compute the partition function at $\beta$ by computing the overlap of two states evolved for $\beta / 2$, as long as the Hamiltonian is Hermitian.
+Otherwise, we could still use the same trick, but we would have to compute the evolved states twice, once for $H$ and once for $H^\dagger$.
+
+(TODO) show figure of this trick.
```
leburgel, line 241: `!!! todo`

examples/quantum1d/7.xy-finiteT/main.jl
```diff
@@ -0,0 +1,471 @@
...
+    p1 = plot(βs, abs.(real.(Z_taylor2) .- partition_function.(βs, J, N));
+              label=labels, title="Partition function error",
+              xlabel="β", ylabel="ΔZ(β)")
+    plot!(p1, βs, abs.(real.(Z_mpo_mul) .- partition_function.(βs, J, N));
+          label="MPO multiplication")
+    p2 = plot(βs, abs.(real.(F_taylor2) .- free_energy.(βs, J, N)); label=labels,
+              xlabel="β", ylabel="ΔF(β)", title="Free energy error")
+    plot!(p2, βs, abs.(real.(F_mpo_mul) .- free_energy.(βs, J, N));
+          label="MPO multiplication")
+    plot(p1, p2)
```
leburgel, line 326: The legend in the second plot is positioned quite unfortunately. It would be better to move this to the northwest corner.

examples/quantum1d/7.xy-finiteT/main.jl
```diff
@@ -0,0 +1,471 @@
...
+On the other hand, if we improve the accuracy of the initial density matrix, we could also increase the step size, which would reduce the number of iterations required to reach a certain $\beta$ value.
+Keeping these parameters in balance is necessary to obtain accurate results, and this might require some trial and error.
+"""
+
+md"""
+### MPO multiplication approach (exponential)
+
+If we wish to push the results to even larger $\beta$ values, we can note that taking linear steps in $\beta$ is not the only option.
+To that end, we can use another trick to scan over an exponential range of $\beta$ values: [exponentiating by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring).
+In particular, we note that computing $x^n$ for integer (large) $n$ can typically be done more efficiently than computing $x * x * ... * x$.
```
leburgel, line 350:
```suggestion
In particular, we note that computing $x^n$ for integer (large) $n$ can typically be done more efficiently than computing $x \cdot x \cdot \dots \cdot x$.
```

examples/quantum1d/7.xy-finiteT/main.jl
```diff
@@ -0,0 +1,471 @@
...
+To do so, we note that multiplication is associative, and regroup the factors in such a way that we can compute the result in a logarithmic number of steps.
+Here, we assume $n = 2^m$ for some integer $m$, and note that this could be generalized to any $n$ by decomposing $n$ into a sum of powers of $2$.
+Then, we can write
+
+```math
+x^n = x^{2^m} = x^{2^{m-1}} \cdot x^{2^{m-1}} = (x^{2^{m-2}} \cdot x^{2^{m-2}}) \cdot (x^{2^{m-2}} \cdot x^{2^{m-2}}) = \dots
+```
+
+In other words, we can scan an exponential range of $\beta$ values by squaring the density matrix at each step.
```
leburgel, line 359: The phrasing 'scan an exponential range' is not really clear to me.

examples/quantum1d/7.xy-finiteT/main.jl
```diff
@@ -0,0 +1,471 @@
...
+    plot!(p1, βs_exp, abs.(real.(Z_mpo_mul_exp) .- partition_function.(βs_exp, J, N));
+          label="MPO multiplication exp")
+
+    p2 = plot(βs, abs.(real.(F_taylor2) .- free_energy.(βs, J, N)); label=labels,
+              xlabel="β", ylabel="ΔF(β)", title="Free energy error")
+    plot!(p2, βs, abs.(real.(F_mpo_mul) .- free_energy.(βs, J, N));
+          label="MPO multiplication")
+    plot!(p2, βs_exp, abs.(real.(F_mpo_mul_exp) .- free_energy.(βs_exp, J, N));
+          label="MPO multiplication exp")
+    plot(p1, p2)
```
leburgel, line 400: Same comment about the legend in the right panel, would be better in the northwest corner.

examples/quantum1d/7.xy-finiteT/main.jl
```diff
@@ -0,0 +1,471 @@
...
+### Time evolution approach
+
+Finally, we can also note that the partition function is characterized by the following differential equation:
+
+```math
+\frac{dZ}{d\beta} = -H \cdot Z
+\implies Z(\beta) = e^{-\beta H} \cdot Z(0)
+```
+
+In other words, we can compute the partition function at $\beta$ by evolving the partition function at $0$ for a time $d\tau = -im \beta$.
```
leburgel, line 422:
```suggestion
In other words, we can compute the partition function at $\beta$ by evolving the partition function at $0$ for a time $d\tau = -i \beta$.
```
Latex does not speak Julia :)

leburgel (commented):
Just merging this as-is without much cleanup is definitely fine by me. It's a strict addition and the example is extremely nice!

However, I'm not super enthusiastic about merging in the actual implementation into the main package without tests or explanation. The general utilities like the MPO trace and the infinite temperature density matrix are great, but it would be nice if we had a test (quickly adapted from the example?). For the implementation of the derivatives and the transfer functions, a test would be nice, but also some clarification of the implementation itself would be good I think. It's implicitly assumed that given an MPO tensor $O$ we apply $O \otimes \mathbb{1}$. Also the fact that the second physical index corresponds to the bra-side of a density matrix and the fact that the physical indices are traced shut is implicit, while this is really a choice that was made. I'm totally fine with the choices, I just think they should be briefly but.explicitly stated somewhere.

examples/quantum1d/6.hubbard/main.jl
```diff
@@ -76,8 +76,8 @@ function compute_groundstate(psi, H;
     ## convergence steps
     psi, = changebonds(psi, H, SvdCut(; trscheme=truncbelow(svalue)))
     psi, = find_groundstate(psi, H,
-                            VUMPS(; tol=svalue, verbosity) &
-                            GradientGrassmann(; tol=svalue / 100, verbosity))
+                            VUMPS(; tol=svalue / 100, verbosity) &
+                            GradientGrassmann(; tol=svalue / 1000, verbosity))
```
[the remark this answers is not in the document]
leburgel: As long as this fixes things that's fine. The zero step sizes are indeed very annoying, but the solution to this will have to come from OptimKit.

examples/quantum1d/7.xy-finiteT/main.jl
```diff
@@ -0,0 +1,471 @@
...
+
+md"""
+### Exact diagonalization
+
+We can check our results by comparing them to the exact diagonalization of the Hamiltonian.
+"""
+
+H = periodic_boundary_conditions(XY_hamiltonian(; J, N=Inf), 6)
+H_dense = convert(TensorMap, H);
+vals = eigvals(H_dense)[Trivial()] ./ 6
```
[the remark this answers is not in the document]
leburgel: Just checking if this was intentional, but the reasoning sounds good to me so let's leave it like this.

leburgel (approved):
Looks perfect!

Qiaoyi-Li:
I am also working on MPO-based finite-T calculations, so I took a look at the example script. The comments in it are quite insightful---many of the observations align with challenges we have encountered as well. Some tricks may be helpful, listed below:
- Using a bilayer MPO $\rho(\beta/2)$ to represent the density operator $\rho(\beta)$ indeed improves accuracy. If you are interested,  some benchmark results can be found in [PhysRevB.95.144428(2017)](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.95.144428).
- The "MPO multiplication approach (exponential)" (we call it "XTRG"[PhysRevX.8.031082(2018)](https://journals.aps.org/prx/abstract/10.1103/PhysRevX.8.031082)) is a robust technical route, as if the initialized high-T state is accurate enough.  Initializing a more accurate high-T state([PhysRevB.95.161104(2017)](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.95.161104) may be helpful) may help reduce  errors?
- Thanks to the TDVP, the time-evolution approach definitely improves the finite-T calculations with long-range interactions (as the key step is TDVP, a tangent-space technique, we call it "tanTRG"[PhysRevLett.130.226502(2023)](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.130.226502)). In our initial benchmark tests, we found directly performing imaginary-time evolution with constant step length from an identity MPO leads to large free energy errors at low temperatures. You might try a linearly-increasing step length (equivalent to an exponential scaled $\beta$ list) in the initial stage, which works well in all models we have computed.

lkdvos:
Hi Qiaoyi-Li, thanks for your comments and suggestions! It is really nice to hear some of the tips and tricks from someone that is actually working with this, since often the technical tricks get lost in the relevant papers.

- I agree that it would be better to have a fully bilayer approach. However, as I'm not really working on this topic myself, I lack the time to do so. As always, I would be more than happy to collaborate on this if you would like, and contributions are more than welcome.
- I am aware of that work for writing initial high-T states. However, I'm not convinced that this would increase accuracy. The initial state presented here is based on [Efficient higher-order matrix product operators](https://scipost.org/SciPostPhys.17.5.135), which provide MPOs that are fully correct up to the specified order. The work you link involves truncations, which suffer from the same problem as all the methods presented here: interpreting an MPO as an MPS, and then truncating it is a truncation in the frobenius norm of the MPS. For states, this is indeed the relevant norm, while for operators we would really want the operator norm instead. In that sense, if one only cares about the accuracy of the initial state, I would argue that it is better to take the Taylor expansion up to the maximal order that fits in memory, and then making the initial beta smaller until the desired accuracy is reached. This aside, I would love to see an actual comparison of these two in practice, since there is always a trade-off between accuracy and CPU time and memory usage, and for concrete models one might be interested in different results.
- For the algorithm names, I didn't see enough reason to change `TDVP` and `DMRG` to something else. The core algorithm really is exactly the same as `TDVP` and `DMRG`, and the only alteration is the specific contraction of the effective local operator. Therefore I think it is fair to use the same name, and also convenient, since otherwise I would have to duplicate a whole bunch of code, whereas now I simply modified the contractions. Obviously I have no problem linking these papers, somehow I just kind of forgot to put the citations in. If you want to, I'd be more than happy if you could add these performance tips and tricks and the references in the example.

Qiaoyi-Li:
Hi Lukas, I ran some quick tests using my own package (since I am more familiar with it than MPSKit) and uploaded the benchmark script into [https://github.com/Qiaoyi-Li/FiniteMPS.jl/tree/main/benchmark/XY_chain](https://github.com/Qiaoyi-Li/FiniteMPS.jl/tree/main/benchmark/XY_chain).
- In all three test cases, even the worst one (with linearly increasing $\beta$) resulted in free energy errors of at most $\lesssim 10^{-5}$. However, when running the example script in MPSKit, I observed much larger errors. I’m not sure if there might be some undetected bugs in it?
- Exponentially increasing $\beta$ significantly reduces errors. In fact, we proposed this choice in the maintext of [PhysRevLett.130.226502(2023)](https://doi.org/10.1103/PhysRevLett.130.226502). 
- The XY chain might be too simple to reveal the benefits of SETTN initialization, making it nonessential in this case.  A more complex test (4-leg spinless fermions) is provided in Fig.7 of [SM](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.130.226502#supplemental), in where higher order expansion can further reduce errors in the initial stage.
- Since I haven’t implemented XTRG in my package, I cannot provide a script to convince you.  I suspect that truncating via Frobenius norm is not a problem just because XTRG finally works in many cases.  I would be interested in seeing benchmark results from MPSKit if you have any.
- I do not have any comment on the names of algorithms, please feel free to use any function names you prefer in your package.

lkdvos:
Thanks for checking this: indeed there was a mistake. I was using analytic results for periodic boundary conditions, leading to small discrepancies.
I'm rerunning the examples and now simply basing of your (very nice!) BenchmarkFreeFermions results, to avoid these issues.

Also, I'm not saying these things are not working, I just mean that one should be a little careful about it. The frobenius norm is still a norm, so as long as the truncations aren't too severe this should still be fine. I just want to be careful advertising things in the examples so users are not entering with too high expectations: we are still solving an exponential problem, and we know that it shouldn't be possible to deal with arbitrary finite temperatures since there is no area law (I think)?

Anyways, thanks a lot for your tips and comments, this has been very helpful so I appreciate this greatly!

PR #265, opened 2025-03-20
Refactor VUMPS and VOMPS to avoid boxed variables 

lkdvos:
I wanted to include OhMyThreads v0.8, which now throws errors for boxed variables in closures.
Since I had wanted to improve the code reuse a bit anyways, I went ahead and fully refactored VUMPS and VOMPS to make better use of the iterative solvers approach.

- How do we feel about `dominant_eigsolve`, the internal function I now used to map both `leading_boundary` and `find_groundstate` to? It definitely avoids some code duplication in the different VUMPS implementations. Naming suggestions also welcome of course.
- Is this new approach clearer/cleaner to everyone? I find it a bit easier to follow, and it definitely splits the big VUMPS functions into smaller parts so it should also be easier on the compiler etc.
- @leburgel, do you know if this affects the boundary MPS stuff in PEPSKit?


VictorVanthilt (commented):
- The `dominant_eigsolve` name is self-explanatory and I can't come up with a better name so I'm okay with it.
- The usage of `Base.iterate` in an iterative solver is indeed a readability improvement, it's way more clear what steps the different algorithms entail.

Could you point out where Boxed variables were showing up in closures? I want to be able to avoid ruining performance when using closures in the future.

lkdvos:
I can, but this really does not "ruin" performance at all, you want to avoid this in cases where your iterations take ~1ms or less, sure, but for VUMPS that's really just not the case.

The boxed variables appear whenever you are capturing a variable in a closure, but also modifying it outside of the closure. The exact mechanism of when it does and doesn't happen is also not fully clear to me, but it tends to be a combination of scoping, closures, and reassignment that does the trick. Here are some examples to give an idea:

```julia
let
    A = zeros(10)
    for _ in 1:10
        res = tmap(eachindex(A)) do i
            A[i] = i
            A[i]
        end
        A = zeros(10)
    end
end # A boxed

let
    A = zeros(10)
    for _ in 1:10
        A = zeros(10)
        res = tmap(eachindex(A)) do i
            A[i] = i
            A[i]
        end
    end
end # A boxed


let
    for _ in 1:10
        A = zeros(10)
        res = tmap(eachindex(A)) do i
            A[i] = i
            A[i]
        end
    end
end # fine
```

Note that the easiest way to fix this, which you might have seen a couple times in different places, is to introduce a new `let` block, where now the name of the variable in the inner scope is no longer tied to the same one as the outer scope, and therefore the closure doesn't need to box the variable anymore:

```julia
let
    A = zeros(10)
    for _ in 1:10
        res = let A = A
            tmap(eachindex(A)) do i
                A[i] = i
                A[i]
            end
        end
        A = zeros(10)
    end
end # A boxed
```

VictorVanthilt (approved):

PR #266, opened 2025-03-20
Fix finite temperature example

lkdvos:
Thanks to @Qiaoyi-Li for pointing out a mistake in the finite T example: I was comparing open boundary MPS with periodic boundary analytic results. Should be resolved now


lkdvos:
https://quantumkithub.github.io/MPSKit.jl/previews/PR266/

PR #268, opened 2025-03-30
Change `environments` calling syntax for inifinite MPS

### LINKED ISSUE #267: A bug?
The approximate function throw an ERROR

```
using TensorKit, MPSKit
s1, s2, s3, s4 = Z2Space(0 => 1, 1 => 1), Z2Space(0 => 2, 1 => 2), Z2Space(0 => 3, 1 => 3), Z2Space(0 => 4, 1 => 4)
Sop = TensorMap(randn, s2⊗s1 ← s1⊗s2)
mpo = InfiniteMPO([Sop])
state = InfiniteMPS([s1], [s3])
state2 = InfiniteMPS([s1], [s4])
approximate(state2,(mpo,state),IDMRG())
```

`ERROR: SpaceMismatch("ProductSpace(Rep[ℤ₂](0=>3, 1=>3)) ≠ ProductSpace(Rep[ℤ₂](0=>4, 1=>4))")
`

leburgel:
Changes the syntax convention for infinite MPS environment-related methods to `(below, operator, above=below)`, in line with the finite MPS convention. Since `above` is always the optional one from the point of view of boundary MPS algorithms, this made more sense. I didn't change any of the `TransferMatrix` related code, which still assumes `(above, operator, below=above)` for its arguments. I suspect changing this would cause a lot more downstream issues.

In the process I noticed that `calc_galerkin` would have been wrong when used in `approximate` for finite MPS, and after that noticed that we don't actually use the proper Galerkin error in the finite approximate. I think we could, and probably should? But this can be addressed in a follow-up I think.

Fixes #267.

### DIFF
--- a/src/algorithms/toolbox.jl
+++ b/src/algorithms/toolbox.jl
@@ -36,8 +36,8 @@ function infinite_temperature_density_matrix(H::MPOHamiltonian)
 end
 
 """
-    calc_galerkin(above, operator, below, envs)
-    calc_galerkin(pos, above, operator, below, envs)
+    calc_galerkin(below, operator, above, envs)
+    calc_galerkin(pos, below, operator, above, envs)
 
 Calculate the Galerkin error, which is the error between the solution of the original problem, and the solution of the problem projected on the tangent space.
 Concretely, this is the overlap of the current state with the single-site derivative, projected onto the nullspace of the current state:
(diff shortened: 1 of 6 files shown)


src/algorithms/toolbox.jl
```diff
@@ -46,25 +46,25 @@ Concretely, this is the overlap of the current state with the single-site deriva
 \\epsilon = |VL * (VL' * \\frac{above}{\\partial AC_{pos}})|
 ```
 """
-function calc_galerkin(pos::Int, above::Union{InfiniteMPS,FiniteMPS,WindowMPS}, operator,
-                       below, envs)
-    AC´ = ∂∂AC(pos, above, operator, envs) * above.AC[pos]
+function calc_galerkin(pos::Int, below::Union{InfiniteMPS,FiniteMPS,WindowMPS}, operator,
+                       above, envs)
+    AC´ = ∂∂AC(pos, below, operator, envs) * above.AC[pos]
```
lkdvos, line 51: Good catch here, indeed this was previously just wrong whenever `below !== above`.

lkdvos (approved):
Great changes, thanks!
Indeed, when I previously looked at this I realized there were definitely some issues but didn't feel like untangling that mess, but this looks relatively contained.

Slightly off-topic, but I wanted to share my view anyways:
For the transfer stuff, I'm a bit on the fence whether or not we should also change that (definitely not in this PR). Especially for the actual structs it seems like `top middle bottom` is not a strange order of the fields.
I think I would for that case prefer to just introduce a function `transfermatrix` that has similar syntax to our other functions which returns the object, and as usual the fields of that object are considered implementation details so that should work out.

PR #270, opened 2025-03-31
Refactor derivatives and add `JordanMPOTensor`

lkdvos:
This is quite a large refactor of the `derivatives` objects.

In particular, I rewrote the hamiltonian derivatives to make use of the particular Jordan block structure, which leads to *very* large speedups, especially for nearest-neighbor Hamiltonians.

While on the topic, I also immediately specialized a blocktensor type for this Jordan block structure, mostly so this information is stored at the type level, and we don't have to assume that the MPOTensors have the correct properties.

Finally, I also noticed that the `SVD` implementation is a **lot** slower than the `SDD` one, so I made this accessible from the relevant algorithms, and restored the default to `SDD` everywhere but in the Grassmann implementation. This might lead to spurious LAPACK errors again, but at least this can now be resolved by changing the settings.

In principle this could do with some more documentation, but this might be left for a dedicated docs update session...


src/algorithms/timestep/taylorcluster.jl
```diff
@@ -78,7 +78,7 @@ function make_time_mpo(H::MPOHamiltonian, dt::Number, alg::TaylorCluster;
 
     # loopback step: Algorithm 1
     # constructing the Nth order time evolution MPO
-    mpo = MPO(parent(H_n))
+    mpo = MPO(map(SparseBlockTensorMap, parent(H_n)))
```
VictorVanthilt, line 81: Just because I'm curious, what problem arose that lead to this change?

src/operators/jordanmpotensor.jl
```diff
@@ -0,0 +1,345 @@
...
+function BlockTensorKit.nonzero_length(W::JordanMPOTensor)
+    return nonzero_length(W.A) + nonzero_length(W.B) + nonzero_length(W.C) +
+           nonzero_length(W.D) + Int(size(W, 1) > 1) + Int(size(W, 4) > 1)
+end
+
+# linalg
+# ------
+# do we want this?
+function Base.:+(W1::JordanMPOTensor, W2::JordanMPOTensor)
+    return SparseBlockTensorMap(W1) + SparseBlockTensorMap(W2)
```
VictorVanthilt, line 281: if W1 and W2 both have a two-site interaction term, do they just get added? Leading to (C1 + C2)*(B1 + B2) = C1B1 + C2B2 + incorrect cross terms. Or do the interactions get their own rows and columns to make sure this doesn't happen?

src/operators/jordanmpotensor.jl
```diff
@@ -0,0 +1,345 @@
...
+end
+function Base.:-(W1::JordanMPOTensor, W2::JordanMPOTensor)
```
VictorVanthilt, line 283: Same comment here as with `Base.:+`

src/operators/mpohamiltonian.jl
```diff
@@ -739,46 +716,24 @@ Base.:-(H::MPOHamiltonian, λs::AbstractVector{<:Number}) = H + (-λs)
...
-    foreach(enumerate(parent(H))) do (i, h)
-        if i != length(H)
-            scale!(h[1, 1, 1, 2:end], λ) # multiply top row (except BraidingTensor)
-        else
-            scale!(h[1, 1, 1, end], λ) # multiply right column (except BraidingTensor)
-        end
+function VectorInterface.scale!(H::MPOHamiltonian{O},
+                                λ::Number) where {O<:JordanMPOTensor}
+    for i in 1:length(H)
+        scale!(H[i].C, λ)
```
VictorVanthilt, line 722: This is so much simpler than before! Great.

src/utility/defaults.jl
```diff
@@ -8,6 +8,7 @@ module Defaults
 import KrylovKit: GMRES, Arnoldi, Lanczos
 using OhMyThreads
 using ..MPSKit: DynamicTol
+using TensorKit: TensorKit
```
VictorVanthilt, line 11: What does this do?

VictorVanthilt (approved):
Great addition to the code! I like the new drawings of the derivative operators, they make it way easier to spot which of the operators you need and what you are exactly calculating without having to reverse-engineer the contraction using the @plansor call.

Is there a definitive answer as to how much faster the new JordanMPO derivatives make things like DMRG?

src/operators/jordanmpotensor.jl
lkdvos: Yes and no: this definition isn't actually used for adding MPOs at all, you can find that [here](https://github.com/QuantumKitHub/MPSKit.jl/blob/03d2a208a3c51fb1a371ea7caec9ca0fbba00043/src/operators/mpohamiltonian.jl#L675) which was also more or less how this was before: summing individual tensors does exactly that, and it is your responsibility to not sum individual tensors when you actually want to sum MPOs

src/utility/defaults.jl
lkdvos: This brings `TensorKit` (and only that name) into the namespace, which is used for the default svd algorithm: `TensorKit.SDD()`. In principle I could have also written `using TensorKit: SDD`

src/algorithms/timestep/taylorcluster.jl
lkdvos: Here we go from `MPOHamiltonian` to `MPO`, which means that we no longer are limited to the Jordan form, in particular we also need to add elements below the diagonal.

src/operators/jordanmpotensor.jl
VictorVanthilt: But then why include this method and risk people not knowing this is wrong?

VictorVanthilt: I would include `Base.:+` but make it the right implementation like in the code you linked. And then use `Base.:+` there. But it's a minor thing, proceed as you wish!

VictorVanthilt: Agreed, let's leave it then.

src/operators/jordanmpotensor.jl
```diff
@@ -0,0 +1,345 @@
...
+        return haskey(W.B, CartesianIndex(row - 1, 1, 1))
+    elseif 1 < row < sz[1] && 1 < col < sz[4]
+        return haskey(W.A, CartesianIndex(row - 1, 1, 1, col - 1))
+    else
+        return false
+    end
+end
+
+# TODO: avoid this slow fallback wherever possible:
+Base.parent(W::JordanMPOTensor) = parent(SparseBlockTensorMap(W))
```
leburgel, line 150: Where would this be used now?

leburgel (commented):
Looks great! Left some comments, but nothing major.

src/algorithms/toolbox.jl
```diff
@@ -363,6 +363,8 @@ function periodic_boundary_conditions(H::InfiniteMPOHamiltonian, L=length(H))
                 F_right = fusers[site + 1][k, i, l]
                 j′ = indmap[j, i, l]
                 k′ = indmap[k, i, l]
+                ((j′ == 1 && k′ == 1) ||
+                 (j′ == size(output[site], 1) && k′ == size(output[site], 4))) && continue
```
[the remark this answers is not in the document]
lkdvos: Not really: these are the braiding tensors on the first and last entry of the diagonal of the resulting MPO, which now implicitly have `BraidingTensor`s instead of explicitly. (because of the `JordanMPOTensor`). These contractions don't actually generate a `BraidingTensor` however, but rather a tensormap which represents a braiding tensor. I could go in and verify that if you are setting `W[1, 1, 1, 1] = t` the `t` is (approximately) a braidingtensor, but that requires choosing a tolerance, and first computing the result and then discarding it, so I just hijacked this here and not computed them in the first place.

src/operators/jordanmpotensor.jl
lkdvos: Many of the `AbstractBlockTensorMap` implementations that are "array-like" call out to this function. I've tried to avoid most of them by overloading faster fallbacks, but sometimes this is more or less the most efficient way. I'll just delete the TODO for now, should not be an issue if it doesn't show up in the profiler.

src/algorithms/changebonds/svdcut.jl
```diff
@@ -7,7 +7,10 @@ An algorithm that uses truncated SVD to change the bond dimension of a ψ.
 
 $(TYPEDFIELDS)
 """
-@kwdef struct SvdCut <: Algorithm
+@kwdef struct SvdCut{S} <: Algorithm
+    "algorithm used for the singular value decomposition"
+    alg_svd::S = Defaults.alg_svd()
+
     "algorithm used for [truncation][@extref TensorKit.tsvd] of the gauge tensors"
     trscheme::TruncationScheme = notrunc()
```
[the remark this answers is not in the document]
lkdvos: Definitely wasn't on purpose. I'm okay with either option to be honest, `notrunc` still seems opinionated

src/algorithms/approximate/idmrg.jl
```diff
@@ -90,10 +90,8 @@ function approximate!(ψ::MultilineMPS, toapprox::Tuple{<:MultilineMPO,<:Multili
                 for row in 1:size(ψ, 1)
                     # TODO: also write this as ac2_proj?
                     AC2 = ϕ.AL[row, col] * _transpose_tail(ϕ.AC[row, col + 1])
-                    AC2′ = ∂AC2(AC2, O[row, col], O[row, col + 1],
-                                leftenv(envs[row], col, ψ[row]),
-                                rightenv(envs[row], col, ψ[row]))
-                    al, c, ar, = tsvd!(AC2′; trunc=alg.trscheme, alg=TensorKit.SVD())
+                    AC2′ = ∂∂AC2(row, col, ψ, O, envs) * AC2
```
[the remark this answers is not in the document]
lkdvos: I can't in the current implementation: `ac2_proj` makes `AC2` by contracting `AC * AR`, here I really need `AL * AC`.
I also strongly dislike this, but for now I'm okay with just keeping this with the todo...

src/algorithms/derivatives/derivatives.jl
```diff
@@ -0,0 +1,258 @@
...
+ │     │ 
+┌┴┐   ┌┴┐
+│ ├───┤ │
+└┬┘   └┬┘
+ │     │ 
+ └─   ─┘ 
+```
+
+See also [`∂C`](@ref).
+""" ∂∂C
```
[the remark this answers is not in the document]
lkdvos: Honestly, I am very happy to change these names to something better, or at the very least have the unicode things as an alias.
I'am happy to go for `c_derivative` etc, but that sounds a bit like the single `∂` functions to me, and these are really just `c_derivative_operator`, which might be too long?

While we're on the topic, I also much rather would not have the `row, col` syntax and have these as a single `CartesianIndex{2}`, to keep the number of arguments consistent.
Similarly, I would also much rather have the consistent `c_derivative(site, below, operator, above, envs)`.

However, let's keep that for a separate PR, since this thing is already a monster?

src/algorithms/derivatives/mpo_derivatives.jl
```diff
@@ -0,0 +1,56 @@
...
+    V_r = right_virtualspace(H.rightenv)
+    V_o = prod(physicalspace, H.O; init=one(V_l))
+    return V_l ⊗ V_o ⊗ V_r
+end
+
+# Actions
+# -------
+(h::MPO_∂∂C)(x) = ∂C(x, h.leftenv, h.rightenv)
+(h::MPO_∂∂AC)(x) = ∂AC(x, h.operators[1], h.leftenv, h.rightenv)
+(h::MPO_∂∂AC2)(x) = ∂AC2(x, h.operators[1], h.operators[2], h.leftenv, h.rightenv)
```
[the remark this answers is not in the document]
lkdvos: I definitely agree, I was just wondering if anyone ever wanted to have these functions without having to construct the operator first.
(also, specializing on the type parameters of the `MPO_∂∂C` looks a bit more messy)

I will reverse the dependency order and have `∂C` etc call the mpo implementation by default.

src/algorithms/approximate/idmrg.jl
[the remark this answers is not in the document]
leburgel: Okay, as long as it's deliberate it's fine for me.

src/algorithms/changebonds/svdcut.jl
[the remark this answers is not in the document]
leburgel: Then let's maybe make it mandatory, that way they all are and it's consistent again?

src/operators/jordanmpotensor.jl
```diff
@@ -0,0 +1,345 @@
...
+                                              A::SparseBlockTensorMap{TA,E,S,2,2},
+                                              B::SparseBlockTensorMap{TB,E,S,2,1},
+                                              C::SparseBlockTensorMap{TC,E,S,1,2},
+                                              D::SparseBlockTensorMap{TD,E,S,1,1}) where {E,
+                                                                                          S,
+                                                                                          TA,
+                                                                                          TB,
+                                                                                          TC,
+                                                                                          TD}
+        # TODO: add space and size checks
```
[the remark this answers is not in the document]
lkdvos: I forgot to delete this todo: the spacechecks are in the outer constructor without the type parameters.

src/algorithms/derivatives/derivatives.jl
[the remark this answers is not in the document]
leburgel: Agreed to keep this for a follow-up then, specifically targeting the name and calling signature for the local effective Hamiltonians.

src/algorithms/toolbox.jl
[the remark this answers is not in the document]
leburgel: I see, I didn't realize the `BraidingTensor`s were now implicit. Sounds like a solid fix to me!

src/operators/abstractmpo.jl
```diff
@@ -203,6 +203,11 @@ end
 
 function _fuse_mpo_mpo(O1::MPOTensor, O2::MPOTensor, Fₗ, Fᵣ)
     return if O1 isa BraidingTensor && O2 isa BraidingTensor
+        # shouldn't happen
```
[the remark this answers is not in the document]
leburgel: My thought was that if we somehow introduce a bug at a higher level, having another fallback implemented here might hide that since it would then start using this without any errors or warnings. But that's probably overthinking things, if this works and isn't dangerously slow there's not really an issue.

leburgel (approved):

PR #271, opened 2025-04-04
Change `sector` convention for excitations

lkdvos:
Change `sector` convention to be more intuitive, in particular allowing for:

```julia
eigvals(convert(TensorMap, H))[sector] == exact_diagonalization(H; sector)
```

and equivalent results for the excitations.

## to do

- [x] ED
- [x] excitations
- [x] `add_physical_charge`
- [x] rebase on #272

## Sidenote

While doing this I discovered that the excitation code was type unstable for some reasons, which I also resolved, #272 should be merged first


src/algorithms/ED.jl
```diff
@@ -31,47 +31,51 @@ equivalent to dense eigenvectors.
...
 """
 function exact_diagonalization(H::FiniteMPOHamiltonian;
-                               sector=first(sectors(oneunit(physicalspace(H, 1)))),
-                               len::Int=length(H), num::Int=1, which::Symbol=:SR,
+                               sector=one(sectortype(H)),
+                               num::Int=1, which::Symbol=:SR,
                                alg=Defaults.alg_eigsolve(; dynamic_tols=false))
-    left = ℂ[typeof(sector)](sector => 1)
-    right = oneunit(left)
+    L = length(H)
```
leburgel, line 37: Happy to see the inexplicable potential `len` mismatch gone :)

src/algorithms/ED.jl
```diff
@@ -31,47 +31,51 @@ equivalent to dense eigenvectors.
 
 """
 function exact_diagonalization(H::FiniteMPOHamiltonian;
```
leburgel, line 33: We should really add a test for this if we export it. Not really the point of this PR, but we should probably add it to the v0.13 checklist.

src/states/finitemps.jl
```diff
@@ -234,13 +234,13 @@ function FiniteMPS(f, elt, Pspaces::Vector{<:Union{S,CompositeSpace{S}}},
 
     Vspaces[1] = left
     for k in 2:N
-        Vspaces[k] = infimum(fuse(Vspaces[k - 1], fusedPspaces[k]), maxVspaces[k - 1])
+        Vspaces[k] = infimum(fuse(Vspaces[k - 1], fusedPspaces[k - 1]), maxVspaces[k - 1])
```
leburgel, line 237: Was this a bug then?

src/states/finitemps.jl
```diff
@@ -383,18 +383,36 @@ function TensorKit.space(ψ::FiniteMPS{<:GenericMPSTensor}, n::Integer)
     return ProductSpace{S}(space.(Ref(t), Base.front(Base.tail(TensorKit.allind(t)))))
 end
 
+"""
+    max_virtualspaces(ψ::FiniteMPS)
+    max_virtualspaces(Ps::Vector{<:Union{S,CompositeSpace{S}}}; left=oneunit(S), right=oneunit(S))
+
+Compute the maximal virtual spaces of a given MPS or its spaces.
```
leburgel, line 390:
```suggestion
Compute the maximal virtual spaces of a given finite MPS or its physical spaces.
```

src/states/finitemps.jl
```diff
@@ -383,18 +383,36 @@ function TensorKit.space(ψ::FiniteMPS{<:GenericMPSTensor}, n::Integer)
...
+"""
+function max_virtualspaces(Ps::Vector{<:Union{S,CompositeSpace{S}}}; left=oneunit(S),
```
leburgel, line 392: I like the addition, but is seems like this is not covered by the tests at all which is a bit unfortunate.

leburgel (commented):
Looks perfect to me, but it's a bit annoying that really none of the changed parts were covered by the tests to begin with so we're essentially eyeballing to check that things work. The PR itself achieves the goal though, so I would be okay with merging despite the lack of tests and add these in a follow-up.

As a final comment, it might be good to list some necessary downstream changes to account for the changes here while we're at it, just to keep track of things. Is there anything else besides MPSKitModels.jl and PEPSKit.jl?

ZongYongyue:
Hi Lukas,

I really like this part of the improvement, but I encountered an issue when trying to use the new method to construct a charged MPS. When I place one electron on the right side, everything works correctly—both the ground state energy and particle number match expectations. However, when I place two electrons, the results remain the same as with one electron. Could you please help me identify where the issue might be?

I tested this on a Hubbard model with size = 4. The first half of the implementation uses the new way, while the second half uses my old way for benchmarking:
```julia
using TensorKit
using MPSKit
using MPSKitModels:FiniteChain
using DynamicalCorrelators: hubbard, randFiniteMPS

N = 4
elt = Float64

I = FermionParity ⊠ SU2Irrep ⊠ U1Irrep 
pspace = Vect[I]((0, 0, -1) => 1, (1, 1//2, 0) => 1, (0, 0, 1) => 1)
right=Vect[I]((1, 1//2, 1)=> 2)
H = hubbard(Float64, SU2Irrep, U1Irrep, FiniteChain(N); t=1, U=8, μ=0)
st = FiniteMPS(MPSKit.physicalspace(H), MPSKit.max_virtualspaces(MPSKit.physicalspace(H); right)[2:(end - 1)];right)
gs, envs, delta = find_groundstate(st, H, DMRG2(trscheme= truncerr(1e-6)));
E0 = expectation_value(gs, H)
n = TensorMap(zeros, elt, pspace ← pspace)
block(n, I((0,0,1))) .= 2
block(n, I((1,1//2,0))) .= 1
id = isomorphism(n)
num = sum([correlator(gs, MPSKit.add_util_leg(n), MPSKit.add_util_leg(id), 1, 2),
        correlator(gs, MPSKit.add_util_leg(n), MPSKit.add_util_leg(id), 2, 3),
        correlator(gs, MPSKit.add_util_leg(n), MPSKit.add_util_leg(id), 3, 4),
        correlator(gs, MPSKit.add_util_leg(id), MPSKit.add_util_leg(n), 1, 4)])
println("E0: ", real(E0),"  Particle number: ", real(num))

filling = (3,2)
P, Q = filling
st = randFiniteMPS(elt, SU2Irrep, U1Irrep, N; filling=filling)
H = hubbard(elt, SU2Irrep, U1Irrep, FiniteChain(4); filling=filling, t=1, U=8, μ=0)
gs, envs, delta = find_groundstate(st, H, DMRG2(trscheme= truncerr(1e-6)));
E0 = expectation_value(gs, H)
dim.(MPSKit.max_virtualspaces(gs))
pspace = Vect[I]((0,0,-P) => 1, (0,0,2*Q-P) => 1, (1,1//2,Q-P) => 1)
n = TensorMap(zeros, elt, pspace ← pspace)
block(n, I((0,0,2*Q-P))) .= 2
block(n, I((1,1//2,Q-P))) .= 1
id = isomorphism(n)
num = sum([correlator(gs, MPSKit.add_util_leg(n), MPSKit.add_util_leg(id), 1, 2),
        correlator(gs, MPSKit.add_util_leg(n), MPSKit.add_util_leg(id), 2, 3),
        correlator(gs, MPSKit.add_util_leg(n), MPSKit.add_util_leg(id), 3, 4),
        correlator(gs, MPSKit.add_util_leg(id), MPSKit.add_util_leg(n), 1, 4)])
println("E0: ", real(E0),"  Particle number: ", real(num))

```
the results:
```julia
[ Info: DMRG2   1:      obj = +5.791932749013e+00       err = 8.7722753525e-01  time = 0.02 sec
[ Info: DMRG2 conv 2:   obj = +5.791932749013e+00       err = 1.3322676296e-15  time = 0.03 sec
E0: 5.7919327490129024  Particle number: 5.0000000000000275
[ Info: DMRG2   1:      obj = +1.353907857862e+01       err = 5.2638077148e-01  time = 0.01 sec
[ Info: DMRG2 conv 2:   obj = +1.353907857862e+01       err = 1.5543122345e-15  time = 0.01 sec
E0: 13.539078578622963  Particle number: 6.00000000000003
```

lkdvos:
This is definitely an interesting question, it might actually be worth it to make this a separate issue just to keep it easily accessible in the future.

For the reply, I think the issue is `right=Vect[I]((1, 1//2, 1)=> 2)` is not entirely what you are thinking: this is an auxiliary space that determines the charge, adding degeneracies to it does not actually alter the charge, you could think about this as now searching for a degenerate subspace instead of a eigenvector. (although the algorithms will probably fail)

What you probably want is `right = Vect[I]((0, 0, 2) => 1))` or `right = Vect[I]((0, 1, 2) => 1)` I think, where this latter one would be the sector of two electrons paired as a triplet. In other words, you want to use the charge of the auxiliary leg, not use the degeneracy.

lkdvos:
I don't think you can combine more than one charge and degeneracy together. What I mean is that you can have a single `right = Vect[I]((i, j, k) => 1)` and optimize for that, and have to do a separate optimization for a separate sector.

The picture I usually have in my mind is the following: imagining a operator `H`, this can be fully diagonalized as `H * V = V * D` with `V` the square matrix of eigenvectors.
Then, a single eigenvector is obtained by slicing a single column out of `V`, so this would be a `N x 1` matrix. (I'm explicitly keeping the `1`-dimension here because of what comes next).

If `H` was symmetric, it is actually block-diagonal with the blocks labeled by the different sectors.
Similarly, `V` becomes block diagonal, and you can label the eigenvectors with a sector, so the `N x 1` eigenvector actually carries a sector on the `1` as well.
This is precisely the auxiliary space we are adding to the MPS, which is used to select which sector we want to target.
It has to have a single charge and degeneracy 1, since otherwise it would no longer be an eigenvector, but rather an eigenspace that you are targeting (similar to selecting multiple columns from `V`).
This is not something that our algorithms are set up to handle, and this degeneracy would very likely break DMRG and VUMPS etc, since effectively you would be looking for multiple solutions at the same time, while the algorithms typically assume a unique solution.

This is at least how I try to internalize this, I hope this helps!

ZongYongyue:
I see——thank you very much for your clear and easy-to-understand explanation!

PR #274, opened 2025-04-07
Add tests for fermionic periodic boundary conditions

lkdvos:
Implement fixes for handling fermionic periodic boundary conditions and adds some tests.

I have to say that this still absolutely confuses me, and I can't really wrap my head around this very well.
The bottom line is that if I understand everything correctly, even though our formalism works, it cannot correctly account for the anticommutation relations on different sites:

```julia
using MPSKitModels
h_ij = c_plusmin() + c_minplus()
h_ji = permute(h_ij, ((2, 1), (4, 3)))
h_ij ≈ -h_ji # false!!
h_ij ≈ h_ji # true
```

I don't think there is any way around this, but I might be mistaken here.

In any case, with this in mind, the previous implementation of PBC did not account for this, so here I added a twist on every link that needs to wrap around.

I think the tests should capture this correctly, but I'm still confused by this entire endeavor, since if I am understanding this correctly, for fermionic systems the following does not hold, while it is what I understand from translation invariance for bosonic systems:

```julia
H = # some tensor representing a translation invariant hamiltonian
isapprox(H, permute(H, (TupleTools.circshift(codomainind(H), 1), TupleTools.circshift(domainind(H), 1)))
```

@Jutho if you could confirm that this is indeed expected, that would be greatly appreciated.

### DIFF
--- a/src/utility/utility.jl
+++ b/src/utility/utility.jl
@@ -89,7 +89,6 @@ function _embedders(spaces)
     return maps
 end
 
-
 #=
 map every element in the tensormap to dfun(E)
 allows us to create random tensormaps for any storagetype
@@ -151,5 +150,5 @@ function check_ambiguous_braiding(::Type{Bool}, V::VectorSpace)
 end
 function check_ambiguous_braiding(V::VectorSpace)
     return check_ambiguous_braiding(Bool, V) ||
-        throw(ArgumentError("cannot unambiguously braid $V"))
+           throw(ArgumentError("cannot unambiguously braid $V"))
 end
--- a/src/utility/utility.jl
+++ b/src/utility/utility.jl
@@ -133,6 +133,15 @@ function fuser(::Type{T}, V1::S, V2::S) where {T,S<:IndexSpace}
     return isomorphism(T, fuse(V1 ⊗ V2), V1 ⊗ V2)
 end
 
+# TODO: remove once TensorKit supports this
+_twist(t::AbstractTensorMap, i::Int) = _twist(BraidingStyle(sectortype(t)), t, i)
+_twist(::SymmetricBraiding, t::AbstractTensorMap, i::Int) = t
+function _twist(::NoBraiding, t::AbstractTensorMap, i::Int)
+    check_ambiguous_braiding(space(t, i))
+    return t
+end
+_twist(::BraidingStyle, t::AbstractTensorMap, i::Int) = twist(t, i)
+
 # Verify that the braiding is unambiguous: either through symmetric braiding or because the
 # sectors are all trivial
 function check_ambiguous_braiding(::Type{Bool}, V::VectorSpace)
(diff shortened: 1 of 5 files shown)


src/utility/utility.jl
```diff
@@ -143,3 +131,24 @@ end
 function fuser(::Type{T}, V1::S, V2::S) where {T,S<:IndexSpace}
     return isomorphism(T, fuse(V1 ⊗ V2), V1 ⊗ V2)
 end
+
+# TODO: remove once TensorKit supports this
+_twist(t::AbstractTensorMap, i::Int) = _twist(BraidingStyle(sectortype(t)), t, i)
+_twist(::Bosonic, t::AbstractTensorMap, i::Int) = t
+function _twist(::NoBraiding, t::AbstractTensorMap, i::Int)
```
leburgel, line 138: Is this superseded by Jutho/TensorKit.jl#240 now? If not, a more descriptive name might be good.

src/utility/utility.jl
```diff
@@ -143,3 +131,24 @@ end
...
+    check_unambiguous_braiding(space(t, i))
+    return t
+end
+_twist(::BraidingStyle, t::AbstractTensorMap, i::Int) = twist(t, i)
+
+# Verify that the braiding is unambiguous: either through symmetric braiding or because the
```
leburgel, line 144: Would it be better to make the comment into a docstring?

test/algorithms.jl
```diff
@@ -759,13 +759,24 @@ end
 
 @testset "periodic boundary conditions" begin
     Hs = [transverse_field_ising(), heisenberg_XXX(), classical_ising(), sixvertex()]
-    for N in 2:6
+    for N in 2:5
```
leburgel, line 762: Why the change here? Just for runtime?

leburgel (commented):
Changes themselves look good to me. I think I understand the fix, but I have to say I don't really understand what the problem actually was.

test/algorithms.jl
```diff
@@ -766,6 +766,17 @@ end
...
         end
     end
+
+    # fermionic tests
+    for N in 3:5
+        h = real(c_plusmin() + c_minplus())
+        H = InfiniteMPOHamiltonian([space(h, 1)], (1, 2) => h)
+        H_periodic = periodic_boundary_conditions(H, N)
+        terms = [(i, i + 1) => h for i in 1:(N - 1)]
+        push!(terms, (1, N) => h)
```
Jutho, line 776: I do think writing `push!(terms, (1, N) => permute(h, ((2,1), (4,3)))` would make it more clear why this is the correct term, and would allow to generalize this test to other Hamiltonians which would not have the symmetry that `permute(h, ((2,1), (4,3))) == h`

lkdvos: #277

PR #279, opened 2025-04-20
Fix time dependent integral of TDVP and TDVP2

GiggleLiu:
Thank you for developing such a great package!

When evolving the system with a time dependent Hamiltonian, I observed
1. the outcome is not accurate enough when comparing with the exact simulator.
2. hamiltonian at a nagative time is requested.

This PR will fix the above issues.


GiggleLiu:
The break of documentation seems to be irrelevant to this PR.

lkdvos (approved):
Thanks a lot for the interest and the fixes, this is really greatly appreciated.
The documentation failure is indeed unrelated, should be something that is fixed once we release the new version.
I'll merge this and try to get this released asap.

PR #283, opened 2025-04-29
Make `changebonds` truncation use relative truncation schemes.

lkdvos:
This PR changes the `changebonds` implementations to first normalize before computing the truncated SVD.
As we are not using the resulting `S`, this effectively has no influence on truncations like `truncdim`, however for `truncbelow` this will now truncate a normalized spectrum, therefore being a bit more "scale-invariant".
Effectively, this is manually converting the absolute truncation schemes defined in TensorKit to relative ones.

### DIFF
--- a/src/algorithms/changebonds/optimalexpand.jl
+++ b/src/algorithms/changebonds/optimalexpand.jl
@@ -30,7 +30,7 @@ function changebonds(ψ::InfiniteMPS, H::InfiniteMPOHamiltonian, alg::OptimalExp
         # Use the nullspaces and SVD decomposition to determine the optimal expansion space
         VL = leftnull(ψ.AL[i])
         VR = rightnull!(_transpose_tail(ψ.AR[i + 1]))
-        intermediate = adjoint(VL) * AC2 * adjoint(VR)
+        intermediate = normalize!(adjoint(VL) * AC2 * adjoint(VR))
         U, _, V, = tsvd!(intermediate; trunc=alg.trscheme, alg=alg.alg_svd)
 
         AL′[i] = VL * U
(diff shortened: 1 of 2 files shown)


VictorVanthilt (commented):
Makes sense, this should probably be reflected in the docstring. The original paper seems to only talk about `truncdim`-type behaviour so defining what we do with `truncbelow` seems okay.

lkdvos:
> if I understand correctly this now makes sure the squared singular values sum to 1

If you mean the singular values of the intermediate tensor: yes, this rescales the spectrum such that the singular values sum to 1.
If you mean the resulting MPS, not really - we're discarding the `S` to begin with, so this overall scale doesn't factor into the final result. Given that we are truncating the `U` and `V`, these are not isometries but rather projectors, so depending on the amount of truncation this process "loses norm". Interestingly though, that doesn't really matter all that much: since these are orthogonal to the original MPS I don't think this affects the final norm.

> it would be slightly more natural to normalize the AC2 resulting from the effective Hamiltonian right away, rather than normalizing after absorbing in the null spaces. But this would give just the same result, right?

It wouldn't, the `AC2` gets **projected** onto the nullspace, ie the norm is not preserved.

> where did you encounter a need to normalize before performing the SVD?

With the finite temperature methods (but this would also be relevant for boundary MPS type applications), the norm of the finite MPS or the MPO actually has meaning. In my case, it is precisely the value of the partition function, which is extensive, so it is actually possible to blow up certain entries if you naively ignore this and go to larger chain lengths.
In order to combat this, there are various ways to redistribute global scalars and factorize out normalizations. One option is to take the norm of the mpo and divide each tensor by the N-th square root, an alternative is to simply normalize the local tensors to have frobenius norm one.

While I was playing around with these different methods to see which one makes the most sense, the results changed quite drastically, because depending on the norm of the operator and MPS, `O * psi` can have a very large or very small norm, and using `truncbelow(1e-5)` is not a scale-invariant setting. Basically, even though the results are consistent, the truncation parameter discriminates between the two.

Long story short, we should really just have `truncbelow(; atol, rtol)`, but since that is not yet a thing (coming soon ™️ ), this fixes that in the meantime.

PR #286, opened 2025-05-12
Fix_notebook.jl

dartsushi:
The fixing of degeneracy was incorrect. I fixed the fix_degeneracies function and added some comments concerning the relation between the momentum and conformal spins.


examples/quantum1d/1.ising-cft/main.jl
```diff
@@ -51,33 +51,29 @@ either diagramatically as
...
-
+    
 function fix_degeneracies(basis)
-    N = zeros(ComplexF64, length(basis), length(basis))
+    L = length(basis[1])
     M = zeros(ComplexF64, length(basis), length(basis))
     for i in eachindex(basis), j in eachindex(basis)
-        N[i, j] = dot(basis[i], basis[j])
-        M[i, j] = momentum(basis[i], basis[j])
+        M[i, j] = dot(basis[i],O_shift(L)*basis[j])
```
lkdvos, line 74: probably want to take the `O_shift` constructor out of the loop here to avoid having to recompute it every time.

examples/quantum1d/1.ising-cft/main.jl
```diff
@@ -51,33 +51,29 @@ either diagramatically as
...
 """
-
-id = complex(isomorphism(ℂ^2, ℂ^2))
-@tensor O[-1 -2; -3 -4] := id[-1, -3] * id[-2, -4]
-T = periodic_boundary_conditions(InfiniteMPO([O]), L)
+function O_shift(L)
+    id = complex(isomorphism(ℂ^2, ℂ^2))
+    @tensor O[-1 -2; -3 -4] := id[-1, -3] * id[-2, -4]
+    T = periodic_boundary_conditions(InfiniteMPO([O]), L)
+    return T;
```
lkdvos, line 58:
```suggestion
    return T
```

lkdvos (commented):
Just making sure I get this right, the issue effectively was that the `momentum` function already took the angle, while it shouldn't have?
Also, the normalization was unnecessary because the vectors already come out as normalized?

Otherwise I'm definitely happy with this change, thanks for the fix!
If you don't mind I'll rerun the example and push the changes here so the docs get updated appropriately?

dartsushi:
>Just making sure I get this right, the issue effectively was that the momentum function already took the angle, while it shouldn't have?
>Also, the normalization was unnecessary because the vectors already come out as normalized?
Yes, basically!

>If you don't mind I'll rerun the example and push the changes here so the docs get updated appropriately?
Yes please :)

dartsushi:
I also realized that imposing Z2 helps to keep the degeneracies and thus the better momentum

docs/src/examples/quantum1d/1.ising-cft/main.ipynb
```diff
@@ -95,9 +95,11 @@
    "outputs": [],
    "cell_type": "code",
    "source": [
-    "id = complex(isomorphism(ℂ^2, ℂ^2))\n",
-    "@tensor O[-1 -2; -3 -4] := id[-1, -3] * id[-2, -4]\n",
-    "T = periodic_boundary_conditions(InfiniteMPO([O]), L)"
+    "function O_shift(L)\n",
+    "    id = complex(isomorphism(ℂ^2, ℂ^2))\n",
+    "    @tensor O[-1 -2; -3 -4] := id[-1, -3] * id[-2, -4]\n",
```
Jutho, line 100: How about? `id` is already a function in TensorKit.jl
```julia
    "    I = id(ComplexF64, ℂ^2)\n",
    "    @tensor O[-1 -2; -3 -4] := I[-1, -3] * I[-2, -4]\n",
```

examples/quantum1d/1.ising-cft/main.jl
```diff
@@ -52,33 +52,30 @@ either diagramatically as
 or in the code as:
 """
 
-id = complex(isomorphism(ℂ^2, ℂ^2))
-@tensor O[-1 -2; -3 -4] := id[-1, -3] * id[-2, -4]
-T = periodic_boundary_conditions(InfiniteMPO([O]), L)
+function O_shift(L)
+    id = complex(isomorphism(ℂ^2, ℂ^2))
+    @tensor O[-1 -2; -3 -4] := id[-1, -3] * id[-2, -4]
```
Jutho, line 57: Same here.

docs/src/examples/quantum1d/1.ising-cft/main.ipynb
Jutho: Oops, I didn't notice this PR is merged already; was reading it via email.

Issue #288, opened 2025-05-13
Copying `InfiniteMPO`s returns `FiniteMPO`s


VictorVanthilt:
When I copy `InfiniteMPO`s, I get back `FiniteMPO`s.
Small but annoying bug.

```julia
using MPSKit, MPSKitModels
H = transverse_field_ising(; J = J, g = h)
```

```julia-repl
julia> O = make_time_mpo(H, 0.1, TaylorCluster(2, true, true))
single site InfiniteMPO{BlockTensorKit.SparseBlockTensorMap{AbstractTensorMap{ComplexF64, Comp
lexSpace, 2, 2}, ComplexF64, ComplexSpace, 2, 2, 4}}:
╷  ⋮
┼ O[1]: 3×1×1×3 SparseBlockTensorMap(((ℂ^1 ⊕ ℂ^1 ⊕ ℂ^1) ⊗ ⊕(ℂ^2)) ← (⊕(ℂ^2) ⊗ (ℂ^1 ⊕ ℂ^1 ⊕ ℂ^1
)))
╵  ⋮

julia> copy(O)
single site FiniteMPO{BlockTensorKit.SparseBlockTensorMap{AbstractTensorMap{ComplexF64, Comple
xSpace, 2, 2}, ComplexF64, ComplexSpace, 2, 2, 4}}:
┬ O[1]: 3×1×1×3 SparseBlockTensorMap(((ℂ^1 ⊕ ℂ^1 ⊕ ℂ^1) ⊗ ⊕(ℂ^2)) ← (⊕(ℂ^2) ⊗ (ℂ^1 ⊕ ℂ^1 ⊕ ℂ^1
)))
```

Version information:

```julia-repl
julia> versioninfo()
Julia Version 1.11.5
```
```shell
Status `~/Documents/work/projects/temp/Project.toml`
  [bb1c41ca] MPSKit v0.13.1
  [ca635005] MPSKitModels v0.4.2
  [07d1fe3e] TensorKit v0.14.6
```
I will work on fixing it later today.

PR #289, opened 2025-05-13
Fix MPO copy issue of #288

VictorVanthilt:
This PR implements 2 explicit `copy` implementations for both `FiniteMPO` and `InfiniteMPO` to ensure the type of the output is the same as the original MPO.

I also implemented some small test to ensure this behaviour in the future making sure #288 does not return in the future.

### DIFF
--- a/src/operators/mpo.jl
+++ b/src/operators/mpo.jl
@@ -55,7 +55,8 @@ DenseMPO(mpo::MPO) = mpo isa DenseMPO ? copy(mpo) : MPO(map(TensorMap, parent(mp
 Base.parent(mpo::MPO) = mpo.O
 Base.copy(mpo::MPO) = MPO(map(copy, mpo))
 
-Base.copy(mpo::FiniteMPO) = FiniteMPO(map(copy, mpo)) # Explicitly split up Finite and Infinite to ensure proper type after copy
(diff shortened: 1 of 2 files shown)


src/operators/mpo.jl
```diff
@@ -55,6 +55,10 @@ DenseMPO(mpo::MPO) = mpo isa DenseMPO ? copy(mpo) : MPO(map(TensorMap, parent(mp
 Base.parent(mpo::MPO) = mpo.O
 Base.copy(mpo::MPO) = MPO(map(copy, mpo))
```
VictorVanthilt, line 56: Does it make sense to remove this line now?

lkdvos, line 56:
```suggestion
Base.copy(mpo::MPO) = MPO(copy.(parent(mpo)))
```

I think this would also fix the issue actually: the problem is that you want to preserve the container type, which actually does happen with broadcasting.

This entire thing is caused by the annoying fact that mpo's are abstractvector, and similar gives you a regular vector by default, thereby losing the `Infinite` information. In hindsight these really should not be `AbstractVector` to begin with, but oh well...

lkdvos (approved):

PR #293, opened 2025-05-20
fix `braille` and add small test

VictorVanthilt:
This PR makes the `braille` function usable again for `MPOHamiltonian`s, which are no longer considered `SparseMPO`s. Additionally it adds some hard-coded tests to ensure is keeps working in the future.

I had to add a new file `src/operators/utility.jl` which is `include`d after all the other operator code, to make sure the `SparseMPO` and `MPOHamiltonian` structs are defined before the `braille` function. If someone comes up with a better solution, please share.


src/operators/utility.jl
```diff
@@ -0,0 +1,33 @@
+function braille(io::IO, H::Union{SparseMPO,MPOHamiltonian})
+    isfinite = (H isa FiniteMPO) || (H isa FiniteMPOHamiltonian)
```
lkdvos, line 2: I think `isfinite(H) === Base.isfinite(H)` is actually defined directly (but check this)

lkdvos (commented):
Definitely makes sense to have a separate file, this is a good solution. Maybe it might be reasonable to simply define a `operators/show.jl` file and move all of this there?

I'm also okay with leaving it as it is now.

src/operators/utility.jl
VictorVanthilt: indeed this is defined, changed the code.

VictorVanthilt:
I moved the abstract mpo show methods, together with braille method to `operators/show.jl`. I also added a docstring to the braille function to make it less of a secret to the users.

lkdvos (approved):
Good for me if tests pass

Issue #292, opened 2025-05-20
Issues Encountered with IDMRG2 Algorithm


sunnowrun:
I’m currently using IDMRG2 in MPSKit, but have encountered some unexpected behavior.
Below is the relevant code snippet (I’m using the `MPS2MPO `function to handle Grassmann MPSs).
```
using TensorKit, MPSKit

function MPS2MPO(mps::MPSKit.InfiniteMPS)
	mpo = [zeros(scalartype(mps.AL[i]), codomain(mps.AL[i]) ← space(mps.AL[i],2) ⊗ space(mps.AL[i],3)') for i in eachindex(mps.AL)]
	for i in eachindex(mpo)
		for (f1,f2) in fusiontrees(mpo[i])
			f1′ = FusionTree((f1.uncoupled[1], Irrep[ℤ₂](f1.uncoupled[2].n + 1 % 2)), Irrep[ℤ₂](f1.coupled.n+1%2), f1.isdual)
			f2′ = FusionTree(f2.uncoupled[[2]], f2.uncoupled[2], f2.isdual[[2]])
			if f2.uncoupled[1].n == 0
				copyto!(mpo[i][f1,f2] , mps.AL[i][f1,f2′])
			elseif f1.uncoupled[2].n == 1
				copyto!(mpo[i][f1,f2] , mps.AL[i][f1′,f2′])
			end
		end
		for (f1,f2) in fusiontrees(mpo[i])
			if f1.uncoupled[1].n == 1 && f2.uncoupled[1].n == 1
				mpo[i][f1,f2] .*= -1
			end
		end

	end
	return MPSKit.InfiniteMPO(mpo)

end
function test(T = ComplexF64, L = 2)
	Dh = 3
	x = MPSKit.InfiniteMPS(randn, T, fill(Rep[ℤ₂](0=>1, 1=>1), L), fill(Rep[ℤ₂](0=>Dh, 1=>Dh), L))
	y = MPSKit.InfiniteMPS(randn, T, fill(Rep[ℤ₂](0=>1, 1=>1), L), fill(Rep[ℤ₂](0=>Dh, 1=>Dh), L))
	x′ = MPS2MPO(x)

	res0 = x′ * y

	Dh = 18
	z1 = MPSKit.InfiniteMPS(randn, T, fill(Rep[ℤ₂](0=>1, 1=>1), L), fill(Rep[ℤ₂](0=>Dh, 1=>Dh), L))
[...]

PR #294, opened 2025-05-29
IDMRG updates

lkdvos:
This PR refactors a bunch of the IDMRG implementations.

In particular, I've removed all hardcoded `@plansor` calls to make sure that the implementations are also working for MPS with multiple legs.
Additionally, somehow the edge handling for the `find_groundstate` and `approximate` implementations was very different.
This is now all the same, although I'm not entirely sure if this is correct, and in the future we should really look into de-duplicating this code, similar to what was done for VUMPS/VOMPS.

I've also added `leading_boundary` implementations to allow the PEPS algorithms to function.

Finally, I've fixed the following issue when working with IDMRG environments for statmech systems: because the expectation value scales multiplicative with system size, we have to renormalize the environments to ensure the values don't blow up. Previously, we did this in an asymmetric manner, absorbing the norm in the left environment. This was mostly to avoid taking square roots of negative real values, which would lead to complex environments whenever we split the norm equally over left and right.
However, this leads to a different problem: the norm of the right environment grows exponentially, while the left environment shrinks exponentially to ensure their product remains one. At some point, LAPACK complains and starts failing.
I've changed this to now always normalize the right environment to 1, and the left environment as before to ensure the inner product is 1.

### DIFF
--- a/src/algorithms/approximate/idmrg.jl
+++ b/src/algorithms/approximate/idmrg.jl
@@ -21,7 +21,7 @@ function approximate!(ψ::MultilineMPS, toapprox::Tuple{<:MultilineMPO,<:Multili
             end
 
             # right to left sweep
-            for col in size(ψ, 2):-1:1
+            for col in reverse(1:size(ψ, 2))
                 for row in 1:size(ψ, 1)
                     ψ.AC[row + 1, col] = AC_projection(CartesianIndex(row, col),
                                                        ψ, toapprox, envs)
@@ -71,7 +71,7 @@ function approximate!(ψ::MultilineMPS, toapprox::Tuple{<:MultilineMPO,<:Multili
             C_current = ψ.C[:, 0]
 
             # sweep from left to right
-            for site in 1:(size(ψ, 2) - 1)
+            for site in 1:size(ψ, 2)
                 for row in 1:size(ψ, 1)
                     AC2′ = AC2_projection(CartesianIndex(row, site), ψ, toapprox, envs;
                                           kind=:ACAR)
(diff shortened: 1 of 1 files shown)


src/algorithms/approximate/idmrg.jl
```diff
@@ -127,7 +177,7 @@ function approximate!(ψ::MultilineMPS, toapprox::Tuple{<:MultilineMPO,<:Multili
 
     # TODO: immediately compute in-place
     alg_gauge = updatetol(alg.alg_gauge, iter, ϵ)
-    ψ′ = MultilineMPS(map(x -> x, ψ.AR); alg_gauge.tol, alg_gauge.maxiter)
+    ψ′ = MultilineMPS(map(identity, ψ.AR); alg_gauge.tol, alg_gauge.maxiter)
```
VictorVanthilt, line 180: Learned something new today :)

src/states/finitemps.jl
```diff
@@ -318,6 +318,9 @@ Base.@propagate_inbounds function Base.getindex(ψ::FiniteMPS, i::Int)
     end
 end
 
+# TODO: check where gauge center is to determine efficient kind
```
VictorVanthilt, line 321: This would be a nice addition indeed, and to my knowledge the first ever use of the `center` functionality I implemented a while ago.

VictorVanthilt (commented):
Should we write some extra tests to test the handling of the edges of `find_groundstate` and `approximate`?

src/algorithms/approximate/idmrg.jl
lkdvos: Hehe it is a fun trick to try and get a matrix of all the tensors

lkdvos:
Yeah probably some checks are a good idea, especially given the reported issues in #292, which I still need to further delve into.

Issue #295, opened 2025-05-31
In MPSKit v0.13.2, ``timestep!`` yields unphysical results, whereas in v0.12.6, it does not


ZongYongyue:
For a L=8 1d Hubbard model with 1/4 hole dopping, the correct result is Fig.1 but now with MPSKit v0.13.2 it's like Fig.2. The only variable between these two different results is the MPSKit version.

<img width="680" alt="Image" src="https://github.com/user-attachments/assets/bd0e3136-10cc-4d2e-a149-fda19d33abb6" />
<img width="681" alt="Image" src="https://github.com/user-attachments/assets/143a7632-76c7-4e5b-9e19-eacbb6059e78" />

lkdvos:
Is it possible that this is the result of the change in convention for `sector` keywords etc?

ZongYongyue:
I don't know if this is related to what you mentioned, but I used both the method of adjusting filling with shift MPS and MPO quantum numbers (as mentioned in Mortier, Q. et al. Fermionic tensor network methods. arXiv (2024)), and the method of injecting quantum numbers into the MPS from one end, and compared DMRG with ED, obtaining consistent results. The ground state, ground state energy, and static correlations are all correct under both versions of MPSKit. Only during time evolution does v0.12 remain normal while v0.13 contains errors. Therefore I believe the issue must lie solely in the error within timestep!.

ZongYongyue:
If the error is in the ``sector`` level, the DMRG results might also be wrong I thought?

ZongYongyue:
MPSKit@v0.12.6:

```julia
using TensorKit
using MPSKit
using MPSKitModels: FiniteChain
using DynamicalCorrelators: hubbard, randFiniteMPS, e_min, chargedMPS

elt = Float64
I = FermionParity ⊠ SU2Irrep ⊠ U1Irrep 
L = 4
H = hubbard(elt, SU2Irrep, U1Irrep, FiniteChain(L); t=0.5, U=3, μ=0)
dop = -2
spin = 0
Vs = repeat([Vect[I]((Int(!isinteger(i)), i, j) => 1  for i in 0:1//2:1 for j in -(abs(dop) + 1):(abs(dop)+1)),], L - 1)
Ps = Vect[I]((0,0,-1)=>1, (1,1//2,0)=>1, (0,0,1)=>1)
[...]

PR #296, opened 2025-06-08
Fix derivatives at the edges of finite operators

lkdvos:
This is a bugfix for behavior described in #295, which still needs some extra tests.
In particular, we should probably test somewhere that time-evolving a groundstate yields the desired results, which is how I debugged this (as originally shown by #295 as well).

## To do
- [x] Add test for phase of evolution of eigenstate
- [x] Add test for handling real inputs

### DIFF
--- a/src/algorithms/derivatives/hamiltonian_derivatives.jl
+++ b/src/algorithms/derivatives/hamiltonian_derivatives.jl
@@ -107,7 +107,9 @@ function AC_hamiltonian(site::Int, below::_HAM_MPS_TYPES,
     end
 
     # not_started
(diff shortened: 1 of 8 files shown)


ZongYongyue:
> Is this with the latest version of this branch already? Could you tell me what you are running and what the error message is? I can't seem to reproduce it for now
> 
> [edit] found it, should be fixed soon

It is fine now😄

VictorVanthilt:
Wow, didn't know the tests run on different versions than the `[compat]` in the Project.toml

lkdvos:
I agree that this is not necessarily a great design choice but oh well 🤷

PR #297, opened 2025-06-13
Compatibility with multifusion categories

borisdevos:
This PR provides the changes required to be able to consider the multifusion categories of (Multi)TensorKit. 

All changes boil down to either
1) not being able to evaluate the `unitspace` of a `BimoduleSector`, or
2) not being able to use braiding tensors for `BimoduleSector` `GradedSpace`s,
and finding ways around to this.

~A potential thing to add is tests with `BimoduleSector` MPS/MPOs or so? For this, I would need to first complete benchmarking MultiTensorKit, hence the draft PR. I might also need to format.~

### DIFF
--- a/src/algorithms/excitation/quasiparticleexcitation.jl
+++ b/src/algorithms/excitation/quasiparticleexcitation.jl
@@ -104,13 +104,20 @@ function excitations(
         verbosity = Defaults.verbosity, num = 1,
         sector = leftunit(lmps), parallel = true, kwargs...
     )
-    Toutput = Core.Compiler.return_type(
-        excitations,
-        Tuple{
-            typeof(H), typeof(alg), eltype(momenta), typeof(lmps),
-            typeof(lenvs), typeof(rmps), typeof(renvs),
-        }
-    )
+    # wrapper to evaluate sector as positional argument
+    Toutput = let
+        function wrapper(H, alg, p, lmps, lenvs, rmps, renvs, sector; kwargs...)
+            return excitations(
+                H, alg, p, lmps, lenvs, rmps, renvs; sector, kwargs...
+            )
+        end
+        Core.Compiler.return_type(
+            wrapper, Tuple{
+                typeof(H), typeof(alg), eltype(momenta), typeof(lmps),
+                typeof(lenvs), typeof(rmps), typeof(renvs), typeof(sector),
+            }
+        )
+    end
 
     results = similar(momenta, Toutput)
     scheduler = parallel ? :greedy : :serial


src/algorithms/excitation/exci_transfer_system.jl
```diff
@@ -1,3 +1,4 @@
+# is this function ever called? b/c isid no longer exists
```
lkdvos, line 1: it isnt 🙃

borisdevos:
I may have added some confusing terminology with the recent commits, so let me clarify as much as possible.

First of all, I defined a `left/rightunit` for MPS and MPO. For the MPS this makes most sense, as my virtual space will in general be graded by module objects, so the "coloring" to the left and right of this is different. So in our convention of physical legs down, `rightunit` captures the coloring of the physical space, while `leftunit` captures the coloring of auxiliary legs such as in `excitations`, `transfer_spectrum`, etc.
For MPOs I'm still on the fence, because while in general my MPO's virtual space can also be graded by module objects, the only thing I've tested up till now is Hamiltonians where it's fully colored by the `rightunit` of the corresponding MPS. So this notation is introduced to ease notation mostly. I can deal with not having this; this appears only in the keyword argument of `exact_diagonalization`.

Connected to the last remark, many functions construct an MPS given an MPO based on the latter's virtual spaces. Since the MPOs I've considered up till now don't have different colorings on its virtual spaces, there is currently no difference between the left unit and right unit of an MPO. Thus, there's no difference between the `leftunitspace` and `rightunitspace` of the virtual space of an MPO, and I simply systematically chose to put `rightunitspace` to accentuate the "physicalness" of the coloring.

I've also replaced the remaining `oneunit`s with `unitspace`s, where I didn't think too hard whether it was a `leftunitspace` or `rightunitspace` because I'd have to fiddle a bit everywhere to actually get the information of the coloring (reminder that `left/rightunitspace` can't be called on a space type), and in my usage I didn't encounter these constructors. If this is wished, I can more extensively test how these behave with a multifusion sector. Most are easy to tell beforehand that they can't be used. For example, anything requiring non-trivial braiding cannot be done.

src/operators/mpo.jl
```diff
@@ -223,8 +223,9 @@ function Base.:*(mpo1::FiniteMPO{<:MPOTensor}, mpo2::FiniteMPO{<:MPOTensor})
     N = check_length(mpo1, mpo2)
     (S = spacetype(mpo1)) == spacetype(mpo2) || throw(SectorMismatch())
 
-    if (left_virtualspace(mpo1, 1) != oneunit(S) || left_virtualspace(mpo2, 1) != oneunit(S)) ||
-            (right_virtualspace(mpo1, N) != oneunit(S) || right_virtualspace(mpo2, N) != oneunit(S))
+
+    if (isunitspace(left_virtualspace(mpo1, 1)) || isunitspace(left_virtualspace(mpo2, 1))) ||
```
lkdvos: Might be missing a negation here?

src/utility/utility.jl
```diff
@@ -56,7 +56,20 @@ Add trivial one-dimensional utility spaces with trivial sector to the left and r
 given tensor map, i.e. as the first space of the codomain and the last space of the domain.
 """
 function add_util_leg(tensor::AbstractTensorMap{T, S, N1, N2}) where {T, S, N1, N2}
-    ou = oneunit(_firstspace(tensor))
+    ou = unitspace(_firstspace(tensor))
```
lkdvos: Can we replace this function with `insertleftunit` etc?

test/multifusion.jl
lkdvos: Is there any chance these things can easily be interspersed with the other tests?
In any case, we should really clean up the tests of MPSKit at some point, so if this isn't too easy I'm fine keeping it like this and tackling it in the future

src/algorithms/timestep/taylorcluster.jl
```diff
@@ -186,14 +186,14 @@ function make_time_mpo(
     H′ = copy(parent(H))
 
     V_left = left_virtualspace(H[1])
-    V_left′ = ⊞(V_left, oneunit(V_left), oneunit(V_left))
+    V_left′ = ⊞(V_left, rightunitspace(V_left), rightunitspace(V_left))
```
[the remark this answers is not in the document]
borisdevos: The Hamiltonian itself is diagonal, meaning its virtual spaces are graded the same way as the physical space. In that sense, taking the left or right unit space is the same. I haven't thought too much about direct summing between different gradings and whether that makes sense.
Either way, `unitspace` is not the way to go, because we've designed that now to return all the units of a given sector.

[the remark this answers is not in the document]
lkdvos: Ah right, I forgot about that! Would this mean that we could take the left unit space of the physical space instead? That at least sounds spatially correct in my head 😂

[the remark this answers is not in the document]
borisdevos: I can do that, and will do so in other places where `MPOHamiltonian` is involved. I'd like to add an assertion to make sure this is fine. However, for MPOs I'd like to keep the freedom of potential off-diagonal virtual spaces, even though I haven't tested those yet.

PR #303, opened 2025-06-30
Runic formatter

lkdvos:


VictorVanthilt (approved):
No objections. I like the [QuantumKitHubActions](https://github.com/QuantumKitHub/QuantumKitHubActions/) repo although changes to it will not be explicitly reflected in the git history of the repos dependent on it. I don't see any issues arising from this but this is something we need to keep in the back of our minds.

lkdvos:
yeah, technically I should version control that as well (and in principle that is definitely possible), I just haven't gotten to that

PR #304, opened 2025-07-03
Fix translation operator in Ising CFT example

leburgel:
Fixes the translation operator in the Ising CFT example. This was (I assume) accidentally changed to the identity in #290, making all the momenta identically zero. Originally noticed by noticed by Meng Cheng, and reported by @tangwei94.

While I was at it, I changed the example checksum generation to use the relative path similar to what was done in QuantumKitHub/PEPSKit.jl#227. When attempting to rebuild the Ising example I noticed that it was actually rebuilding all of them, so seemed appropriate to squeeze this in here to prevent this in the future.


leburgel:
@dartsushi, can you check that the plots [here](https://quantumkithub.github.io/MPSKit.jl/previews/PR304/examples/quantum1d/1.ising-cft/#The-Ising-CFT-spectrum) now give what you would expect?

examples/quantum1d/1.ising-cft/main.jl
```diff
@@ -53,8 +53,8 @@ or in the code as:
 """
 
 function O_shift(L)
-    τ = BraidingTensor{ComplexF64}(ℂ^2, ℂ^2)
-    O = TensorMap(τ)
+    I = id(ComplexF64, ℂ^2)
+    @tensor O[-1 -2; -3 -4] := I[-1, -3] * I[-2, -4]
```
lkdvos, line 57: Ah I see how I messed this up... Also, then this is probably:
```suggestion
    O = id(ComplexF64, (ℂ^2)^2)
```

leburgel: Yes, should be the same. However, think the single `id` on the tensor product space is a bit more opaque than tensoring the two identities together. Just for pedagogical purposes, took a second to convince myself the shorter version was the same. What do you think?

dartsushi:
> @dartsushi, can you check that the plots [here](https://quantumkithub.github.io/MPSKit.jl/previews/PR304/examples/quantum1d/1.ising-cft/#The-Ising-CFT-spectrum) now give what you would expect?

Perfect. Thanks for fixing those!

examples/quantum1d/1.ising-cft/main.jl
leburgel: Agreed on the labels. Do you think plane directions are clear enough
```julia
@tensor O[W S; N E] := I[W; N] * I[S; E]
```
or would something more explicit like
```julia
@tensor O[left bottom; top right] := I[left; top] * I[bottom; right]
```
be even better?

Jutho: This shifts everything one site to the left if I read it correctly. Wouldn't you say that the elementary translation operator shifts everything to the right? Or am I mixing the transformation on the basis vs on the coefficients / active vs passive / ... ?

leburgel: > Wouldn't you say that the elementary translation operator shifts everything to the right?

I don't know what the standard is. But I would abuse the choice between active or passive viewpoint to pick whichever corresponds to the current implementation, since this is the most natural choice with respect to the domain/codomain partition convention of MPO tensors.

lkdvos (approved):
Thanks for fixing this!

PR #305, opened 2025-07-09
Set up optional color inversion for docs images

leburgel:
Sets up optional color inversion for images in the docs. Seems to work fine for both .png and .svg images, by including images through raw HTML and adding a `color-invertible` class:
````
```@raw html
<img src="path/to/image.svg" class="color-invertible"/>
```
````
That way images are not inverted by default, which seems appropriate for e.g. plots. This also allows for specifying the size by adding something like `width="50%"`, so I think this is sufficient for our use cases.

Fixing the logos to only have one image file was a bit more tricky. In particular, the only way I managed to make the README logo invertible was to embed some custom css into an svg and then including the normal logo through base6464 encoding. A little shady, and if the logo changes this README logo would have to be manually updated. But I can't really see a way around this.

I was also planning on updating and adding some actual images, but I'm not sure when I'll get to that so it might be good to merge just the setup first.

Preview is [here](https://quantumkithub.github.io/MPSKit.jl/previews/PR305/).

### DIFF
--- a/examples/quantum1d/3.ising-dqpt/main.jl
+++ b/examples/quantum1d/3.ising-dqpt/main.jl
@@ -1,5 +1,5 @@
 md"""
-# DQPT in the Ising model(@id demo_dqpt)
+# [DQPT in the Ising model](@id demo_dqpt)
 
 In this tutorial we will try to reproduce the results from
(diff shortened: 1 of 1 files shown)


lkdvos (approved):
Thanks, looks good to me. I like that it is optional, and that the required setup is pretty minimal.

PR #310, opened 2025-08-02
[Utility] Be more consistent about what space functions work for what arguments

lkdvos:
Here I'm just being slightly more careful about defining `physicalspace`, `left_virtualspace` and `right_virtualspace` both with a coordinate and without, which was already partially implemented for some things and not for others and kept confusing me.


lkdvos:
I really love how you phrased this ❤️

VictorVanthilt:
Thanks for caring!

Issue #311, opened 2025-08-03
U(2) DMRG performance unexpectedly slower than U(1) ITensor version


drichg:
Hi, thank you for the great package!
I'm benchmarking DMRG calculations using MPSKit.jl with SU(2) symmetry, and I encountered a significant performance discrepancy compared to a U(1)-symmetric calculation done with ITensor.
Here are my benchmarks:
**ITensors (Julia):**
Symmetry: U(1)
Bond dimension: D = 4800
Time per sweep: 20.92 hours.
output:
`After sweep 1 energy=-337.23217296396973  maxlinkdim=4800 maxerr=3.46E-05 time=70248.585
After sweep 2 energy=-337.2389290839439  maxlinkdim=4800 maxerr=3.78E-05 time=75324.747
`
**MPSKit (0.12.6)**
Symmetry: SU(2)
Max bond dimension: D = 4000
Time per sweep: 31.66 hours.
output:
`[ Info: DMRG2   1:	obj = -3.164484013797e+02	err = 9.1771680577e-01	time = 4.42 min
[ Info: DMRG2   2:	obj = -3.292824077866e+02	err = 1.7325899602e-02	time = 17.07 min
[ Info: DMRG2   3:	obj = -3.348343234046e+02	err = 4.5124942262e-03	time = 5.39 hr
[ Info: DMRG2   4:	obj = -3.360135495584e+02	err = 8.0470138898e-04	time = 31.66 hr
`
Both calculations were performed on the same hardware, using a 64-core, 128-thread CPU machine. This seems counterintuitive, as I expected that the SU(2) symmetry would help reduce the computational cost due to block structure. Is this behavior expected, or might I have misconfigured something in my setup?
Thanks again for your hard work on this project!

lkdvos:
Thanks for sharing these results, this is definitely very valuable! Would you mind sharing a bit more details about what model you are trying out? I agree that we are expecting that the symmetry should lead to an increase in performance, so I'm happy to check in more detail what's going on.

If you prefer to share in a more confident setting, also do feel free to email me about the details, I'm definitely also happy to try and debug what's going on that way.

drichg:
> Thanks for sharing these results, this is definitely very valuable! Would you mind sharing a bit more details about what model you are trying out? I agree that we are expecting that the symmetry should lead to an increase in performance, so I'm happy to check in more detail what's going on.
> 
[...]

PR #312, opened 2025-08-04
[Feature] FiniteMPOHamiltonian compression

lkdvos:
This is an implementation of the algorithm described in [Parker et al](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.102.035147), which allows compression of MPOHamiltonian objects without ruining the `JordanMPOTensor` structure.
There are still some questions about implementation details, tests and docs, and I would like to at least at the reference somewhere before merging, but I already wanted to add this here so people might start playing around with it.
Also, I would like to add the infinite version as well.

Importantly, this keeps the `JordanMPOTensor` structure but ruins the sparsity, so this is not in general something we always want to do. However, especially when working with quasi-2D systems, often terms of the form `XiXi+1` and `X_iX_i+L` appear and can be efficiently compressed, which is not something that our MPO construction scheme would naturally do.

See also #311

### DIFF
--- a/src/operators/ortho.jl
+++ b/src/operators/ortho.jl
@@ -1,16 +1,18 @@
-function TensorKit.leftorth!(W::JordanMPOTensor; alg = QRpos())
+function TensorKit.leftorth(W::JordanMPOTensor; alg = QRpos())
     # orthogonalize second column against first
     WI = removeunit(W[1, 1, 1, 1], 1)
     @tensor t[l; r] := conj(WI[p; p' l]) * W.C[p; p' r]
     I = sectortype(W)
     S = spacetype(W)
 
-    @tensor t[l; r] := conj(removeunit(W[1, 1, 1, 1], 1)[p; p' l]) * W.C[p; p' r]
-    @tensor C′[p; p' r] := W.C[p; p' r] - WI[p; p' l] * t[l; r]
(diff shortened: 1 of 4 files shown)


src/operators/ortho.jl
```diff
@@ -0,0 +1,187 @@
...
+    1 ≤ i < length(H) || throw(ArgumentError("Bounds error in canonicalize"))
+
+    W = H[i]
+    S = spacetype(W)
+    d = sqrt(dim(physicalspace(W)))
+
+    # orthogonalize second column against first
+    WI = removeunit(W[1, 1, 1, 1], 1)
+    @tensor t[l; r] := conj(WI[p; p' l]) * W.C[p; p' r]
+    # @plansor C′[p; p' r] := W.C[p; p' r] - WI[p; p' l] * t[l; r]
```
VictorVanthilt, line 14: small cleanup

VictorVanthilt (approved):
Other than the small comment this is good to go!

src/operators/ortho.jl
VictorVanthilt, line 14:
```suggestion

```

lkdvos: Actually that is there on purpose, it's because of a bug in TensorKit that has been resolved but not fixed. I'll add a better comment though, it seems reasonable that this isn't clear without context ;)

PR #313, opened 2025-08-07
Try to fix README badges

kshyatt:
The `PkgEval` badge had the wrong initial, and hopefully pointing coverage at a specific branch will cause it to actually display.


lkdvos:
Cool, thanks!

kshyatt:
Now the badges are rendering, hooray! (Though the results they are rendering are rather sad...)

PR #315, opened 2025-08-20
Remove IDMRG logging incorrect energy

borisdevos:
During the IDMRG iterations, the objective which is supposed to represent the expectation value of the Hamiltonian is completely incorrect. This is because the environment only gets recalculated with `recalculate!` at the end of the sweeps, so these intermediate values are irrelevant. 

I chose to remove showing these values, but another option is to calculate the expectation value without passing the environments, thus recalculating at every step. I don't know how costly this step is, so I chose to do it like in `IDMRG2`.


borisdevos:
The docs failing is probably because TensorOperations moved to the QuantumKitHub, right?

PR #318, opened 2025-08-28
Refactor quasiparticle regularization

lkdvos:
This PR simplifies and reuses some code by using functions that already existed, and additionally refactors that to avoid using braiding tensors.
Overall this should be more friendly to multifusion things, and possibly slightly more efficient.

### DIFF
--- a/test/algorithms.jl
+++ b/test/algorithms.jl
@@ -479,8 +479,7 @@ module TestAlgorithms
             g = 4
             H = repeat(transverse_field_ising(Z2Irrep; g = g, L = Inf), 2)
             V = Z2Space(0 => 24, 1 => 24)
-            P = Z2Space(0 => 1, 1 => 1)
-            ψ = InfiniteMPS([P, P], [V, V])
+            ψ = InfiniteMPS(physicalspace(H), [V, V])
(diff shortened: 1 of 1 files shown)


src/algorithms/excitation/exci_transfer_system.jl
```diff
@@ -59,8 +15,9 @@ function left_excitation_transfer_system(
...
         start = scale!(last(found[1:i] * T), cis(-mom * len))
         if exci.trivial && isidentitylevel(H, i)
-            @plansor start[-1 -2; -3 -4] -= start[1 4; -3 2] * r_RL(exci.right_gs)[2; 3] *
-                τ[3 4; 5 1] * l_RL(exci.right_gs)[-1; 6] * τ[5 6; -4 -2]
+            ρ_left = l_RL(exci.right_gs)
+            ρ_right = r_RL(exci.right_gs)
+            regularize!(start, ρ_right, ρ_left)
         end
 
         found[i] = add!(start, GBL[i])
```
borisdevos, line 23: Same comment here as below (oops)

borisdevos:
This seems to fix the issue locally. I'm also a fan of the name `istopological`.

lkdvos:
Any chance you can migrate the charged quasiparticle tests here? It would be nice to make sure this is actually passing

borisdevos (approved):

PR #325, opened 2025-09-24
Small fixes

VictorVanthilt:
This PR implements 2 small fixes:
1. Removes the 2 files that were related to the YAS style formatting that has been abandoned in favor of runic in #303 .
2. Update the Citation file, along with the DOI in this file and the DOI in the readme.

A different DOI is associated with every release, which is rather cumbersome but since the one in the citation file and the readme was from MPSKit 0.10 I figured an update was needed.

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,1 +1,0 @@
-style = "yas"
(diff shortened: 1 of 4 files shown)


VictorVanthilt:
There is! Learned something new.

Issue #320, opened 2025-09-03
Expectation value of MPO tensor


borisdevos:
Some context: I have a partition function tensor `O`. I find the boundary MPS fixed point `psi` (and environments `envs`) with VOMPS. I can get to the free energy through `expectation_value(psi, InfiniteMPO([O]), envs)`.

I now have an MPO tensor `E` representing a different quantity (here some energy, but not a Hamiltonian term!) . I did the following:
```julia
mpo = InfiniteMPO([E]) # E is a (2,2) tensor map
expectation_value(psi, mpo, envs)
```
However, this gives an incorrect result. I think this is because the environments get lost in the conversion to multiline things: https://github.com/QuantumKitHub/MPSKit.jl/blob/8cf041ff823492e41fd5d812fe5b03cf45ac54aa/src/algorithms/expval.jl#L138
where there is a TODO noting this.
Edit: this is also flawed because I want the expectation value of replacing 1 `O` tensor with `E`, not a chain of `E`s.

The best I can do right now is 
```julia
MPSKit.contract_mpo_expval(psi.AC[1], envs.GLs[1], E, envs.GRs[1], psi.AC[1])
```
to get the right result.

Something similar to [PEPSKit's `contract_local_operator`](https://quantumkithub.github.io/PEPSKit.jl/dev/lib/lib/#PEPSKit.contract_local_operator-Union{Tuple{N},%20Tuple{S},%20Tuple{T},%20Tuple{NTuple{N,%20CartesianIndex{2}},%20TensorKit.AbstractTensorMap{T,%20S,%20N,%20N},%20InfinitePEPS,%20InfinitePEPS,%20CTMRGEnv}}%20where%20{T,%20S,%20N}) is probably what I want.

lkdvos:
Yeah, using an InfiniteMPO implicitly places that everywhere so that’s definitely no good. Unfortunately this is one of those not so hard to implement but hard to come up with a nice interface problems, if you have some inspiration I’m all ears!

See also https://github.com/QuantumKitHub/PEPSKit.jl/pull/117 and the discussions there

PR #327, opened 2025-09-29
`local_expectation_value` function

borisdevos:
This PR introduces the function `local_expectation_value`. This was originally meant for in the context of (infinite) partition functions and thus MPO tensors (see #320), but I extended the definition to MPO hamiltonians. 

The local expectation values are currently not normalised in the MPO case (as can be seen in the added tests). I don't know what's the cleanest way to do that, since the environments are not from the MPO tensor itself. I'm open to suggestions. Maybe @VictorVanthilt can chime in, since I'm doing the complementary thing to TNR.

I didn't add anything for the finite case, partly because I don't need it, and mostly because it's annoying to make a test. If it's really wanted, I can look into it, but this won't happen any time soon 🫠.

### DIFF
--- a/src/MPSKit.jl
+++ b/src/MPSKit.jl
@@ -42,7 +42,7 @@ export DynamicalDMRG, NaiveInvert, Jeckelmann
 export exact_diagonalization, fidelity_susceptibility
 
 # toolbox:
-export expectation_value, correlator, variance
+export expectation_value, local_expectation_value, correlator, variance
 export correlation_length, marek_gap, transfer_spectrum
 export entropy, entanglement_spectrum
 export open_boundary_conditions, periodic_boundary_conditions
--- a/src/algorithms/expval.jl
+++ b/src/algorithms/expval.jl
@@ -119,6 +119,26 @@ function local_expectation_value(
     )
 end
 
+# MPO tensor
+#-----------
+function expectation_value(
+        ψ::InfiniteMPS, O::MPOTensor,
+        envs::AbstractMPSEnvironments
+    )
+    return sum(1:length(ψ)) do site
+        return local_expectation_value(ψ, O, envs, site)
+    end
+end
+
+function local_expectation_value(
+        ψ::InfiniteMPS, O::MPOTensor,
+        envs::AbstractMPSEnvironments, site::Int=1
+    )
+    return contract_mpo_expval(
+        ψ.AC[site], envs.GLs[site], O, envs.GRs[site]
+    )
+end
+
 # DenseMPO
 # --------
 function expectation_value(ψ::FiniteMPS, mpo::FiniteMPO)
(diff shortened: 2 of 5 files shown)


lkdvos (commented):
Thank you for the PR!

Some comments about the `MPOHamiltonian` case: we previously had functionality similar to this, but we actually decided to remove this. The main issue is that a "local" expectation value is actually very much not well-defined in this context. The idea is that if you encode your Hamiltonian in different but equivalent ways into an MPO, you will obtain different values.
For example, in the implementation you have here you make the choice to count for the local expectation value all of the interactions that end on a given site. You could of course just as well take the interactions that begin at a site, or the average, or ... Really the only thing that is well defined is the total sum, which is why we only expose that. If you want a "local" expectation value, you have to say which operator you want and on which sites it acts, through `expectation_value(state, inds => local_operator)`

For the case you have this doesn't apply, since the user is providing the tensor you are inserting, so I definitely agree that this is good functionality to have.
Considering the interface, we already have the following syntax for local (hamiltonian) operators:
```julia
expectation_value(state, i => O_1, [envs])
expectation_value(state, (i, j, ...) => O_N, [envs])
```
I think I would prefer to adopt a similar approach here, to avoid having to introduce another function name. This is mostly because I think `expectation_value` already works for local things, so then it might be strange to only have to add `local_` for the MPO case.

Nevertheless, simply copying that has two issues. One the one hand, there is the normalization issue that you already mentioned, since you don't have access to the original MPO anymore and the environments are normalized to have a zero-site expectation value of 1 (instead of the single site expectation value), and on the other hand I am not so fond of making the `envs` a required argument, since this is mostly an implementation detail in most of our cases.
Therefore, thinking about this for a bit, might I suggest either of the following signatures:

```julia
expectation_value(state, (mpo, site => mpotensor), envs = environments(state, mpo))
expectation_value(state, mpo, site => mpotensor, envs = environments(state, mpo))
```

The idea is to adhere to the same style as we had before, and not have to introduce another function, to have the environments as optional arguments and to have enough information to properly normalize.
I think I slightly like the first form better since then the first argument is always the state and the second argument always the "operator", but I'm open to suggestions.

Obviously, open to discuss this further, and possibly involving more people that would want to use this.

---

As a final wild idea that I want to share, in principle what you are asking for is the expectation value of a window-mpo of length 1. We could consider making that the actual implementation, and working with datastructures like that. This has the advantage that it would also generalize to longer ranges, and could be implemented in a completely generic way.
I'm not actually suggesting to do that here, since I think this is a little bit of work that might not (yet) be worth it (for example our current window structures always start at site 1), but I wanted to share that idea as a possible thing to keep in mind in case someone at some point wants to do more advanced things with this.

kshyatt:
@borisdevos the algorithm tests are also timing out on `main` so I wouldn't say those are blocking. Are you ready for this to go in?

lkdvos:
I think this was already the case before so should be fine

PR #328, opened 2025-10-02
Modify `_totuple`

Yue-Zhengyuan:
This PR catches the edge case `tensorexpr(:name, :i, :j)` so it can be converted to `tensorexpr(:name, (:i,), (:j,))` correctly.
```julia
julia> tuple(:a)
(:a,)

julia> Tuple(:a)
ERROR: MethodError: no method matching length(::Symbol)
The function `length` exists, but no method is defined for this combination of argument types.
...
```
See also https://github.com/QuantumKitHub/PEPSKit.jl/pull/268.

### DIFF
--- a/src/utility/utility.jl
+++ b/src/utility/utility.jl
@@ -112,7 +112,7 @@ function randomize!(a::AbstractBlockTensorMap)
     return a
 end
 
-_totuple(t) = t isa Tuple ? t : tuple(t)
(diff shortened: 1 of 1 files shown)


lkdvos (approved):

Yue-Zhengyuan:
@lkdvos Can you help with bumping the version and releasing it? I don't quite know how it's done, and might mess up the release notes or something else...

PR #330, opened 2025-10-06
precompute contraction GL*A and A*GR in JordanMPO_AC_Hamiltonian and JordanMPO_AC2_Hamiltonian

AFeuerpfeil:
This PR changes the way `JordanMPO_AC2_Hamiltonian` and `JordanMPO_AC_Hamiltonian` deal with the A-block of the MPO: 
To reduce contraction const in the application to an MPS `MPSTensor`, we contract `GL*A` and `A*GR` in the constructor for the 2-site derivative and `GL*A` for the 1-site derivative (absorbing A into the left environment is purely conventional. If the mpo leg of the left and right environments have different sizes, one of the 2 conventions could be superior to the other.)


src/algorithms/derivatives/hamiltonian_derivatives.jl
```diff
@@ -303,9 +314,9 @@ function (H::JordanMPO_AC_Hamiltonian)(x::MPSTensor)
     ismissing(H.C) || @plansor y[-1 -2; -3] += x[-1 2; 1] * H.C[-2 -3; 2 1]
     ismissing(H.B) || @plansor y[-1 -2; -3] += H.B[-1 -2; 1 2] * x[1 2; -3]
 
-    GL, A, GR = H.A
-    if nonzero_length(A) > 0
-        @plansor y[-1 -2; -3] += GL[-1 5; 4] * x[4 2; 1] * A[5 -2; 2 3] * GR[1 3; -3]
+    if !ismissing(H.A)
+        GLW, GR = H.A
+        @tensor y[-1 -2; -3] += GLW[-1 -2 3; 1 2] * x[1 2; 4] * GR[3 4; -3]
```
lkdvos, line 319: I think here there is an interesting difference between planar and non-planar implementations.
I think your current implementation is optimal, unfortunately it is not planar, so will break our fermionic and anyonic code. The reason is that in planar mode we only allow cyclic permutations, which is not the case in the intermediate:
`GLWx[-1 -2 3; 4] * GR[3 4; -3]` has to cross legs 3 and 4 (which is precisely the uncrossing of the permutation you did in the constructor).
The interesting part is that we have to permute `GLWx` anyways, to bring it back into block-matrix form, but from a first guess this order is actually more beneficial... We can discuss tomorrow a bit more if you like.

TLDR, I think it is probably easier for now to keep `GR` unpermuted, and therefore also change the contraction to:
 
```suggestion
        @plansor y[-1 -2; -3] += GLW[-1 -2 3; 1 2] * x[1 2; 4] * GR[4 3; -3]
```

src/algorithms/derivatives/hamiltonian_derivatives.jl
```diff
@@ -323,11 +334,9 @@ function (H::JordanMPO_AC2_Hamiltonian)(x::MPOTensor)
...
     ismissing(H.EE) || @plansor y[-1 -2; -3 -4] += x[1 -2; -3 -4] * H.EE[-1; 1]
 
-    GL, A1, A2, GR = H.AA
-    if nonzero_length(A1) > 0 && nonzero_length(A2) > 0
-        # TODO: there are too many entries here, this could be further optimized
-        @plansor y[-1 -2; -3 -4] += GL[-1 7; 6] * x[6 5; 1 3] * A1[7 -2; 5 4] *
-            A2[4 -4; 3 2] * GR[1 2; -3]
+    if !ismissing(H.AA)
+        GLW, GWR = H.AA
+        @tensor y[-1 -2; -3 -4] += GLW[-1 -2 3; 1 2] * x[1 2; 4 5] * GWR[3 4 5; -3 -4]
```
lkdvos, line 339: Same comment here, you want to construct GWR such that the contraction ends up as:

```suggestion
        @plansor y[-1 -2; -3 -4] += GLW[-1 -2 5; 1 2] * x[1 2; 3 4] * GWR[3 4 5; -3 -4]
```

src/algorithms/derivatives/hamiltonian_derivatives.jl
```diff
@@ -138,9 +138,12 @@ function AC_hamiltonian(
         finished = removeunit(GL[end], 2)
     end
 
-    # continuing
-    A = W.A
-    continuing = (GL[2:(end - 1)], A, GR[2:(end - 1)])
+    if nonzero_length(W.A) > 0
+        @tensor GLW[-1 -2 -3; -4 -5] ≔ GL[2:(end - 1)][-1 1; -4] * W.A[1 -2; -5 -3]
```
lkdvos, line 142:
```suggestion
        @plansor GLW[-1 -2 -3; -4 -5] ≔ GL[2:(end - 1)][-1 1; -4] * W.A[1 -2; -5 -3]
```
This is a fermion/anyon thing, I can explain tomorrow if you like.

src/algorithms/derivatives/hamiltonian_derivatives.jl
```diff
@@ -138,9 +138,12 @@ function AC_hamiltonian(
...
+        continuing = (GLW, permute(GR[2:(end - 1)], (2, 1), (3,)))
```
lkdvos, line 143:
```suggestion
        continuing = (GLW, GR[2:(end - 1)])
```
See comment below

src/algorithms/derivatives/hamiltonian_derivatives.jl
```diff
@@ -287,7 +290,15 @@ function AC2_hamiltonian(
 
     # continuing - continuing
     # TODO: MPODerivativeOperator code reuse + optimization
-    AA = (GL[2:(end - 1)], W1.A, W2.A, GR[2:(end - 1)])
+    # TODO: One should only calculate these operators if necessary. One can do this by checking nonzero_length(A1) > 0 && nonzero_length(A2) > 0 and then setting AA=missing or as we discussed via bit matrix multiplication.
+    ## TODO: Think about how one could and whether one should store these objects and use them for (a) advancing environments in iDMRG, (b) reuse ind backwards-sweep in IDMRG, (c) subspace expansion
+    if nonzero_length(W1.A) > 0 && nonzero_length(W2.A) > 0
+        @tensor GLW[-1 -2 -3; -4 -5] ≔ GL[2:(end - 1)][-1 1; -4] * W1.A[1 -2; -5 -3]
+        @tensor GWR[-1 -2 -3; -4 -5] ≔ W2.A[-1 -5; -3 1] * GR[2:(end - 1)][-2 1; -4]
```
lkdvos, line 297:
```suggestion
        @tensor GWR[-1 -2 -3; -4 -5] ≔ W2.A[-2 -5; -3 1] * GR[2:(end - 1)][-1 1; -4]
```
See comment below

lkdvos (commented):
Amazing, thanks for the quick work! Do you already have some idea of the performance difference for your specific case?

I'll try and come up with some reasonable set of cases to actually benchmark this on, I'd like to experiment with how much of a difference it actually makes even for the more modest MPO sizes too.

src/algorithms/derivatives/hamiltonian_derivatives.jl
AFeuerpfeil: The correct planar contraction should be 
`@plansor GWR[-1 -2 -3; -4 -5] ≔ W2.A[-3 -5; -2 1] * GR[2:(end - 1)][-1 1; -4]`
right?

AFeuerpfeil:
Thanks for your comments.
Making the contractions planar instead of choosing the optimal contraction and permutation order leads to an increase in computation time and total allocations of around 30-50%.
Let's discuss tomorrow how to go about this.

AFeuerpfeil:
(@lkdvos: Please ignore the latest commit, I forgot to create a new branch)

PR #335, opened 2025-10-21
Various small fixes and improvements to `RandExpand`

lkdvos:
- The `changebonds(state, ::RandExpand)` algorithm now no longer needs to multiply and decompose a random twosite state, and instead generates the correct vectors directly.
- The dynamic tolerances now properly handle `iter = 0`
- The `InfiniteMPOHamiltonian` environments now properly clamp the Krylov dimension if the MPS bond dimension is low.

### DIFF
--- a/src/algorithms/changebonds/randexpand.jl
+++ b/src/algorithms/changebonds/randexpand.jl
@@ -36,7 +36,7 @@ function changebonds!(ψ::InfiniteMPS, alg::RandExpand)
         XL = randisometry(scalartype(VL), right_virtualspace(VL) ← V)
         AL′[i] = VL * XL
         XR = randisometry(scalartype(VR), space(VR, 1) ← V)
-        AR′[i + 1] = XR * VR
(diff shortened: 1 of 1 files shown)


AFeuerpfeil:
Thanks. The error is gone, but at least in my examples the GMRES for the environment does not converge after expanding leading to weird errors and Lapack Exceptions, when the local update is ill-defined.
Let's discuss that tomorrow, if you have some time in the afternoon.

lkdvos:
@AFeuerpfeil as we discussed I abandoned this approach, after realizing that the proper way of doing this is to sequentially expand and sweep through the unit cell. I will open a new PR for that.

lkdvos:
I'm currently trying that out, but I felt like at that point it no longer fits the changes I'm doing to `changebonds` so I didn't want to mix this in here. (Basically I want to still merge the small improvements I have here but start fresh on the other ideas)

VictorVanthilt (approved):
G2G for me!

PR #339, opened 2025-10-30
[Feature] MPS manifolds

lkdvos:
I have for a long time wanted to change the way our mps are constructed, and I finally started to have a first go at the idea I had in mind.
Disclaimer: everything is up for debate, including names and functionality.

The main idea is that I want to mimic what `Base` and `TensorKit` do for their constructors, which basically boils down to having
```julia
rand([T], structure) -> output
```
In order to achieve this, I therefore added some structure to dispatch on, which I tentatively named `MPSManifold`.
This is just a glorified collection of all of the spaces, but I would argue that it nicely organizes a bit more of the logic surrounding spaces away from the logic concerning the actual constructors.
The updated workflow would now be:
```julia
manifold = FiniteMPSManifold(pspaces, vspaces)
mps = rand(Float64, manifold)
```

I like that it separates out some code logic, especially surrounding the "full-rank-ness" of MPS, and I like the syntax, and I love that we no longer have a million weird constructors that would have to be supported, but before I go in and update the docs, and do the same for the `InfiniteMPS`, I would like some feedback.
Does this make sense? Is it too convoluted? Do you think it is an improvement?

It would be fairly straightforward to add various utility functions to this interface as well, such as `vcat`, `repeat`, ...

### DIFF
--- a/docs/src/man/states.md
+++ b/docs/src/man/states.md
@@ -24,13 +24,13 @@ If you already have the state of interest, it is straightforward to convert it t
 ```
 
 However, typically this is not the case, as storing the full state becomes expensive rather quickly.
-Then, `FiniteMPS` are best constructed by first specifying a [`FiniteMPSManifold`](@ref) that encodes the physical and (maximal) virtual spaces:
+Then, `FiniteMPS` are best constructed by first specifying a [`FiniteMPSStructure`](@ref) that encodes the physical and (maximal) virtual spaces:
 
 ```@example states
 pspaces = fill(ℂ^2, 10)         # physical spaces (Vector)
 max_virtualspace = ℂ^4          # single max virtual space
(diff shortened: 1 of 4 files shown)


VictorVanthilt:
I'm definitely in favor of streamlining the way (in)finiteMPS objects are created. I also like the idea of making it very similar to other construction methods in the ecosystem like `rand(scalartype, abstractwhateverspace)`. I do like the fact that there are a multitude of ways to create an MPS at the moment, like giving a physical dimension, max virtual dim. and a length for example. These are all very easy to parse into the standard `FiniteMPSManifold` form though.

On a related note:
Creating a specific state instead of just a random one is too complicated at the moment in my opinion.
Oftentimes I want to start my simulation from some computational basis state, or from another simple product state and making this by hand is rather cumbersome. Some functionality to make some simple states instead of just random ones would be something i'd like very much and would definitely want to help with.

PR #340, opened 2025-10-30
add AbstractInfiniteMPS

AFeuerpfeil:
In this pull request, I add the abstract type `AbstractInfiniteMPS` and overload the relevant functions for me.
The reason for this abstraction is that I can then define my own `InfiniteMPS` type, which allows me to overload the `getindex` and `setindex!` functions such that they are compatible with MPI without any changes to the algorithms.


AFeuerpfeil:
I might have figured out a solution that does not require a new type for the different `MPS`. I will first investigate that idea and come back to this pull request in case it fails.

PR #343, opened 2025-10-31
Fix small errors, use IterativeSolver interface for IDMRG and remove unnecessary type requirements

AFeuerpfeil:
In this PR I refactored the IDMRG interface to simplify overloading the update for specific `MPO` and `MPS` types

### DIFF
--- a/src/algorithms/groundstate/idmrg.jl
+++ b/src/algorithms/groundstate/idmrg.jl
@@ -58,7 +58,7 @@ end
 
 
 # Internal state of the IDMRG algorithm
-struct IDMRGState{S, O, E, T<:Number}
+struct IDMRGState{S, O, E, T <: Number}
     mps::S
     operator::O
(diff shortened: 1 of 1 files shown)


AFeuerpfeil:
@lkdvos 
Is there a reason, why you use
```julia 
# write in terms of MultilineMPS
function changebonds(
        ψ::InfiniteMPS, operator::InfiniteMPO, alg, envs = environments(ψ, operator)
    )
    ψ′, envs′ = changebonds(
        convert(MultilineMPS, ψ), convert(MultilineMPO, operator), alg, Multiline([envs])
    )
    return convert(InfiniteMPS, ψ′), envs
end
```
only for an `InfiniteMPO` and not also for an `InfiniteMPOHamiltonian`?
As far as I can tell both use the exact same algorithm, so one could switch to the Multiline case anyway.

PR #345, opened 2025-11-03
[perf] Benchmarks and AC/AC2 contraction improvements

lkdvos:
This is a combination of setting up a more thorough benchmark suite to allow further testing in the future, and various changes for our AC and AC2 contractions.

As already pointed out in #270, and a follow-up in #330 when we are performing repeated applications of the effective local hamiltonian, it can actually be beneficial to alter the contraction order to hoist certain contractions out of the loop.
This is especially important for symmetries, as this avoids doing more permutations within the loop, which becomes increasingly important.

With this in mind, I first refactored the way we are "preparing the operators" before `fixedpoint` is invoked. Basically, I added an optional hook there to do some pre- and post-processing on the operator and initial vector precisely at the point where we know there will be a repeated application of the operator.
This also provides the entry point of the benchmark suite, to be able to gauge the preparation time vs application time.

Then, I've added 3 updated contraction schemes, each with different optimizations.

### 0. Before the changes

The current state of affairs is that there are two different implementations, one for MPO's and one for Hamiltonians. The Hamiltonian implementation attempts to maximally make use of the specific structure of the `JordanMPOTensor`s, by more or less manually writing out the different contraction paths.

From the tests and benchmarks, this seems to basically only matter for (next-)nearest-neighbor layouts, and the added code complexity should definitely be taken into account.
Additionally, this implementation silently assumes that the environments are coming from MPS that are properly gauged, as it replaces `GL[1]` and `GR[end]` with identities to avoid having to do the contractions.

### I. Precomputed derivatives

This first implementation is just the simple contraction order change, from `(((GL * x) * O) * O) * GR` to `((GL * O) * x) * (O * GR)`.
This increases the preparation time, but tests seem to show that this contraction order is beneficial over repeated applications, as long as the MPO virtual space is large compared to the MPS physical space.
An additional optimization here is the realization that `GL * O` and `O * GR` are now dense `BlockTensorMap`s, so it pays of to convert them to `TensorMap` outside of the loop to further avoid some overhead.

### II. Precomputed derivatives

The second implementation adds an optimization on top of this for all symmetry classes but the `NoBraiding` ones, by doing a non-planar change of the index order of some of the intermediate tensors
The idea is that when permuting`(GL * O) * x` such that it can be BLAS-contracted with `(O * GR)`, we can either do the planar `repartition(GLOX, 2, 3)`, or the non-planar `permute(GLOX, ((1, 2), (3, 4, 5)))`.
For non-symmetric tensors, it should be obvious that the second is beneficial since it avoids having to do an intermediate permutation altogether, as that permutation is trivial, while the benchmarks seem to show that this effect persists for other symmetries as well.
This is presumably caused by a better data-locality in this permutation vs the repartition.

### III. Precomputed derivatives

The final change is even more radical by additionally fusing some of the indices that can be kept together during the entire contraction.
Basically, the contraction is modified to `(Fl * GL * O * Fl') * (Fl * x * Fr') * (Fr * O * GR * Fr')`, and the resulting eigenvector is unfused in a post-processing step.

There are several benefits to this approach.
The first is that the number of subblocks, and therefore the overhead of dealing with symmetries, is reduced.
The operations can deal with larger contiguous memory regions, therefore further increasing the performance.
A secondary benefit however is that this actually results in `AC` and `AC2` contractions that are indistinguishable from the compilers' perspective, therefore reducing the number of contractions that have to be written and compiled.
This somewhat surprisingly leads to code simplification rather than code complexity, which I definitely like.

---

My benchmark results are added below, where everything is plotted relative to the current status.
I'm not too sure about how to interpret the results.
As suspected, for large MPOs this clearly outperforms the previous implementation, with speedups ranging from 2x to 5x depending on the symmetry and specific bond dimensions, but for smaller MPOs we can clearly see that actually, these "optimizations" are slowing things down.
Theoretically, I think we should expect that as the bond dimension increases further, the effects of changing the contraction order should actually start to dominate and the "normal contraction order" should ultimately win, but it is currently not clear if we can ever reach bond dimensions that reach that regime.

bench_times_relative
bench_memory_relative

I think I am slightly in favor of changing the implementations, mostly because this means that we are relying as much as possible on TensorKit's contractions, so we can focus our efforts concerning multithreading and hardware acceleration in a single place.
Additionally, I obviously also like that this implementation is a lot easier to maintain, and no longer silently assumes that the MPS are properly gauged.

For further reference, here are the same plots, but now including the "preparation time" of the MPOs, by adding `prep_time + n_applications * contract_time + unprep_time`, with n=3 and n=10, which does clearly show that this is only something we want to do for repeated applications.

bench_prep_times_relative_n=3
bench_prep_times_relative_n=10

### DIFF
--- a/test/algorithms/excitations.jl
+++ b/test/algorithms/excitations.jl
@@ -19,7 +19,7 @@ verbosity_conv = 1
         ψ, envs, _ = find_groundstate(
             ψ, H; maxiter = 400, verbosity = verbosity_conv, tol = 1.0e-10
         )
-        energies, ϕs = @inferred excitations(
+        energies, ϕs = @testinferred excitations(
             H, QuasiparticleAnsatz(), Float64(pi), ψ, envs
         )
         @test energies[1] ≈ 0.41047925 atol = 1.0e-4
@@ -46,7 +46,7 @@ verbosity_conv = 1
         ψ, envs, _ = leading_boundary(
             ψ, H, VUMPS(; maxiter = 400, verbosity = verbosity_conv, tol = 1.0e-10)
         )
-        energies, ϕs = @inferred excitations(
+        energies, ϕs = @testinferred excitations(
             H, QuasiparticleAnsatz(), [0.0, Float64(pi / 2)], ψ, envs; verbosity = 0
         )
         @test abs(energies[1]) > abs(energies[2]) # has a minimum at pi/2
@@ -66,7 +66,7 @@ verbosity_conv = 1
             ψ, envs, = find_groundstate(ψ, H; verbosity)
 
             # find energy with quasiparticle ansatz
-            energies_QP, ϕs = @inferred excitations(H, QuasiparticleAnsatz(), ψ, envs)
+            energies_QP, ϕs = @testinferred excitations(H, QuasiparticleAnsatz(), ψ, envs)
             @test variance(ϕs[1], H) < 1.0e-6
 
             # find energy with normal dmrg
@@ -74,14 +74,14 @@ verbosity_conv = 1
                     DMRG(; verbosity, tol = 1.0e-6),
                     DMRG2(; verbosity, tol = 1.0e-6, trscheme = trunctol(; atol = 1.0e-4)),
                 )
-                energies_dm, _ = @inferred excitations(H, FiniteExcited(; gsalg), ψ)
+                energies_dm, _ = @testinferred excitations(H, FiniteExcited(; gsalg), ψ)
                 @test energies_dm[1] ≈ energies_QP[1] + expectation_value(ψ, H, envs) atol = 1.0e-4
             end
 
             # find energy with Chepiga ansatz
-            energies_ch, _ = @inferred excitations(H, ChepigaAnsatz(), ψ, envs)
+            energies_ch, _ = @testinferred excitations(H, ChepigaAnsatz(), ψ, envs)
             @test energies_ch[1] ≈ energies_QP[1] + expectation_value(ψ, H, envs) atol = 1.0e-4
-            energies_ch2, _ = @inferred excitations(H, ChepigaAnsatz2(), ψ, envs)
+            energies_ch2, _ = @testinferred excitations(H, ChepigaAnsatz2(), ψ, envs)
             @test energies_ch2[1] ≈ energies_QP[1] + expectation_value(ψ, H, envs) atol = 1.0e-4
             return energies_QP[1]
         end


lkdvos:
@leburgel, @VictorVanthilt, @afeuerpfeil, I would love to hear your opinions on this

AFeuerpfeil:
Looks very fascinating, but it is not particularly clear to me, whether one option is objectively better.
I will be back at the institute on Wednesday evening, so we can discuss Wednesday or Thursday.

lkdvos:
I did a bit more work on this, trying to bring back the JordanMPO machinery while also getting the benefits of the precontracted systems. I ran some more benchmarks, for which the results are below.

The versions are:
- v1: base case
- v2: precontract mpo and environment, fuse legs that don't have to be separate, use dedicated matrix contraction to avoid recoupling
- v3: similar to v2, but explicit loop over matrix contractions to avoid having to copy due to unmatching strides in the dedicate matrix contraction routine
- v4: using a buffer

The main things to notice is that the spread is a lot smaller now for the nearest neighbour and next-nearest neighbour cases, which is to be expected since I didn't really touch the JordanMPO specializations that are mostly useful for that. Looking at the exact numbers, I would even dare say that some of that is likely just due to noise in the data, so I would claim the performance in these cases is unchanged.  
I do still have the massive speedup of the larger MPO cases, which is why I would argue that once I finish this PR it is actually worth it to go and merge in its current state.

<details> <summary> Relative time of prepared application </summary>

prepared_times_rel


</details>

<details> <summary> Relative memory of prepared application </summary>

prepared_memory_rel


</details>

A different thing to look at is the "unprepared application" vs the "preparation + prepared application" times.
To gauge this, I separated out part of the JordanMPO optimizations I did as well, to really show even in the base case what the costs are that are associated to this.
In the plots below, I'm showing only the unprepared contraction for v1 and v4, and then showing the time per application, including the preparation time amortized over either 1, 3 or 10 applications. It's a bit harder to see, but Ive already spent too much time trying to visualize this properly and the main point is that I think the v4 is overall a good fit.
I'll go ahead and iron out the kinks that make the tests fail, but otherwise I would be happy to include these changes

<details> <summary> Times for n=1 application </summary>

amortized_times_1

</details>

<details> <summary> Times for n=3 application </summary>

amortized_times_3

</details>

<details> <summary> Times for n=10 application </summary>

amortized_times_10

</details>

AFeuerpfeil:
Sounds good!
Looking forward to it.

PR #347, opened 2025-11-06
add Base.isfinite to mps types

AFeuerpfeil:
This PR adds `Base.isfinite` to the different MPS types such that algorithms such as VUMPS or IDMRG can check the correctness of the input data without restricting the types to only e.g. `InfiniteMPS`.


AFeuerpfeil:
Sure!

PR #350, opened 2025-11-06
Update WindowMPS example

VicVanderLinden:
Updating the example of windowMPS so it can run in latest version

### DIFF
--- a/examples/windowmps.jl
+++ b/examples/windowmps.jl
@@ -24,11 +24,11 @@ let
     envs = environments(mpco, th)
 
     szdat = [expectation_value(mpco, i => sz) for i in 1:length(mpco)]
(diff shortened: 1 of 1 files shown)


lkdvos:
Thanks for helping out, I greatly appreciate this!
It seems like there are still some small formatting changes required, would you mind also changing that?

In the future, it would be nice to promote this to a full example that we can actually include in the documentation. If this is something you are interested in, I'm definitely happy to help you get that set up.

VicVanderLinden:
Hi, thanks! I am still getting the hang of Github but I definitely would be open to doing something like that. Maybe an entire example showing the WindowMPS for an improved version of https://arxiv.org/abs/1907.02474 would be nice?

PR #352, opened 2025-11-12
Add `OperatorStyle` and `GeometryStyle`

AFeuerpfeil:
This PR is the start of some work to make it possible/easier to add custom state or operator types without having to alter their base implementations, and without having to redefine a bunch of functionality that already just works.

The main idea is to avoid having to encode distinctions between finite/infinite or hamiltonian/mpo as hard type constants or abstract types, since this often clashes with Julia's type tree architecture (you cannot subtype multiple things).
Instead the idea is to use traits for dispatching, and gradually start loosening the type restrictions on the inputs of various functions here.

The plan is to gradually start incorporating this strategy, to avoid having a massive PR that has to change everything at once, and keep this non-breaking.
This PR is the first part, which simply introduces the traits.

### DIFF
--- a/test/other.jl
+++ b/test/other.jl
@@ -78,7 +78,6 @@ module TestMiscellaneous
         ... 🭻⎡⠉⢈⎤🭻 ...
              ⎣⠀⢀⎦ 
         """
-
         @test output == check
 
         O = make_time_mpo(H, 1.0, TaylorCluster(3, false, false))
@@ -88,7 +87,6 @@ module TestMiscellaneous
         ... 🭻⎡⡏⠉⠛⠟⎤🭻 ...
              ⎣⡇⠀⠀⡂⎦ 
         """
-
         @test output == check
 
         # Finite Hamiltonians and MPOs
--- a/src/MPSKit.jl
+++ b/src/MPSKit.jl
@@ -105,8 +105,6 @@ include("utility/utility.jl") # random utility functions
 include("utility/plotting.jl")
 include("utility/linearcombination.jl")
 
-
-
 # maybe we should introduce an abstract state type
 include("states/abstractmps.jl")
 include("states/infinitemps.jl")


AFeuerpfeil:
One question I came up with when rewriting the environment constructors is:
How should we implement the logic that checks that e.g. `above`, `below`, `operator` all have the same `GeometryStyle`?

In order to not always have to write e.g. 
```julia
@assert allequal([GeometryStyle(above),GeometryStyle(below),GeometryStyle(operator)])
```

I would argue for overloading e.g. `Base.:&` to add this logic in the following way:
```julia
Base.:&(::FiniteStyle, ::FiniteStyle) = FiniteStyle()
Base.:&(::InfiniteStyle, ::InfiniteStyle) = InfiniteStyle()
```
Then, one can call 
```julia
geometry_style = GeometryStyle(above) & GeometryStyle(below) & GeometryStyle(operator)
```
which will either give the common style of the 3 or error, if they are not identical. (If we want to throw a specific kind of error, one can also overload 
```julia
Base.:&(::FiniteStyle, ::InfiniteStyle) = throw(...)
Base.:&(::InfiniteStyle, ::FiniteStyle) = throw(...)
```

test/other.jl
```diff
@@ -72,19 +74,15 @@ module TestMiscellaneous
         buffer = IOBuffer()
         braille(buffer, H)
         output = String(take!(buffer))
-        check = """
-        ... 🭻⎡⠉⢈⎤🭻 ...
-             ⎣⠀⢀⎦ 
-        """
+        check = "... 🭻⎡⠉⢈⎤🭻 ...\n     ⎣⠀⢀⎦ \n"
+
```
lkdvos: Why this change? I don't think it is related and I think the previous version was more read-friendly

src/states/multilinemps.jl
```diff
@@ -101,7 +101,7 @@ Base.convert(::Type{InfiniteMPS}, st::MultilineMPS) = only(st)
 Base.eltype(t::MultilineMPS) = eltype(t[1])
 Base.copy!(ψ::MultilineMPS, ϕ::MultilineMPS) = (copy!.(parent(ψ), parent(ϕ)); ψ)
 
-Base.isfinite(ψ::MultilineMPS) = false
+Base.isfinite(::Type{<:MultilineMPS}) = false
```
lkdvos: Here you probably also need the `isfinite(x) = isfinite(typeof(x))`, since `MultilineMPS` is not an `AbstractMPS`.

src/utility/styles.jl
```diff
@@ -0,0 +1,43 @@
+"""
+`OperatorStyle`
```
lkdvos:
```suggestion
    abstract type OperatorStyle end
```

src/utility/styles.jl
```diff
@@ -0,0 +1,43 @@
...
+
+Holy trait used as a dispatch tag for operator representations.
+Concrete subtypes (`MPOStyle` and `HamiltonianStyle`) indicate
+whether an operator is stored as an MPO or as a Hamiltonian.
```
lkdvos:
```suggestion
whether an operator represents a Hamiltonian operator (sum of terms) or a transfer matrix (product of factors).
```

test/other.jl
AFeuerpfeil: After adding a new testset to other.jl, the old version failed in all tests because it corresponded to `"... 🭻⎡⠉⢈⎤🭻 ...\n     ⎣⠀⢀⎦\n"`, i.e. a missing space in the end.
The runic formatting also failed, which is why I changed it.

src/utility/styles.jl
VictorVanthilt: I find it rather confusing to call regular dense MPOs a "transfermatrix". Can you explain the reasoning here?

PR #356, opened 2025-11-19
Implement Styles for `AbstractMPS`

AFeuerpfeil:
This PR implements the methods `AC2`, `dot`, `isapprox`, `norm`, `normalize(!)` as well as the left and right fixedpoints for an `AbstractMPS` by dispatching on `GeometryStyle`. These functions are currently defined for `FiniteMPS` and `InfiniteMPS`. For other types of `AbstractMPS`, there exist more specialized functions, so the dispatch on `GeometryStyle` is skipped making these changes non-breaking.

Examples for the `Style` system are
```julia 
function TensorKit.dot(::FiniteChainStyle, ψ₁::AbstractMPS, ψ₂::AbstractMPS)
    #todo : rewrite this without having to gauge
    length(ψ₁) == length(ψ₂) || throw(ArgumentError("MPS with different length"))
    ρr = TransferMatrix(ψ₂.AR[2:end], ψ₁.AR[2:end]) * r_RR(ψ₂)
    return tr(_transpose_front(ψ₁.AC[1])' * _transpose_front(ψ₂.AC[1]) * ρr)
end
```
for the `FiniteMPS` and 
```julia 
function AC2(::InfiniteChainStyle, ψ::AbstractMPS, i::Integer; kind = :ACAR)
    if kind == :ACAR
        return ψ.AC[i] * _transpose_tail(ψ.AR[i + 1])
    elseif kind == :ALAC
        return ψ.AL[i] * _transpose_tail(ψ.AC[i + 1])
    else
        throw(ArgumentError("Invalid kind: $kind"))
    end
end
```
for the `InfiniteMPS`.

### DIFF
--- a/src/states/abstractmps.jl
+++ b/src/states/abstractmps.jl
@@ -256,8 +256,7 @@ TensorKit utility
 ===========================================================================================#
 
 function TensorKit.dot(ψ₁::AbstractMPS, ψ₂::AbstractMPS; kwargs...)
-    geometry_style = GeometryStyle(ψ₁, ψ₂)
-    return TensorKit.dot(geometry_style, ψ₁, ψ₂; kwargs...)
+    return TensorKit.dot(GeometryStyle(ψ₁, ψ₂), ψ₁, ψ₂; kwargs...)
 end
 function Base.isapprox(ψ₁::AbstractMPS, ψ₂::AbstractMPS; kwargs...)
     return isapprox(dot(ψ₁, ψ₂), 1; kwargs...)
--- a/src/states/finitemps.jl
+++ b/src/states/finitemps.jl
@@ -337,7 +337,7 @@ Base.@propagate_inbounds function Base.getindex(ψ::FiniteMPS, i::Int)
 end
 
 # TODO: check where gauge center is to determine efficient kind
-AC2(psi::FiniteMPS, site::Int) = psi.AC[site] * _transpose_tail(psi.AR[site + 1])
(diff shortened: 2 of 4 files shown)


src/states/abstractmps.jl
```diff
@@ -245,4 +234,93 @@ physicalspace(ψ::AbstractMPS) = map(Base.Fix1(physicalspace, ψ), eachsite(ψ))
 
 Return an iterator over the sites of the MPS `state`.
 """
-eachsite(ψ::AbstractMPS) = eachindex(ψ)
+eachsite(ψ::AbstractMPS) = eachsite(GeometryStyle(ψ), ψ)
+
+eachsite(::GeometryStyle, ψ::AbstractMPS) = eachindex(ψ)
```
lkdvos: Is there a benefit to having this go through the geometrystyle? It looks like we end up with the same fallback anyways, so I don't really see the point here

lkdvos (commented):
I'm a bit confused by the choice of methods you choose to dispatch through the style.

I was expecting this to be more applicable to the higher-level functions, rather than these very low-level functions instead. Note for example that you are still assuming that the `AbstractMPS` has fields `AC` and `AR` etc, so this isn't necessarily "generalizing" the implementations all that much.

Also for the `normalize!` function I'm a bit surprised, since in principle that isn't even very well-defined for the infinite case to begin with.

We can maybe rediscuss this a bit more on Monday, but I feel like in general this might not be the right kind of functions to look at, and rather we should focus on the larger algorithms instead.
Note that the goal is not to push everything through the styles, this really only makes sense if we can find implementations that are generic enough to begin with, since otherwise we will have to overload each specific case anyways.

PR #360, opened 2025-12-02
[WIP] Implement styles in a top-to-bottom approach

AFeuerpfeil:
I am currently working on getting the algorithms to run for user-defined `AbstractMPS` types.
To this end, I will go through the algorithms (currently focussing on `IDMRG` and `VUMPS`) and implement the `Style` interface for the highest-level functions, which are called in these algorithms.

Examples for these are `calc_galerkin`, `expectation_value`, `environments`, `recalculate!`, `(A)C(2)_hamiltonian`, `transfer_leftenv!`, `transfer_rightenv!`.

### DIFF
--- a/src/algorithms/derivatives/hamiltonian_derivatives.jl
+++ b/src/algorithms/derivatives/hamiltonian_derivatives.jl
@@ -69,15 +69,8 @@ end
 
 # Constructors
 # ------------
-const _HAM_MPS_TYPES = Union{
-    FiniteMPS{<:MPSTensor},
-    WindowMPS{<:MPSTensor},
-    InfiniteMPS{<:MPSTensor},
-}
-
-function AC_hamiltonian(
-        site::Int, below::_HAM_MPS_TYPES, operator::MPOHamiltonian{<:JordanMPOTensor},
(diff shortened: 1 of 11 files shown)


lkdvos (commented):
This looks great, I really like that we can relax the input types.

AFeuerpfeil:
It definitely feels like it is going in the right direction.
In my current fork, it is not reducing the size of the code but is also not adding too much boilerplate.
One question I came up with is the following:
Do we at all want to restrict input types to `AbstractMPS`, `AbstractMPO`, `AbstractMPSEnvironments` or just leave that completely open to the user?
And if we want to restrict it, then at which level, only at the topmost function, which dispatches onto the function defined on `Styles` or in all levels?

An example, how the restriction only for the topmost function in the call tree looks like, is
```julia
function environments(above::AbstractMPS, op::AbstractMPO, below::AbstractMPS)
    return environments(GeometryStyle(above, op, below), OperatorStyle(op), above, op, below)
end
```
and then
```julia
function environments(::InfiniteChainStyle, ::HamiltonianStyle, above, op, below)
 ...
end
```

PR #364, opened 2025-12-12
Start on updates for TK 0.16 and MAK 0.6

kshyatt:
Saw an error deeper in the quasiparticle tests locally, let's see how this goes. I would also like to update this utility code to be GPU friendly in advance :)


kshyatt:
Don't worry lol it was like 5 minutes of work

lkdvos:
I'll push when I get to the office, but tldr I think we can delete that code entirely since I introduced that before the orthnull rewrite to temporarily patch that, which is now just handled by the base implementation

PR #367, opened 2025-12-16
More informative errors for finite mps

borisdevos:
Small silly change, but while I was being lazy with variable names and confused my finite MPS with infinite MPS, I realised I could call certain functions and get it to error at weird places. This change fixes that. It seemed good to do since these functions are exported, so the generic method was always available.

Unrelated to this, I was also playing around with Cartesian spaces and `transfer_spectrum` would use the sector `Trivial()` to make a `ComplexSpace`, so now it checks the space type instead of sector type. 

Final thing to add: the eigsolve that happens in `transfer_spectrum` defaults to Arnoldi even though the state is fully real. I didn't change this, because I simply don't know how much support we want for real MPSs.

### DIFF
--- a/src/algorithms/toolbox.jl
+++ b/src/algorithms/toolbox.jl
@@ -151,7 +151,7 @@ function marek_gap(above::InfiniteMPS; tol_angle = 0.1, kwargs...)
     return marek_gap(spectrum; tol_angle)
 end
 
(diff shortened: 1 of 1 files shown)


borisdevos:
I'm confused as well now, because isn't that what the method error does?

lkdvos (approved):

PR #375, opened 2026-02-06
Changes for non-CPU array support

kshyatt:
Mostly allowing passing an array type instead of just an element type to allow GPU arrays to back MPS tensors

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -76,6 +76,3 @@ cuTENSOR = "011b41b2-24ef-40a8-b3eb-fa098493e9e1"
 
 [targets]
 test = ["Aqua", "Adapt", "CUDA", "cuTENSOR", "Pkg", "Test", "TestExtras", "Plots", "Combinatorics", "ParallelTestRunner", "TensorKitTensors"]
-
-[sources]
-TensorKit = {url="https://github.com/QuantumKitHub/TensorKit.jl", rev="ksh/cuda_tweaks"}
--- a/Project.toml
+++ b/Project.toml
@@ -76,3 +76,6 @@ cuTENSOR = "011b41b2-24ef-40a8-b3eb-fa098493e9e1"
 
 [targets]
 test = ["Aqua", "Adapt", "CUDA", "cuTENSOR", "Pkg", "Test", "TestExtras", "Plots", "Combinatorics", "ParallelTestRunner", "TensorKitTensors"]
+
+[sources]
+TensorKit = {url="https://github.com/QuantumKitHub/TensorKit.jl", rev="ksh/cuda_tweaks"}
--- a/src/operators/mpohamiltonian.jl
+++ b/src/operators/mpohamiltonian.jl
@@ -32,6 +32,7 @@ struct MPOHamiltonian{TO <: JordanMPOTensor, V <: AbstractVector{TO}} <: Abstrac
     W::V
 end
 OperatorStyle(::Type{<:MPOHamiltonian}) = HamiltonianStyle()
+TensorKit.storagetype(::Type{MPOHamiltonian{O, V}}) where {O, V} = storagetype(O)
 
 const FiniteMPOHamiltonian{O <: MPOTensor} = MPOHamiltonian{O, Vector{O}}
 Base.isfinite(::Type{<:FiniteMPOHamiltonian}) = true
--- a/src/states/quasiparticle_state.jl
+++ b/src/states/quasiparticle_state.jl
@@ -1,7 +1,7 @@
 #=
 Should not be constructed by the user - acts like a vector (used in eigsolve)
 I think it makes sense to see these things as an actual state instead of return an array of B tensors (what we used to do)
-This will allow us to plot energy density (finite qp) and measure observeables.
+This will allow us to plot energy density (finite qp) and measure observables.
 =#
 
 struct LeftGaugedQP{S, T1, T2, E <: Number}
--- a/src/utility/multiline.jl
+++ b/src/utility/multiline.jl
@@ -106,3 +106,14 @@ function VectorInterface.inner(x::Multiline, y::Multiline)
 end
 
 LinearAlgebra.norm(x::Multiline) = sqrt(real(inner(x, x)))
+
+# TensorKit
+#----------
+
+site_type(::Type{Multiline{S}}) where {S} = site_type(S)
+bond_type(::Type{Multiline{S}}) where {S} = bond_type(S)
+site_type(st::Multiline) = site_type(typeof(st))
+bond_type(st::Multiline) = bond_type(typeof(st))
+TensorKit.sectortype(::Type{Multiline{T}}) where {T} = sectortype(T)
+TensorKit.spacetype(::Type{Multiline{T}}) where {T} = spacetype(T)
+TensorKit.storagetype(::Type{Multiline{T}}) where {T} = storagetype(T)
--- a/test/cuda/operators.jl
+++ b/test/cuda/operators.jl
@@ -17,6 +17,10 @@ MPSKit.Defaults.alg_svd() = CUSOLVER_QRIteration()
     O₁ = rand(T, V^L, V^L)
     O₂ = rand(T, space(O₁))
     O₃ = rand(real(T), space(O₁))
+    
+    dO₁ = adapt(CuArray, O₁)
+    dO₂ = adapt(CuArray, O₂)
+    dO₃ = adapt(CuArray, O₃)
 
     mpo₁ = adapt(CuVector{T, CUDA.DeviceMemory}, FiniteMPO(O₁))
     mpo₂ = adapt(CuVector{T, CUDA.DeviceMemory}, FiniteMPO(O₂))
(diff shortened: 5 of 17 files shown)


src/environments/abstract_envs.jl
```diff
@@ -15,48 +15,52 @@ Base.unlock(envs::AbstractMPSEnvironments) = unlock(envs.lock);
 # ------------------
 function allocate_GL(bra::AbstractMPS, mpo::AbstractMPO, ket::AbstractMPS, i::Int)
     T = Base.promote_type(scalartype(bra), scalartype(mpo), scalartype(ket))
+    TA = similarstoragetype(storagetype(mpo), T)
```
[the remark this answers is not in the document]
kshyatt: Sure, I can modify this to take into account the backing types of `bra` and `ket` too

lkdvos (commented):
Since I stole the commits from here, I thought I'd might as well also go through and indicate some merge changes that were probably no longer needed, and meanwhile left some more comments

src/states/quasiparticle_state.jl
```diff
@@ -216,6 +216,7 @@ GeometryStyle(::Type{<:QP{S, T1, T2}}) where {S, T1, T2} = GeometryStyle(S)
 
 TensorKit.spacetype(::Union{QP{S}, Type{<:QP{S}}}) where {S} = spacetype(S)
 TensorKit.sectortype(::Union{QP{S}, Type{<:QP{S}}}) where {S} = sectortype(S)
+TensorKit.storagetype(qp::QP{S, T1, T2}) where {S, T1, T2} = storagetype(T2)
```
lkdvos: Is this required? this should really be `storagetype(qp) = storagetype(typeof(qp))`, and then being captured by the definion just below, and I had hoped this would be automatically happening because of TensorKit having a fallback?

src/utility/multiline.jl
```diff
@@ -18,6 +18,8 @@ struct Multiline{T}
 end
 Multiline(data::AbstractVector{T}) where {T} = Multiline{T}(data)
 
+TensorKit.storagetype(M::Multiline{T}) where {T} = storagetype(T)
```
lkdvos: As mentioned before, I would like to avoid this definition, and only have it in the case of `MultilineMPS` or `MultilineMPO` specifically

kshyatt: Is there some reason then that `sectortype` and `spacetype` are defined for generic `Multiline` (NOT `MultilineMPS`) in `src/states/multilinemps.jl`? It seems kind of inconsistent to me.

lkdvos: It probably just is kind of inconsistent, but I think what annoys me with `storagetype(::Multiline)` could also mean the typeof `m.data`, although I guess there never is any reason for that to not be `Array` so maybe I can just let this one go. Sorry for being annoying about this, you are probably right that it is better to leave this in...

kshyatt: No worries, I know I'm also being annoying by getting my grubby little fingers all over everything. Why don't I move the generic methods out to the `src/utility/multiline.jl` file?

src/operators/mpohamiltonian.jl
```diff
@@ -32,6 +32,8 @@ struct MPOHamiltonian{TO <: JordanMPOTensor, V <: AbstractVector{TO}} <: Abstrac
     W::V
 end
 OperatorStyle(::Type{<:MPOHamiltonian}) = HamiltonianStyle()
+TensorKit.storagetype(mpo::MPOHamiltonian{O, V}) where {O, V} = storagetype(O)
```
lkdvos:
```suggestion
```
this should automatically dispatch to the line below:
https://github.com/QuantumKitHub/TensorKit.jl/blob/9422cd5c1f0521b48501aa0056f093369b3acfff/src/tensors/abstracttensor.jl#L52

src/utility/multiline.jl
```diff
@@ -106,3 +106,16 @@ function VectorInterface.inner(x::Multiline, y::Multiline)
...
+# TensorKit
+#----------
+
+site_type(::Type{Multiline{S}}) where {S} = site_type(S)
+bond_type(::Type{Multiline{S}}) where {S} = bond_type(S)
+site_type(st::Multiline) = site_type(typeof(st))
+bond_type(st::Multiline) = bond_type(typeof(st))
+TensorKit.sectortype(t::Multiline) = sectortype(typeof(t))
+TensorKit.sectortype(::Type{Multiline{T}}) where {T} = sectortype(T)
+TensorKit.spacetype(t::Multiline) = spacetype(typeof(t))
```
lkdvos:
```suggestion
```

src/utility/multiline.jl
```diff
@@ -106,3 +106,16 @@ function VectorInterface.inner(x::Multiline, y::Multiline)
...
 LinearAlgebra.norm(x::Multiline) = sqrt(real(inner(x, x)))
+
+# TensorKit
+#----------
+
+site_type(::Type{Multiline{S}}) where {S} = site_type(S)
+bond_type(::Type{Multiline{S}}) where {S} = bond_type(S)
+site_type(st::Multiline) = site_type(typeof(st))
+bond_type(st::Multiline) = bond_type(typeof(st))
+TensorKit.sectortype(t::Multiline) = sectortype(typeof(t))
```
lkdvos:
```suggestion
```

src/utility/multiline.jl
```diff
@@ -106,3 +106,16 @@ function VectorInterface.inner(x::Multiline, y::Multiline)
...
+TensorKit.sectortype(::Type{Multiline{T}}) where {T} = sectortype(T)
+TensorKit.spacetype(t::Multiline) = spacetype(typeof(t))
+TensorKit.spacetype(::Type{Multiline{T}}) where {T} = spacetype(T)
+TensorKit.storagetype(M::Multiline{T}) where {T} = storagetype(T)
```
lkdvos: This one should be in the type domain

lkdvos (commented):
I think overall this looks good to go, except for the fact that I don't think it makes sense to include all examples for a GPU version as well.
I do agree that there must be examples about GPU support, but I don't think it makes much sense to translate all of them, since we then also have to maintain all of them 🙃 
What I would like instead is for the tests to include some more GPU cases, but I guess that is a way bigger endeavor so I also am happy to just include the changes without that for now

kshyatt:
I 100% agree about the examples, IMO let me cleanup the multiline stuff a little and try to write a few tests that will actually cover the changes.

kshyatt:
Some of the missing coverage here will be fixed once the Algorithm and Driver changes hit at MatrixAlgebraKit, I'll try to add a few more tests for the MPOs though

Project.toml
```diff
@@ -76,3 +76,7 @@ cuTENSOR = "011b41b2-24ef-40a8-b3eb-fa098493e9e1"
 
 [targets]
 test = ["Aqua", "Adapt", "CUDA", "cuTENSOR", "Pkg", "Test", "TestExtras", "Plots", "Combinatorics", "ParallelTestRunner", "TensorKitTensors"]
+
+[sources]
+TensorKit = {url="https://github.com/QuantumKitHub/TensorKit.jl", rev="ksh/cuda_tweaks"}
+BlockTensorKit = {url="https://github.com/lkdvos/BlockTensorKit.jl", rev="main"}
```
lkdvos:
```suggestion
```
Should now be released, I didn't realize I forgot to tag that 😉

kshyatt:
Some of this needs a fix in TensorOperations, which I'm working on, but I think this can be merged and I'll come back and re-enable the tests affected by `tensoradd`

test/cuda/operators.jl
```diff
@@ -0,0 +1,87 @@
+using .TestSetup
+using Test, TestExtras
+using MPSKit
+using MPSKit: GeometryStyle, FiniteChainStyle, InfiniteChainStyle, OperatorStyle, MPOStyle
+using TensorKit
+using MatrixAlgebraKit
+using TensorKit: ℙ, tensormaptype, TensorMapWithStorage
+using Adapt, CUDA, cuTENSOR
+
+MPSKit.Defaults.alg_svd() = CUSOLVER_QRIteration()
```
lkdvos, line 12: Hahahah I guess this is one way to get around these issues 👍 
Would you mind leaving a TODO note referencing the MatrixAlgebraKit issue for that, so we know to remove this in the future? (also, does this work if Revise is not loaded?)

test/cuda/operators.jl
```diff
@@ -0,0 +1,87 @@
...
+
+    # test addition and multiplication
+    @test convert(TM, mpo₁ + mpo₂) ≈ dO₁ + dO₂
+    @test convert(TM, mpo₁ + mpo₃) ≈ dO₁ + dO₃
+    @test convert(TM, mpo₁ * mpo₂) ≈ dO₁ * dO₂
+    @test convert(TM, mpo₁ * mpo₃) ≈ dO₁ * dO₃
+
+    # test application to a state
+    ψ₁ = adapt(CuArray, rand(T, domain(O₁)))
+    #ψ₂ = adapt(CuArray, rand(real(T), domain(O₂))) # not allowed due to cuTENSOR
```
lkdvos, line 67: what does this mean? Is it because we can't permute with mixed scalartypes?

test/cuda/states.jl
```diff
@@ -0,0 +1,52 @@
...
+
+@testset "CuMPS ($(sectortype(D)), $elt)" for (D, d, elt) in
+    [(ℙ^10, ℙ^2, ComplexF64), (Rep[U₁](1 => 3), Rep[U₁](0 => 1), ComplexF64)]
+    tol = Float64(eps(real(elt)) * 100)
+
+    ψ = adapt(CuArray, InfiniteMPS([rand(elt, D * d, D), rand(elt, D * d, D)]; tol))
+    @test TensorKit.storagetype(ψ) == CuVector{ComplexF64, CUDA.DeviceMemory}
+    @test eltype(ψ) == eltype(typeof(ψ))
+
+    #=for i in 1:length(ψ)
```
lkdvos: Any idea what is missing for this?

lkdvos (approved):
Overall I definitely think we've been going over this for long enough, most of it looks very reasonable to me. I left some final small comments, but otherwise I'd be happy to merge this and continue from there.

test/cuda/operators.jl
kshyatt: Yes and yes, I am actually a total stone age weirdo who doesn't use Revise

test/cuda/operators.jl
kshyatt: Yeah it's not an allowed type combo for CuTensor itself

lkdvos: Fair enough. In principle we should be able to bypass cuTENSOR here since it should really be just a copy, but that isn't too relevant for this PR

kshyatt: We could bypass if we're certain the layouts are the same, right? But otherwise we'd have to go block by block with `copyto!`

test/cuda/states.jl
kshyatt: I forgot to include `cuTENSOR` at the top lol, don't code at 1am, kids...

lkdvos (approved):

PR #377, opened 2026-02-09
Refactor entropy function to also use spectrum directly

AFeuerpfeil:
In this PR, I add the functionality to directly supply the entanglement spectrum obtained from e.g. `entanglement_spectrum(psi, loc)` to calculate the von Neumann entropy.


lkdvos:
Ok, no worries I'm not against this, I would have just been pretty pissed off if Github all of a sudden would decide to just auto comment on all PRs with copilot.

(also, the suggestions are actually not bad 😉 )

PR #389, opened 2026-02-20
Fix Adapt extension

lkdvos:
Might need a release of both BlockTensorKit and TensorKit, will see if this already passes


lkdvos:
Can confirm that it absolutely needs the new TensorKit/BlockTensorKit versions 😉

kshyatt:
> Can confirm that it absolutely needs the new TensorKit/BlockTensorKit versions 😉

Since we're waiting for tags anyway, can we put a `[sources]` block in `Project.toml` to confirm this actually will unbreak 1.11+ at least?

PR #393, opened 2026-02-28
Generalize `regauge!` to `AbstractVector`

AFeuerpfeil:
Generalize `regauge!` to work with `AbstractVector` instead of `Vector`.
(PS: In the current implementation of the algorithms, this is not an issue, as `regauge!` is either only called on single tensors like `regauge!(AC, C; alg = alg_orth)` or broadcasted like `regauge!.(temp_Cs, temp_ACs)`.)

### DIFF
--- a/src/states/ortho.jl
+++ b/src/states/ortho.jl
@@ -168,7 +168,7 @@ function regauge!(
     Q_C, _ = left_orth!(C; alg)
     return mul!(AC, Q_AC, Q_C')
 end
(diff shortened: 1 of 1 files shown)


AFeuerpfeil:
Whatever you prefer, I don't have a clear preference.
Is there any situation where one might want to use the special vector version to multi-thread over the unit cell?
Then one could use the already existing `OhMyThreads.jl` interface.

PR #395, opened 2026-02-28
Generalize `calc_galerkin` to `AbstractMPS`

AFeuerpfeil:
I unified the call to `calc_galerkin` without supplying a position and then lifted the requirements of `Union{InfiniteMPS, FiniteMPS, WindowMPS}` to `AbstractMPS` by keeping the specialized function call for the `MultilineMPS`.

### DIFF
--- a/src/algorithms/toolbox.jl
+++ b/src/algorithms/toolbox.jl
@@ -74,7 +74,7 @@ function calc_galerkin(
         below::AbstractMPS, operator, above, envs
     )
     return maximum(
-        pos -> calc_galerkin(pos, below, operator, above, envs), 
+        pos -> calc_galerkin(pos, below, operator, above, envs),
         CartesianIndices(size(above))
     )
 end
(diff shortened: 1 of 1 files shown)


src/algorithms/toolbox.jl
```diff
@@ -71,13 +71,7 @@ function calc_galerkin(
     return calc_galerkin(col, below[row + 1], operator[row], above[row], envs[row])
 end
 function calc_galerkin(
-        below::Union{InfiniteMPS, FiniteMPS, WindowMPS}, operator, above, envs
-    )
-    return maximum(pos -> calc_galerkin(pos, below, operator, above, envs), 1:length(above))
-end
-function calc_galerkin(
-        below::MultilineMPS, operator::MultilineMPO, above::MultilineMPS,
```
lkdvos: I think `MultilineMPS <: AbstractMPS` is false so you no longer have a definition for `MultilineMPS` here...

lkdvos (approved):

PR #396, opened 2026-02-28
Remove explicit call to `InfiniteMPS` in VUMPS

AFeuerpfeil:
I remove the explicit call to `InfiniteMPS` in `VUMPS` and instead use `gaugefix!` in the `gauge_stetp!`.

### DIFF
--- a/src/algorithms/groundstate/idmrg.jl
+++ b/src/algorithms/groundstate/idmrg.jl
@@ -102,7 +102,7 @@ function find_groundstate(mps, operator, alg::alg_type, envs = environments(mps,
         end
 
         alg_gauge = updatetol(alg.alg_gauge, it.state.iter, it.state.ϵ)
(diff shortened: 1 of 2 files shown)


lkdvos (commented):
While I left a small comment on the implementation here, I think I would be happier with a more fundamental change where instead of the uniform gauge happening in the `InfiniteMPS` constructor, we actually use the dedicated function:
Something along the lines of:
```julia
gaugefix!(similar(it.state.mps), copy(it.state.AR), ...; kwargs...)
```

lkdvos (approved):

lkdvos (approved):

AFeuerpfeil:
I did push the IDMRG change, just messed up the commit message text with another commit. Therefore, it is called "format".
The test errors are due to VUMPS not converging for some reason. I do not really understand why.

lkdvos:
I think I found the issue, let's see if this now works!

PR #397, opened 2026-02-28
Dispatch `environment_alg` on `OperatorStyle`

AFeuerpfeil:
I now dispatch `environment_alg` on `OperatorStyle` (with a special case for quasi particles that we should discuss at some point in the future).


AFeuerpfeil:
@lkdvos As far as I can see this should be ready to merge.

lkdvos (commented):
Is there a reason to include the geometrystyle here? I think this is only relevant for infinite systems, or are you thinking about having different infinite styles?

PR #406, opened 2026-03-26
More tests for CUDA + MPS

kshyatt:
Revealed a bunch more missing things needing fixes, so I will keep this as draft for now.

### DIFF
--- a/src/operators/abstractmpo.jl
+++ b/src/operators/abstractmpo.jl
@@ -99,7 +99,7 @@ Base.:\(α::Number, mpo::AbstractMPO) = scale(mpo, inv(α))
 
 function VectorInterface.scale(mpo::AbstractMPO, α::Number)
     T = VectorInterface.promote_scale(scalartype(mpo), scalartype(α))
-    dst = similar(mpo, T)
+    dst = similar(mpo, TensorKit.similarstoragetype(storagetype(mpo), T))
     return scale!(dst, mpo, α)
 end
 
@@ -153,7 +153,11 @@ Compute the mpo tensor that arises from multiplying MPOs.
 """
 function fuse_mul_mpo(O1, O2)
     TT = promote_type(scalartype(O1), scalartype(O2))
-    T = TensorKit.similarstoragetype(storagetype(O1), TT)
+    T = if O1 isa BraidingTensor
+        TensorKit.similarstoragetype(storagetype(O2), TT)
+    else
+        TensorKit.similarstoragetype(storagetype(O1), TT)
+    end
     F_left = fuser(T, left_virtualspace(O2), left_virtualspace(O1))
     F_right = fuser(T, right_virtualspace(O2), right_virtualspace(O1))
     return _fuse_mpo_mpo(O1, O2, F_left, F_right)
@@ -190,7 +194,7 @@ end
 function add_physical_charge(O::MPOTensor, charge::Sector)
     sectortype(O) === typeof(charge) || throw(SectorMismatch())
     auxspace = Vect[typeof(charge)](charge => 1)'
-    F = fuser(scalartype(O), physicalspace(O), auxspace)
+    F = fuser(storagetype(O), physicalspace(O), auxspace)
     @plansor O_charged[-1 -2; -3 -4] := F[-2; 1 2] *
         O[-1 1; 4 3] * τ[3 2; 5 -4] * conj(F[-3; 4 5])
     return O_charged
--- a/src/operators/abstractmpo.jl
+++ b/src/operators/abstractmpo.jl
@@ -99,7 +99,7 @@ Base.:\(α::Number, mpo::AbstractMPO) = scale(mpo, inv(α))
 
 function VectorInterface.scale(mpo::AbstractMPO, α::Number)
     T = VectorInterface.promote_scale(scalartype(mpo), scalartype(α))
-    dst = similar(mpo, TensorKit.similarstoragetype(storagetype(mpo), T))
+    dst = similar(mpo, T)
     return scale!(dst, mpo, α)
 end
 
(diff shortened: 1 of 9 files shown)


kshyatt:
~~The `algorithms` tests are failing as a consequence of https://github.com/QuantumKitHub/BlockTensorKit.jl/pull/54/changes/677d141f276ddf619f105829bcd1b9d7219a448c, because now `JordanMPOTensor` needs to accept also `Union` types in addition to `AbstractTensorMap`...~~

Fake news: it's due to git eating part of the changes to `main` I was rebasing on top of...

src/operators/jordanmpotensor.jl
```diff
@@ -135,6 +136,10 @@ function jordanmpotensortype(::Type{O}) where {O <: AbstractTensorMap}
 end
 Base.similar(W::JordanMPOTensor, ::Type{TorA}) where {TorA} =
     jordanmpotensortype(spacetype(W), TorA)(undef, space(W))
+Base.similar(W::JordanMPOTensor, V::VectorSpace) =
+    jordanmpotensortype(spacetype(V), storagetype(W))(undef, V)
+Base.similar(W::JordanMPOTensor, V::VectorSpace, ::Type{TorA}) where {TorA} =
+    jordanmpotensortype(spacetype(V), TorA)(undef, V)
```
lkdvos: I think there might be a slight issue here, since `HomSpace` is not actually a subtype of `VectorSpace` so the dispatch probably doesn't work, but there's a secondary problem why I didn't define these methods in the first place.
If you provide a space which is not a (2, 2) space, or alternatively does not have the same in- and out- physicalspaces, then `similar` should probably give you a `SparseBlockTensorMap`, not a `JordanMPOTensor`. This is the same like how `similar(::Diagonal, ::Dims)` doesn't actually give you a `Diagonal` back, because you cannot decide from the types alone what kind of output object you get here.

kshyatt: We might be able to get rid of this now with the changes to `BraidingTensor` anyway

kshyatt: Removed those let's see how this goes

src/operators/abstractmpo.jl
```diff
@@ -99,7 +99,7 @@ Base.:\(α::Number, mpo::AbstractMPO) = scale(mpo, inv(α))
 
 function VectorInterface.scale(mpo::AbstractMPO, α::Number)
     T = VectorInterface.promote_scale(scalartype(mpo), scalartype(α))
-    dst = similar(mpo, T)
+    dst = similar(mpo, TensorKit.similarstoragetype(storagetype(mpo), T))
```
lkdvos: It might be better to update `similar(mpo, ::Type{<:Number})` to do exactly this, so we automatically have this behavior everywhere `similar` is used?

src/operators/abstractmpo.jl
```diff
@@ -153,7 +153,11 @@ Compute the mpo tensor that arises from multiplying MPOs.
 """
 function fuse_mul_mpo(O1, O2)
     TT = promote_type(scalartype(O1), scalartype(O2))
-    T = TensorKit.similarstoragetype(storagetype(O1), TT)
+    T = if O1 isa BraidingTensor
+        TensorKit.similarstoragetype(storagetype(O2), TT)
+    else
+        TensorKit.similarstoragetype(storagetype(O1), TT)
+    end
```
lkdvos:
```suggestion
    T = TensorKit.promote_storagetype(TT, O1, O2)
```

I think we'd want to promote the types, rather than taking the non-braidingtensor here?

src/operators/jordanmpotensor.jl
```diff
@@ -205,7 +205,8 @@ end
 for f in (:real, :complex)
     @eval function Base.$f(W::JordanMPOTensor)
         E = $f(scalartype(W))
-        W′ = JordanMPOTensor{E}(undef, space(W))
+        TE = TensorKit.similarstoragetype(TensorKit.storagetype(W), E)
+        W′ = similar(W, TE)
```
lkdvos: similar comment here, would it make sense to use `similar(W, E)` and implement the `similar` to do the storagetype conversion?

src/operators/ortho.jl
```diff
@@ -11,7 +11,14 @@ function left_canonicalize!(
     d = sqrt(dim(P))
 
     # orthogonalize second column against first
-    WI = removeunit(W[1, 1, 1, 1], 1)
+    Wi = W[1, 1, 1, 1]
+    if Wi isa BraidingTensor
+        Wi′ = removeunit(Wi, 1)
```
lkdvos: Is this because `removeunit(::BraidingTensor)` is losing the storagetype? In that case, might be nice to leave a TODO comment here and open an issue for TensorKit.
If not, it feels like something is wrong and this shouldn't be happening?

test/gpu/cuda/operators.jl
```diff
@@ -83,3 +83,190 @@ using Adapt, CUDA, cuTENSOR
 
     @test dot(mpomps₁, mpomps₁) ≈ dot(mpo₁, mpo₁)
 end
+
+@testset "Finite CuMPOHamiltonian T ($T), V ($(spacetype(V)))" for T in (Float64, ComplexF64), V in (ℂ^2, U1Space(-1 => 1, 0 => 1, 1 => 1))
+    L = 3
+    lattice = fill(V, L)
+    O₁ = randn(T, V, V)
+    O₁ += O₁'
```
lkdvos:
```suggestion
    O₁ = project_hermitian!(randn(T, V, V))
```

test/gpu/cuda/operators.jl
```diff
@@ -83,3 +83,190 @@ using Adapt, CUDA, cuTENSOR
...
+    E = id(CuVector{T, CUDA.DeviceMemory}, domain(O₁))
+    O₂ = randn(T, V^2 ← V^2)
+    O₂ += O₂'
```
lkdvos:
```suggestion
    O₂ = project_hermitian!(randn(T, V^2 ← V^2))
```

Gotta start using our new toys :)

src/operators/ortho.jl
kshyatt: Let me first fix the failures the other changes introduced then I was going to come back and try just removing this check

lkdvos (commented):
Left some final small comments, otherwise good to go for me!

kshyatt:
> Left some final small comments, otherwise good to go for me!

Great, let me get tests fixed up and try to do a little more housekeeping and I'm happy too.

PR #415, opened 2026-04-17
Fixing changebonds, some inconsistency remaining

LHerviou:
Following discussion with @lkdvos 

Here is a small fix for some minor bugs I noticed in changebonds: for RandExpand or SvdCut, the returned environments were not updated after the update of the mps. Schematically, we had
newmps, oldenv = changebonds(oldmps, H, alg = ..., oldenv)
which made any groundstate algorithm failed.
Now the environments are updated such that
newmps, newenv = changebonds(oldmps, H, alg = ..., oldenv)
without touching oldmps or oldenv

Before accepting the update: there are some inconsistencies in the behavior of changebonds.
In particular, for OptimalExpand, env are directly modified in place, which makes oldenv not work with the old mps. Is it intended behavior for memory optimization? If so, I can reproduce a similar behavior for others. It should in any case be advertised.

I modified also slightly the converter between InfiniteMPOHamiltonian and InfiniteMPO which was only used by OptimalExpand. There is some unsafe looking environment update in it, though.

The test pass for FiniteMPS, though I did not investigate it as much. In particular, the environment update in OptimalExpand is hidden in the computation of AC2.

Finally, I added a constructor of the type FiniteMPS(Type, Space, Space) as the syntax works for infiniteMPS and not Finite, which was annoying me.

For reproduction: here is a script that fails before the PR, and pass afterwards:

```
## Infinite chain, InfiniteMPOHamiltonian
L = 2
mps = InfiniteMPS(ComplexF64, repeat([ComplexSpace(2)], L), repeat([ComplexSpace(20)], L) )
H = transverse_field_ising(ComplexF64, Trivial, InfiniteChain(L))
env = environments(mps, H)

for alg in [OptimalExpand, RandExpand, SvdCut, VUMPSSvdCut]
  env = environments(mps, H)
  newmps, newenv = changebonds(mps, H, alg(trscheme = truncrank(10)), env)    
  try
    newmps, newenv, = find_groundstate(newmps, H, VUMPS(; maxiter=1, verbosity=1, tol= 1e-6), newenv)
  catch e
    error("Fail in Infinite $alg")
    #println(e)
  end
end

## Infinite chain, InfiniteMPO
L = 2
mps = InfiniteMPS(ComplexF64, repeat([ComplexSpace(2)], L), repeat([ComplexSpace(20)], L) )
H = InfiniteMPO(transverse_field_ising(ComplexF64, Trivial, InfiniteChain(L)))
env = environments(mps, H)

for alg in [OptimalExpand, RandExpand, SvdCut, VUMPSSvdCut]
  env = environments(mps, H)
  newmps, newenv = changebonds(mps, H, alg(trscheme = truncrank(10)), env)    
  try
    newmps, newenv, = find_groundstate(newmps, H, VUMPS(; maxiter=1, verbosity=1, tol= 1e-6), newenv)
  catch e
    error("Fail in Infinite $alg")
    #println(e)
  end
end


##Finite Chain
L = 10
mps = FiniteMPS(ComplexF64, repeat([ComplexSpace(2)], L), repeat([ComplexSpace(20)], L-1) )
H = transverse_field_ising(ComplexF64, Trivial, FiniteChain(L))
env = environments(mps, H)


for alg in [OptimalExpand, RandExpand, SvdCut]
  env = environments(mps, H)
  newmps, newenv = changebonds(mps, H, alg(trscheme = truncrank(10)), env)    
  try
    newmps, newenv, = find_groundstate(newmps, H, DMRG(; maxiter=1, verbosity=1, tol= 1e-6), newenv)
  catch e
    error("Fail in Finite $alg")
    #println(e)
  end
end```

### DIFF
--- a/src/algorithms/changebonds/optimalexpand.jl
+++ b/src/algorithms/changebonds/optimalexpand.jl
@@ -24,7 +24,7 @@ function changebonds(
     ψ′, envs′ = changebonds(
         convert(MultilineMPS, ψ), convert(MultilineMPO, operator), alg, Multiline([envs])
     )
-    return convert(InfiniteMPS, ψ′), envs #This does not sound safe, it relies on the onsite modification of the environments, and the compatibility of the conversion
+    return convert(InfiniteMPS, ψ′), only(parent(envs′))
 end
 # =#
 
--- a/src/algorithms/changebonds/optimalexpand.jl
+++ b/src/algorithms/changebonds/optimalexpand.jl
@@ -5,7 +5,8 @@ An algorithm that expands the given mps as described in
 [Zauner-Stauber et al. Phys. Rev. B 97 (2018)](@cite zauner-stauber2018), by selecting the
 dominant contributions of a two-site updated MPS tensor, orthogonal to the original ψ.
 
-changedbonds! is only defined for FiniteMPS, and modify its environment.
+!!! note
+    [`changebonds!`](@ref) is only defined for `FiniteMPS`, and modifies both the state and its environment.
 
 ## Fields
 
@@ -19,8 +20,7 @@ $(TYPEDFIELDS)
     trscheme::TruncationStrategy
 end
 
-##= This is only useful for OptimalExpand
-### Simple wrapper to convert between diffrent type of InifniteMPS.
+# Simple wrapper to convert between diffrent type of InifniteMPS.
 function changebonds(
         ψ::InfiniteMPS, operator::InfiniteMPO, alg::OptimalExpand, envs = environments(ψ, operator)
     )
@@ -29,7 +29,6 @@ function changebonds(
     )
     return convert(InfiniteMPS, ψ′), only(parent(envs′))
 end
-# =#
 
 function changebonds(
         ψ::InfiniteMPS, H::InfiniteMPOHamiltonian, alg::OptimalExpand,
--- a/src/algorithms/changebonds/optimalexpand.jl
+++ b/src/algorithms/changebonds/optimalexpand.jl
@@ -5,6 +5,8 @@ An algorithm that expands the given mps as described in
 [Zauner-Stauber et al. Phys. Rev. B 97 (2018)](@cite zauner-stauber2018), by selecting the
 dominant contributions of a two-site updated MPS tensor, orthogonal to the original ψ.
 
+changedbonds! is only defined for FiniteMPS, and modify its environment.
+
 ## Fields
 
 $(TYPEDFIELDS)
@@ -18,6 +20,7 @@ $(TYPEDFIELDS)
 end
 
 ##= This is only useful for OptimalExpand
+### Simple wrapper to convert between diffrent type of InifniteMPS.
 function changebonds(
         ψ::InfiniteMPS, operator::InfiniteMPO, alg::OptimalExpand, envs = environments(ψ, operator)
     )
@@ -50,7 +53,7 @@ function changebonds(
     end
 
     newψ = _expand(ψ, AL′, AR′)
-    recalculate!(envs, newψ, H)
+    envs = environments(newψ, H)
     return newψ, envs
 end
 
@@ -75,13 +78,14 @@ function changebonds(ψ::MultilineMPS, H, alg::OptimalExpand, envs = environment
     end
 
     newψ = _expand(ψ, AL′, AR′)
-    recalculate!(envs, newψ, H)
+    envs = environments(newψ, H) #  recalculate!(envs, newψ, H)
     return newψ, envs
 end
 
 function changebonds(ψ::AbstractFiniteMPS, H, alg::OptimalExpand, envs = environments(ψ, H))
     return changebonds!(copy(ψ), H, alg, envs)
 end
+
 function changebonds!(ψ::AbstractFiniteMPS, H, alg::OptimalExpand, envs = environments(ψ, H))
     #inspired by the infinite mps algorithm, alternative is to use https://arxiv.org/pdf/1501.05504.pdf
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/algorithms/changebonds/optimalexpand.jl [hunk 1]
```diff
@@ -17,6 +17,17 @@ $(TYPEDFIELDS)
     trscheme::TruncationStrategy
 end
 
+##= This is only useful for OptimalExpand
```
lkdvos: I think this is the only comment I'm not fully following, why is this the case?

src/algorithms/changebonds/optimalexpand.jl
```diff
@@ -17,6 +17,17 @@ $(TYPEDFIELDS)
...
+function changebonds(
+        ψ::InfiniteMPS, operator::InfiniteMPO, alg::OptimalExpand, envs = environments(ψ, operator)
+    )
+    ψ′, envs′ = changebonds(
+        convert(MultilineMPS, ψ), convert(MultilineMPO, operator), alg, Multiline([envs])
+    )
+    return convert(InfiniteMPS, ψ′), envs #This does not sound safe, it relies on the onsite modification of the environments, and the compatibility of the conversion
```
lkdvos:
```suggestion
    return convert(InfiniteMPS, ψ′), only(parent(envs′))
```
I think this should solve the problem (you are fully right, we shouldn't have done that :)

lkdvos:
As a general comment, I feel like we are not super consistent about the role of `environments` throughout the entire package to be honest. In general they are somewhat treated as an in-place thing, so I think I would be okay with all functions counting as "destroying the input/taking ownership of that", which would then indeed mean that the environments are no longer compatible with the old state, unless you did an explicit copy. However, this is probably something I should better document 😁

src/algorithms/changebonds/optimalexpand.jl
LHerviou: Well, for all other functions, you have a different version for InfiniteMPO/InfiniteMPS and Multiline/InfiniteHamiltonian
I can make a global writer.

Note that I did this only because I had a conflict with one of the wrappers. I am happy to refactor everything a bit better.

LHerviou:
If you want, I can go through the different changebonds functions to systematize it.
Just let me know whether you want to:
- always modify the environments in place
- never modify the environments

At least for infiniteMPS, we can make stuff consistent.
I will also update the doc accordingly

src/algorithms/changebonds/optimalexpand.jl
LHerviou: It works indeed.

src/algorithms/changebonds/optimalexpand.jl
```diff
@@ -5,6 +5,8 @@ An algorithm that expands the given mps as described in
 [Zauner-Stauber et al. Phys. Rev. B 97 (2018)](@cite zauner-stauber2018), by selecting the
 dominant contributions of a two-site updated MPS tensor, orthogonal to the original ψ.
 
+changedbonds! is only defined for FiniteMPS, and modify its environment.
```
lkdvos:
```suggestion
!!! note
    `changebonds!` is only defined for `FiniteMPS`, and modifies its environment.
```

src/algorithms/changebonds/randexpand.jl
```diff
@@ -9,6 +9,9 @@ parallel, and therefore the expansion will never go beyond the local two-site su
 The truncation strategy dictates the number of expanded states, by generating uniformly
 distributed weights for each state in the two-site space and truncating that.
 
+The environments are not used here.
+changebonds! modify both the provided state and environments.
```
lkdvos:
```suggestion
`changebonds!` modifies both the provided state and environments.
```

src/algorithms/changebonds/optimalexpand.jl
```diff
@@ -39,7 +53,7 @@ function changebonds(
     end
 
     newψ = _expand(ψ, AL′, AR′)
-    recalculate!(envs, newψ, H)
+    envs = environments(newψ, H)
```
lkdvos, line 55: I'm not entirely following this change. Is it not possible to overwrite the old environments?

src/algorithms/changebonds/optimalexpand.jl
```diff
@@ -64,13 +78,14 @@ function changebonds(ψ::MultilineMPS, H, alg::OptimalExpand, envs = environment
     end
 
     newψ = _expand(ψ, AL′, AR′)
-    recalculate!(envs, newψ, H)
+    envs = environments(newψ, H) #  recalculate!(envs, newψ, H)
```
lkdvos, line 80: Same comment here

src/algorithms/changebonds/optimalexpand.jl
lkdvos: Ah, I think I see now what you intended: is it that the non-bang (!) functions shouldn't modify their inputs?
I can see that being a good choice, I think it is mostly that the `envs` are always kind of treated as a special thing that is modified everywhere.
I'm happy either way, I think you have a better feeling of what is convenient/efficient.

src/algorithms/changebonds/vumpssvd.jl
```diff
@@ -3,6 +3,8 @@ $(TYPEDEF)
 
 An algorithm that uses a two-site update step to change the bond dimension of a state.
 
+changedbonds! is not defined.
```
lkdvos:
```suggestion
!!! note
    `changebonds!` is not defined.
```

lkdvos (commented):
Left some minor comments, but I definitely like that this is more principled with the inplace vs non-inplace functions.

lkdvos (approved):

PR #416, opened 2026-04-20
Number of fixes to support latest versions of TensorKit

borisdevos:
`TensorKit.FusionBlockStructure` got reworked in QuantumKitHub/TensorKit.jl#387, so this PR deals with those changes.


lkdvos:
This has to be released after another version of TensorKit is tagged, and subsequently also should increase the minimal TensorKit compat, thanks for already opening the PR!

kshyatt:
(I checked with @borisdevos before doing this) I have added a `sources` pointing to TensorKit `main` so we can figure out what will need fixing for the next release

kshyatt:
Looks like `multifusion` and `algorithms` need some TLC... our old friend `IsingBimodule` has struck again!!

kshyatt:
`multifusion` failure is actually due to `BlockTensorKit`, I'll open the issue there

lkdvos:
I might have some time today to figure the remainder of the issues out here, is it okay if I use this branch for that?

kshyatt:
No objection from me!

borisdevos:
Same here!

lkdvos:
Needs https://github.com/QuantumKitHub/TensorKit.jl/pull/418

lkdvos:
This will have to wait a bit more I fear, since there are still issues in the TensorKit changes that have to go in first.

lkdvos:
I am hoping the tests here will pass now, after which I am confident that https://github.com/QuantumKitHub/TensorKit.jl/pull/424 contains enough fixes to be released today, and this can then be released on top of that.

PR #418, opened 2026-04-23
Update MatrixAlgebraKit algorithm specification

leburgel:
Updates the decomposition algorithm specification for the changes of MatrixAlgebraKit.jl v0.6.5.

For the SVD algorithm I set `DefaultAlgorithm` as the default, since that should just pick the best option regardless of the storage type. For the QR algorithm I couldn't do that because of QuantumKitHub/TensorKit.jl#405, so there I explicitly set it to `HouseHolder(; positive = true)`.


leburgel:
CUDA tests are failing, but I see the same failure on `main` so it doesn't seem to be caused by the changes here.

The issue seems to be that calling
```julia
left_orth!(t::TensorKit.TensorMap{ComplexF64, TensorKit.ComplexSpace, 2, 4, CUDA.CuArray{ComplexF64, 1, CUDA.DeviceMemory}}; trunc::MatrixAlgebraKit.TruncationByValue{Float64, Int64, typeof(abs)})
```
with only a `trunc` specified by no algorithm, is processed further down to
```julia
left_orth!(A::TensorKit.TensorMap{ComplexF64, TensorKit.ComplexSpace, 2, 4, CUDA.CuArray{ComplexF64, 1, CUDA.DeviceMemory}}, VC::Nothing, alg::MatrixAlgebraKit.LeftOrthAlgorithm{:svd, MatrixAlgebraKit.TruncatedAlgorithm{MatrixAlgebraKit.SafeDivideAndConquer{@NamedTuple{}}, MatrixAlgebraKit.TruncationByValue{Float64, Int64, typeof(abs)}}})
```
with an explicit `SafeDivideAndConquer` SVD algorithm. This then leads to
```
ArgumentError: MatrixAlgebraKit.CUSOLVER() does not provide gesdvd!
```
This seems like a problem with the algorithm selection not properly taking into account the storagetype in MatrixAlgebraKit.jl/TensorKit.jl? I can open an issue, but not really sure which one of the two is giving the problem.

lkdvos:
Thanks for this, I will merge this as indeed the CUDA failures are not caused by something here and don't have to be solved here either.

PR #422, opened 2026-05-03
Time-evolution MPO refactoring

lkdvos:
In an attempt to reduce TTFX I refactored some of the code for the time-evolution MPO constructions, mostly just introducing a couple more function barriers.
Especially for the `TaylorCluster` implementation, the `Val(N)` dispatch should really help inference out quite a lot.
Additionally, this does actually affect runtime beyond TTFX, so overall this should be a good improvement. 🤞


VictorVanthilt:
I'm curious, did this significantly improve TTFX?

lkdvos:
Less than I was hoping :( it did actually speed up the code for small hamiltonians by quite a bit, since all of the bookkeeping stuff and indexing stuff was using non-inferred `CartesianIndex{N} where {N}`, so at least there is that

Issue #249, opened 2025-02-06
Add support for Makie.jl plotting capabilities


lkdvos:
As a follow-up to https://github.com/QuantumKitHub/MPSKit.jl/issues/246, it would be nice to have support for plotting through the Makie.jl ecosystem.

PR #428, opened 2026-05-20
Extend plot support for `entanglementplot` and `transferplot` to `Makie.jl`

borisdevos:
Deals with #249. This works locally, below I attach the results for both Plots.jl and CairoMakie.jl*. I didn't really test any other Makie backend, but I think CairoMakie is the one you want anyways in this context.

As you can see in the attachment, I didn't really bother to make the `Plots.jl` plots any prettier than they were, and they remain equally blurry. Anyone willing to prettify these, feel free to do so. I put slightly more effort in the Makie plots, but even there plenty of improvements can be added.

I added todo's here and there, mostly stylistic choices.

*There are two plots per function, just to show that keyword arguments are being taken correctly:
```julia
# gs is the ground-state of the Ising Hamiltonian at g = 0.1
entanglementplot(gs)
entanglementplot(gs, expand_symmetry = true, sortby = maximum, sector_formatter = c -> "Sector $c")

transferplot(gs)
transferplot(gs; sectors = [Z2Irrep(0), Z2Irrep(1)], transferkwargs = (; num_vals = 1))
```
[someplots.pdf](https://github.com/user-attachments/files/28046863/someplots.pdf)


test/Project.toml
```diff
@@ -4,6 +4,7 @@ name = "MPSKitTests"
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 BlockTensorKit = "5f87ffc2-9cf1-4a46-8172-465d160bd8cd"
+CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
```
lkdvos: This should probably be a weakdep?

Project.toml
```diff
@@ -10,24 +10,28 @@ Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 HalfIntegers = "f0d1745a-41c9-11e9-1dd9-e5d34d218721"
 KrylovKit = "0b1a1467-8014-51b9-945f-bf0ae24f4b77"
+LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
```
lkdvos: Does this make sense to have as a default dependency? We could in principle also have a default xlabel that uses Unicode, and require people that wish Latex strings to specify the xlabel themselves (which I would assume people typically want anyways)

lkdvos (commented):
This definitely looks great!
I have a couple questions, mostly since it has been a while since I looked at Makie:

Do you know if the other attributes can still be set/changed? For example, what happens if I want to override the default title/ticks/...?

I seem to recall also that there was some system with observables etc, where various parts where dynamically recalculated. Do you know if this is still relevant?

Project.toml
borisdevos: At least for Makie I could keep the default latexstring expressions and put it as a weakdep, since it's explicitly in the deps of Makie. I put it here in deps to get it working for Plots, but we can indeed default to whatever and let the user overwrite themselves (if possible for Plots, I think I've got it working for Makie right now)

borisdevos:
entspec_makie_plotkwargs
transpec_makie_plotkwargs

I just tried some random keyword arguments for the Makie plots, and they seem to work. I'll test tomorrow if I can do the same for Plots.
(Also I realised I was running Z2-symmetric MPS in the Z2 SSB phase 🫠)

borisdevos:
I don't understand the current aqua test failing. It seems to be complaining about missing compat bounds for every package under extras of the main project.toml (but are specified in the test project.toml), but this previously wasn't failing.

Project.toml
```diff
@@ -40,16 +44,36 @@ Compat = "3.47, 4.10"
...
+ParallelTestRunner = "d3525ed8-44d0-4b2c-a655-542cee43accc"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+TensorKitTensors = "41b62e7d-e9d1-4e23-942c-79a97adf954b"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+TestExtras = "5ed8adda-3752-4e41-b88a-e8b09835ee3a"
+cuTENSOR = "011b41b2-24ef-40a8-b3eb-fa098493e9e1"
+
+[targets]
+test = ["Aqua", "Adapt", "CUDA", "cuTENSOR", "Pkg", "Test", "TestExtras", "Plots", "Makie", "CairoMakie", "Combinatorics", "ParallelTestRunner", "TensorKitTensors"]
```
lkdvos: The new scheme for specifying test dependencies does not work with the `[extras]` and `[targets]` sections anymore, so you should just be updating the `test/Project.toml` directly, and make sure the correct compat entries are filled in there :)

borisdevos:
Okay, I finally managed to look into this again. I'm happy with how things are in terms of being able to use these backends with some customisation. What remains is some stylistic decisions. In particular, I don't know how useful it is to plot transfer matrix spectra the way we are now. These phases don't matter in many cases, so if we allow some toggle to not plot them, there's then a way to plot the different sectors next to each other, similar to the current entanglement plots. 

Some other style choices are in todos, though I'm sure there are many more.

test/runtests.jl
```diff
@@ -41,4 +41,10 @@ const init_code = quote
     const fast_tests = $fast
 end
 
-ParallelTestRunner.runtests(MPSKit, args; testsuite, init_worker_code, init_code)
+# custom workers for different plot backends
```
lkdvos, line 44: why do we need custom workers for the different plot backends? Is this because both are adding methods to the same function and therefore loading both plotting packages at the same time clashes?

Project.toml
```diff
@@ -30,9 +26,17 @@ VectorInterface = "409d34a3-91d5-4945-b6ec-7529ddf182d8"
...
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
+LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 
 [extensions]
 MPSKitAdaptExt = "Adapt"
+MPSKitMakieExt = ["Makie", "LaTeXStrings"]
+MPSKitPlotsExt = "RecipesBase"
+
+[workspace]
+projects = ["test", "docs"]
```
lkdvos, line 39:
```suggestion
projects = ["test", "docs", "examples"]
```

Why remove the "examples"?

ext/MPSKitMakieExt.jl
```diff
@@ -0,0 +1,251 @@
...
+end
+
+# overwrite user-provided axis attributes
+function apply_plotkwargs!(ax, plotkwargs)
+    ax isa Makie.AbstractAxis || return ax
+    for (k, v) in pairs(plotkwargs)
+        setproperty!(ax, k, v)
+    end
+    return ax
+end
```
lkdvos, line 30: This looks a little backwards, I feel like it should be possible to make the attributes respect what was provided, and somehow it feels like we are doing too much work with providing ticks etc, for which there should just be a default that is sensible and otherwise user-overridable?

ext/MPSKitMakieExt.jl
```diff
@@ -0,0 +1,251 @@
...
+        site = 0,
+        expand_symmetry = false,
+        sortby = maximum,
+        sector_margin = 1 // 10,
+        sector_formatter = string,
+    )
+end
+
+function Makie.plot!(ep::EntanglementPlot)
+    #TODO: still want this style where sectors are separated?
```
lkdvos, line 43: I do think so, there are definitely cases where this is useful

ext/MPSKitMakieExt.jl
```diff
@@ -0,0 +1,251 @@
...
+        sectors = sectors[order]
+    end
+
+    ax = Makie.current_axis()
+
+    # Axis styling
+    ax.title = L"\text{Entanglement Spectrum}"
+    ax.titlesize = 24
+
+    ax.xlabel = latexstring("\$\\chi\$ = $(dim(MPSKit._firstspace(mps.C[site])))") # still want this?
```
lkdvos, line 77: happy to remove this

ext/MPSKitMakieExt.jl
```diff
@@ -0,0 +1,251 @@
...
+    ax.xlabelsize = 24
+    ax.xticks = (1:length(sectors), ep.sector_formatter[].(sectors))
+    ax.xticklabelsize = 16
+    ax.xticklabelrotation = 45
+    ax.xticklabelalign = (:right, :top)
+    xlims!(ax, 1, length(sectors) + 1)
+
+    ax.ylabel = L"\log(\lambda)"
+    ax.ylabelsize = 24
+    smallest = minimum(Iterators.filter(>(0), Iterators.flatten(spectrum)); init = 1.0) # safety net
```
lkdvos, line 87: Does this mean that we are effectively hiding pathological cases? Do you think there is a way of still showing them so we can diagnose the issue?

test/runtests.jl
borisdevos: Yup

Project.toml
borisdevos: Bad merging on my end, sorry!

ext/MPSKitMakieExt.jl
borisdevos: I can filter on `==0`, give a warning but still plot the way it is currently?

ext/MPSKitMakieExt.jl
borisdevos: So this part of the code exists to deal with the in-place versions of the plot functions, but you want these things to be attributes of the default? 

The workaround here was because some of the default attributes I set are of the axis, and not of the scatter plot itself. I [looked](https://docs.makie.org/dev/explanations/recipes#Full-recipes-with-the-@recipe-macro) at the whole `ComputeGraph` machinery you once mentioned to me, and I think that's exactly you want. I'll try this out

PR #432, opened 2026-05-29
Bump VectorInterface compat to include v0.6

kshyatt:
Allow use of VectorInterface 0.6


VictorVanthilt:
But v0.6 has not been tagged yet, and therefore not tested here?

lkdvos:
It has, unfortunately TensorKit is blocking this and I was too quick 😢 
Good call-out, this definitely was my error, I'll leave it as-is for now since I dont think there are any consequences, but yeah...

PR #434, opened 2026-06-09
Bump TensorKit v0.17 compatibility

lkdvos:
Stole some of #406 (but not all of it).

This is the minimal changes for working with TensorKit v0.17.


Project.toml
kshyatt: Should we also bump this? In theory compat will force 0.3.14 so maybe it doesn't matter...

kshyatt (approved):

PR #438, opened 2026-06-19
Controlled Bond Expansion: randomized SVD approach

lkdvos:
This is a WIP implementation of a CBE-like expansion scheme based on randomized SVD.
The long story short is that instead of a full twosite (`OptimalExpand`) -based approach, we can sketch out this update to select the expansion directions.
Mainly I'm still struggling slightly with committing to an interface, and checking what goes where.
I'm somewhat convinced that the `changebond!` (as opposed to `changebonds!`) abstraction for doing a single expansion on a bond is probably good, but still struggling a bit to see if I want to default this into `DMRG` (and `TDVP`), or get a dedicated `CBE_DMRG` algorithm for it.
Interestingly, I think I've set it up in such a way that the various expansion schemes can be used interchangibly, so it is quite modular.

The main source of complexity/messiness is mostly that the `FiniteMPS` structs and environments are very untransparant about what they do, so trying to reuse functionality involves a lot of assumptions on which tensors exactly are where and which ones are used.
It would be nice to revisit that at some point, but I also just don't have the time to actually go ahead and do this...

Still to do:
- [x] Implement for TDVP
- [x] Clean up DMRG implementation
- [x] Verify interaction with symmetries and block selection / oversampling
- [x] Check this works for density matrices
- [x] correctly rebase with timeroutputs


VictorVanthilt (approved):

src/algorithms/changebonds/changebonds.jl
```diff
@@ -12,27 +12,41 @@ See also: [`SvdCut`](@ref), [`RandExpand`](@ref), [`VUMPSSvdCut`](@ref), [`Optim
 function changebonds end
 function changebonds! end
 
+@doc """
+    changebond(site, dir, ψ, [H], alg, [envs]) -> ψ
+    changebond!(site, dir, ψ, [H], alg, [envs]) -> ψ
+
+Expand a single bond of `ψ` in place by adding directions orthogonal to the current state, keeping the state in mixed-canonical form around the enriched bond.
```
leburgel, line 19:
```suggestion
Expand a single bond of `ψ` by adding directions orthogonal to the current state, keeping the state in mixed-canonical form around the expanded bond.
```

src/algorithms/changebonds/changebonds.jl
```diff
@@ -12,27 +12,41 @@ See also: [`SvdCut`](@ref), [`RandExpand`](@ref), [`VUMPSSvdCut`](@ref), [`Optim
...
+The sweep direction `dir` is a `Val(:right)` or `Val(:left)` used for dispatch.
+For `Val(:right)` the bond `(site, site + 1)` is enriched on the right tensor (`ψ.AR[site + 1]`) with zero weight added at `ψ.AC[site]`, so that a subsequent single-site optimization of `site` sees the new directions;
```
leburgel, line 21:
```suggestion
For `Val(:right)` the bond `(site, site + 1)` is expanded on the right tensor (`ψ.AR[site + 1]`) with zero weight added at `ψ.AC[site]`, so that a subsequent single-site optimization of `site` sees the new directions;
```

src/algorithms/changebonds/optimalexpand.jl
```diff
@@ -5,6 +5,9 @@ An algorithm that expands the given mps as described in
 [Zauner-Stauber et al. Phys. Rev. B 97 (2018)](@cite zauner-stauber2018), by selecting the
 dominant contributions of a two-site updated MPS tensor, orthogonal to the original ψ.
 
+The expansion does not alter the state: the added directions are connected through a zero block,
+so that the expanded state is identical to the original one (as required for e.g. TDVP).
```
leburgel, line 9: Could this be confusing, in the sense that people think "does not alter" means it acts out-of-place?

src/algorithms/changebonds/optimalexpand.jl
```diff
@@ -81,37 +84,71 @@ function changebonds(ψ::MultilineMPS, H, alg::OptimalExpand, envs = environment
...
+    right = ψ.AR[site + 1]
+    NL = left_null(left)
+    NR = right_null!(_transpose_tail(right; copy = true))
+
+    # two-site update from the projected effective Hamiltonian
+    AC2 = AC2_projection(bond, ψ, H, ψ, envs)
+
+    # select the dominant directions in the complement of the current state
+    g2 = adjoint(NL) * AC2 * adjoint(NR)
+    _, _, Vᴴ = svd_trunc!(normalize!(g2); trunc = alg.trscheme, alg = alg.alg_svd)
```
leburgel, line 102: Just as a reminder, should we add a rename `trscheme -> trunc` to the set of breaking changes for the next breaking MPSKit release?

src/algorithms/changebonds/optimalexpand.jl
```diff
@@ -81,37 +84,71 @@ function changebonds(ψ::MultilineMPS, H, alg::OptimalExpand, envs = environment
...
+    nal_space = codomain(left) ← (only(domain(left)) ⊕ space(Vᴴ, 1))
+    nal, nc = left_gauge(absorb!(zerovector!(similar(left, nal_space)), left))
+    nar = _transpose_front(catcodomain(_transpose_tail(right), ar_re))
+
+    normalize && normalize!(nc)
+    ψ.AC[site] = (nal, nc)
+    ψ.AC[site + 1] = (nc, nar)
+    return ψ
+end
+function changebond!(site::Int, ::Val{:left}, ψ::AbstractFiniteMPS, H, alg::OptimalExpand, envs; normalize::Bool = true)
```
leburgel, line 117: Just a question, does it make sense to try and write `changebond!(site::Int, ::Val{:left}, ...)` in terms of the main workflow of `changebond!(site::Int, ::Val{:right}, ...)` (which would then have to be split off) by transposing/permuting the relevant tensors, or is this a bit tricky?

src/algorithms/changebonds/randexpand.jl
```diff
@@ -68,30 +68,69 @@ end
...
+    nal_space = codomain(left) ← (only(domain(left)) ⊕ space(Vᴴ, 1))
+    nal, nc = left_gauge(absorb!(zerovector!(similar(left, nal_space)), left))
+    nar = _transpose_front(catcodomain(_transpose_tail(right), ar_re))
+
+    normalize && normalize!(nc)
+    ψ.AC[site] = (nal, nc)
+    ψ.AC[site + 1] = (nc, nar)
+    return ψ
+end
+function changebond!(site::Int, ::Val{:left}, ψ::AbstractFiniteMPS, H, alg::RandExpand, envs; normalize::Bool = true)
```
leburgel, line 106: It looks like this one is just never used?

src/algorithms/changebonds/sketchedexpand.jl
```diff
@@ -0,0 +1,141 @@
...
+    nal_space = codomain(left) ← (only(domain(left)) ⊕ space(ar_re, 1))
+    nal, nc = left_gauge(absorb!(zerovector!(similar(left, nal_space)), left))
+    nar = _transpose_front(catcodomain(_transpose_tail(right), ar_re))
+
+    normalize && normalize!(nc)
+    ψ.AC[site] = (nal, nc)
+    ψ.AC[site + 1] = (nc, nar)
+    return ψ
+end
+function changebond!(site::Int, ::Val{:left}, ψ::AbstractFiniteMPS, H, alg::SketchedExpand, envs; normalize::Bool = true)
```
leburgel, line 92: Same question about reusing the core workflow of the right expansion on transposed/permuted tensors.

src/algorithms/timestep/tdvp.jl
```diff
@@ -3,6 +3,19 @@ $(TYPEDEF)
 
 Single site MPS time-evolution algorithm based on the Time-Dependent Variational Principle.
 
+For finite MPS, setting `alg_expand` to a bond-expansion algorithm (e.g. [`OptimalExpand`](@ref),
+[`SketchedExpand`](@ref)) enriches the bond with directions orthogonal to the current state
```
leburgel, line 7: Still struggling with "enriches". Any opposition to just using "expands"?

leburgel (commented):
Too slow to review before this was merged, but I'm giving it anyway just in case there's some things we want to change after the fact.

Everything looks great, but I was wondering: why not do this for infinite MPS right away as well? Is there some obstruction to this?

lkdvos:
I'll open a follow-up PR to adress your comments. 

- I think the one about transposed/permuted tensors is not really viable since there is some fundamental asymmetry in how our MPS tensors are structured, and I really do want to avoid every single unnecessary transpose, since this unfortunately all shows up in non-abelian symmetric cases.
- Considering the language, happy to change.
- For the `trscheme` -> `trunc`, I agree, but would want to evaluate slightly more in-depth: it might actually be reasonable to just have a single field for the decomposition algorithm + truncation scheme together, since that is anyways how they are used and allows algorithms that can't easily be split like that. I want to keep the option of having these as keyword arguments for sure, for convenience and clarity, but then internally store them differently.

lkdvos:
Before I forget: the reason why I can't do this for the infinite case directly is that I'm not sure it really works out so easily: it gives the same kind of issues as we have on the boundaries for IDMRG2, which I'm not sure I want to introduce in general (although we should probably investigate if it is possible). The issue for VUMPS is that adding zero-directions creates a state that is hard to gauge (I think), so the canonicalization step might fail.
TLDR - needs a bit more investigation, and I definitely already wanted to get this in.

PR #444, opened 2026-06-25
Death to `BraidingTensor`

lkdvos:
This is a (WIP) implementation that redesigns the `JordanMPOTensor` interface, instead of keeping the `BraidingTensors` explicitly in the block tensor, which causes a lot of promotion issues, to keep them separate in a "scalars" field.

Right now, we aren't really using the scalars for anything other than detecting where the identity tensors are on the diagonal, so we can properly renormalize variances etc. I'm not sure if this ever really factors into the computations, and in any case with the altered contraction orders for the derivatives this never matters anymore, since we always contract the braiding tensors on one leg only (with the environments instead of with the environment + mps).

@kshyatt I'm not sure if there is a way to check if/how many issues this solves?

### DIFF
--- a/src/operators/jordanmpotensor.jl
+++ b/src/operators/jordanmpotensor.jl
@@ -134,8 +134,18 @@ function JordanMPOTensor(
     )
 end
 
+# whether `t` is (approximately) an identity braiding tensor, so it can be folded back
+# into a scalar identity
+function _isbraiding(t::AbstractTensorMap)
+    t isa BraidingTensor && return true
+    τ = BraidingTensor{scalartype(t), spacetype(t), storagetype(t)}(space(t))
+    return isapprox(t, τ)
+end
+
 function JordanMPOTensor(W::SparseBlockTensorMap{TT, E, S, 2, 2}) where {TT, E, S}
-    @assert W[1, 1, 1, 1] isa BraidingTensor && W[end, 1, 1, end] isa BraidingTensor
+    # the diagonal corners must be (close to) identity braidings; the constructor below
+    # folds them back into scalar identities
+    @assert _isbraiding(W[1, 1, 1, 1]) && _isbraiding(W[end, 1, 1, end])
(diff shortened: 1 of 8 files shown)


src/operators/jordanmpotensor.jl
```diff
@@ -107,28 +134,29 @@ function JordanMPOTensor(
     )
 end
 
+function _isbraiding(t::AbstractTensorMap)
+    t isa BraidingTensor && return true
```
kshyatt: We don't want to do this with dispatch?

src/operators/jordanmpotensor.jl
```diff
@@ -140,31 +168,86 @@ end
 
 # Properties
 # ----------
-TensorKit.space(W::JordanMPOTensor) = W.V
-Base.eltype(::Type{JordanMPOTensor{E, S, TA, TB, TC, TD}}) where {E, S, TA, TB, TC, TD} = TA
+TensorKit.space(W::JordanMPOTensor) = space(W.tensors)
+function Base.eltype(::Type{<:JordanMPOTensor{T, S, A}}) where {T, S, A}
```
kshyatt: Style nit: couldn't this just be one line?

src/operators/jordanmpotensor.jl
```diff
@@ -107,28 +134,29 @@ function JordanMPOTensor(
     )
 end
 
+function _isbraiding(t::AbstractTensorMap)
+    t isa BraidingTensor && return true
+    τ = BraidingTensor{scalartype(t), spacetype(t), storagetype(t)}(space(t))
```
kshyatt: The recurrence of this kind of `BraidingTensor` constructor makes we wonder if we need a `similar_braidingtensor` or somesuch

src/operators/jordanmpotensor.jl
lkdvos: ~~It was less lines to just have the branch here~~, I don't have any strong opinions though 🤷

PR #449, opened 2026-07-03
docs: Diátaxis restructure — tutorials, how-to guides, concepts, library reference

lkdvos:
Restructures the documentation along Diátaxis lines: a sequential tutorial track, task-recipe how-to guides, prose concept pages, and a topic-split library reference, replacing the `man/` section that mixed all four on the same page. Also rewrites the landing page.

**Stacked on #500** (base is `docs/examples-regroup`), so the diff above shows only this PR's own changes. The landing page and the symmetry tutorial link to the regrouped example paths and to the `examples_index` anchor that #500 introduces, so #500 merges first; GitHub retargets this to `main` automatically when it does.

Everything mechanical has been split out of this PR so what remains is the prose that actually needs reading: #497 (VitePress config), #499 (contributing guide), #500 (examples regroup), #501 (citing/migration), #502 (two new examples, stacked on this).

## Adapted to API changes since the branch was cut

- `trscheme` → `trunc` (#482) throughout.
- The pages claimed imaginary-time evolution renormalizes every step. #482 split that into an explicit `normalize` keyword defaulting to `false`, so the claim was false; corrected wherever `imaginary_evolution` is introduced.
- Added coverage of `BUG` (#468) and `Zipup` (#470), which merged after the pages were written.

## Review notes

- Locally verified: the whole Documenter half passes with no errors — every `@example` runs against current `main`, all `@ref`/`@extref`/`@cite` resolve, `checkdocs` clean. The VitePress stage cannot run on my machine (rollup needs a newer glibc), so CI covers the HTML render.
- The intended `make_time_mpo` + `Zipup` recipe had to be dropped: `Zipup` rejects propagator MPOs with a `SpaceMismatch` (#496).
- Concept and how-to pages are cross-linked densely (141 cross-group references), which is why they are one PR rather than four — splitting by page type would mean rewriting tens of links and restoring them later.
- The previous 82-commit history is preserved at `archive/content-wave1-pre-rebase` if anything needs recovering.


lkdvos:
@leburgel, any chance you can confirm this would resolve the two issues you brought up?

leburgel:
> @leburgel, any chance you can confirm this would resolve the two issues you brought up?

The search issue seems to be resolved, which was definitely the most important one.

The docstring header issue improved a bit, but it's not quite solved. The custom `jldocstring-heading` specification is used correctly and shows the corresponding thin bottom border line. But the font weight is still not right for a `<strong>...</strong>` block, and the text still renders as plain instead of bold. When I inspect, I see the font weight of the strong block is 700, which should really be bold. So I have no idea what's going wrong, but I'm not sure if it's worth figuring out. The bottom border is enough to indicate the header, so I guess it's good enough.

leburgel:
Unrelated: it seems the warning and note blocks have lost quite a bit of their style compared to the way they look in standard documenter. This is quite unfortunate, the blocks here look quite a bit worse than they used to.

PR #450, opened 2026-07-05
bench: round-1 investigation harness (Float64/DMRG2 parity, TDVP + thread-scaling suites, Rusty scripts, diagnostics)

lkdvos:
Round 1 of the MPSKit-vs-ITensorMPS performance investigation, per the methodology decisions settled on 2026-07-05.
Targets the \`bench/itensor-comparison\` branch, not \`main\`.

## What changed

**Suites 1–2 (finite DMRG) reworked for exact parity**
- Float64 on both sides (MPSKit's ComplexF64 constructor default was a 2–4x BLAS handicap on a real-symmetric Hamiltonian).
- Both suites now compare MPSKit \`DMRG2(trscheme = truncrank(χ))\` against ITensor's default two-site \`dmrg\` — ITensorMPS exposes no single-site DMRG, so the earlier single-site suite-1 variant is gone.
- Sweep budget 30 → 10 (pilot at χ = 256 running; follow-up commit if it says otherwise).
- Parity table rewritten; rows 2/4/5 now matched/exact.

**New suite 5 — TDVP throughput**
- Global quench from the Néel state, spin-1 Heisenberg N = 100; two-site growth phase then single-site measure phase, mirrored protocols; per-step (⟨Sz⟩, χ, walltime) trajectories; wall-seconds per unit physical time as the metric.
- Krylov exponentiation tolerance matched at 1e-10 on both sides (ITensor would otherwise inherit KrylovKit's tighter 1e-12 and do strictly more work).
- Smoke gate: mid-chain ⟨Sz⟩ agrees across libraries to **~1e-13** at both χ — same trajectory, same protocol.

**New suite 7 — thread scaling**
- Suite-1 workload at one (N, χ) point, one process per (JULIA_NUM_THREADS × BLAS-threads) grid point, grid driven by the sbatch scripts.

**Rusty submission scripts (\`benchmark/slurm/\`)**
- Six sbatch scripts (single exclusive node, constraint placeholder to pin one CPU generation) + README with submission order and the sanity-gate checklist. Manual submission only.

**Diagnostics**
- Per-sweep GC seconds + allocated bytes recorded on both sides.
- \`profile_sweep.jl\`: flat/tree profile of one DMRG2 sweep.
- \`plot_results.jl\`: cross-library overlays, sanity-gate + time-to-accuracy tables, TDVP observable-gate/throughput panels, thread-scaling speedups.
- \`ANALYSIS.md\`: question-driven skeleton, filled only from real cluster results.

## Verification

- All four suites smoke-run clean on both libraries (workstation, JULIA_NUM_THREADS=1, BLAS=1).
- DMRG smoke energies agree to leading digits; U(1) χ=8 pair agrees to 6e-5; suite-1 χ=8 ITensor point sits in its known no-noise fixed-χ local minimum (documented).
- TDVP observable trajectories agree to ~1e-13.
- \`plot_results.jl\` runs end-to-end on the smoke JSONs and emits all comparison artifacts.

## Not in this PR

Full-scale numbers (they come from Rusty via \`benchmark/slurm/\`, submitted manually), any \`docs/src/benchmarks.md\` content, SU(2)/TeNPy/cylinder suites.


Jutho:
This looks interesting. You had a discussion about this today? With Claude then? What is Rusty? Where can I see the results 😄 ?

lkdvos:
Accidentally opened a PR for this, still trying to run the benchmark on our cluster (rusty)

PR #452, opened 2026-07-06
refactor: `transfer_spectrum` and `correlation_length` default to sector-centered implementation

lkdvos:
This PR refactors the `transfer_spectrum` function to be more in line with the `entanglement_spectrum` function, and return a `SectorVector` for the different sectors that may appear.
This also gives us the opportunity to actually compute the correlation length in the dominant sector, rather than defaulting to the trivial sector instead, which is something that had surfaced before: See also https://github.com/QuantumKitHub/PEPSKit.jl/issues/216

Something else that had been bothering me for a long time is that the krylov dimension is selected once up front, and then never changed, so it really required quite a bit of manual scripting to actually get a reasonable set of transfer values for all different sectors.
Here i've introduced `oversampling` and `oversampling_factor` as ways to set the krylov dimension from the requested values instead, which might be a bit more ergonomic?

Finally, this is in preparation of improving the plotting functionality, see also #428, which I'd like to give similar interfaces so it might be useful to think about this a bit.


docs/src/examples/classic2d/1.hard-hexagon/index.md
```diff
@@ -53,12 +53,12 @@ V = virtual_space(D)
 ) # use non-hermitian eigensolver
 F = real(expectation_value(ψ, mpo))
 S = real(first(entropy(ψ)))
-ξ = correlation_length(ψ)
+ξ = correlation_length(ψ; sector = leftunit(ψ))
```
borisdevos, line 57: Even though I was the one to introduce the multifusion language, I feel like we should try to avoid it as much as possible for the average MPSKit user. Specifically here, I think some `unit(::InfiniteMPS)` function would be helpful and less confusing. Of course internally the defaults still need to use `left/rightunit`s, but none of the examples are multifusion.

src/algorithms/transfer_spectrum.jl
```diff
@@ -0,0 +1,179 @@
...
+"""
+function transfer_spectrum(
+        above::InfiniteMPS, below::InfiniteMPS = above, alg = nothing;
+        num_vals = 20, kwargs...
+    )
+    transferspace = fuse(left_virtualspace(above, 1) * left_virtualspace(below, 1)')
+    howmany = _transfer_howmany(num_vals, transferspace)
+
+    T = complex(scalartype(above))
+    S = spacetype(above)
```
borisdevos:
```suggestion
    S = TensorKit.check_spacetype(above, below)
```
Should be put before defining `transferspace` then. The potential space mismatch is a more informative error message than the promote one

src/algorithms/transfer_spectrum.jl
```diff
@@ -0,0 +1,179 @@
...
+- `above::InfiniteMPS`: the state for the "above" leg of the mixed transfer matrix.
+- `below::InfiniteMPS=above`: the state for the "below" leg; defaults to the pure transfer matrix of `above`.
+- `alg`: the eigensolver algorithm specification, resolved per sector via [`MatrixAlgebraKit.select_algorithm`](@extref MatrixAlgebraKit.select_algorithm).
+  This can be a KrylovKit algorithm instance (used verbatim for every sector), a `MatrixAlgebraKit.DefaultAlgorithm` or `NamedTuple` bundling keyword arguments,
+  or `nothing` (the default) to construct the eigensolver from the keyword arguments below.
+
+# Keyword arguments
+
+- `num_vals=20`: the number of eigenvalues to compute. This can either be a single `Int`, which is used for every sector of the transfer space,
+  or an `AbstractDict`/iterable of `sector => count` pairs to restrict the computation to specific sectors and request a different number of values per sector.
```
borisdevos, line 19: Just messing around with different iterables of pairs, I feel like the only one that behaves as expected is a dictionary. Calling `pairs` on a vector or tuple behaves different to what I expected at least, and just gives the dictionary back in that case.

borisdevos (approved):
Minor comments, but looks good to me!

src/algorithms/transfer_spectrum.jl
lkdvos: Woops, that was my bad - should just iterate the thing that was handed to me, not actually iterate the pairs specifically

docs/src/examples/quantum1d/5.haldane-spt/figure-1.png
leburgel, line 1: This one seems to have changed quite drastically, was this expected?

examples/classic2d/1.hard-hexagon/main.jl
```diff
@@ -49,16 +50,19 @@ V = virtual_space(D)
...
 ## The scaling hypothesis
 
-The dominant eigenvector is of course only an approximation. The finite bond dimension enforces a finite correlation length, which effectively introduces a length scale in the system. This can be exploited to formulate a scaling hypothesis [pollmann2009](@cite), which in turn allows to extract the central charge.
+The dominant eigenvector is of course only an approximation.
+The finite bond dimension enforces a finite correlation length, which effectively introduces a length scale in the system.
+This can be exploited to formulate a scaling hypothesis [pollmann2009](@cite), which in turn allows to extract the central charge.
 
-First we need to know the entropy and correlation length at a bunch of different bond dimensions. Our approach will be to re-use the previous approximated dominant eigenvector, and then expanding its bond dimension and re-running VUMPS.
-According to the scaling hypothesis we should have ``S \propto \frac{c}{6} log(ξ)``. Therefore we should find ``c`` using
+First we need to know the entropy and correlation length at a bunch of different bond dimensions.
```
leburgel:
```suggestion
First we need to know the entropy and correlation length at several different bond dimensions.
```

examples/classic2d/1.hard-hexagon/main.jl
```diff
@@ -20,7 +20,8 @@ md"""
 The [hard hexagon model](https://en.wikipedia.org/wiki/Hard_hexagon_model) is a 2-dimensional lattice model of a gas, where particles are allowed to be on the vertices of a triangular lattice, but no two particles may be adjacent.
 This can be encoded in a transfer matrix with a local MPO tensor using anyonic symmetries, and the resulting MPO has been implemented in MPSKitModels.
 
-In order to use these anyonic symmetries, we need to generalise the notion of the bond dimension and define how it interacts with the symmetry. Thus, we implement away of converting integers to symmetric spaces of the given dimension, which provides a crude guess for how the final MPS would distribute its Schmidt spectrum.
+In order to use these anyonic symmetries, we need to generalise the notion of the bond dimension and define how it interacts with the symmetry.
+Thus, we implement away of converting integers to symmetric spaces of the given dimension, which provides a crude guess for how the final MPS would distribute its Schmidt spectrum.
```
leburgel:
```suggestion
Thus, we implement a way of converting integers to symmetric spaces of the given dimension, which provides a crude guess for how the final MPS would distribute its Schmidt spectrum.
```

examples/classic2d/1.hard-hexagon/main.jl
```diff
@@ -68,16 +72,17 @@ function scaling_simulations(
     entropies = similar(Ds, Float64)
     correlations = similar(Ds, Float64)
     alg = VUMPS(; verbosity, tol, alg_eigsolve)
+    sector = unit(sectortype(mpo)) # dominant correlation functions are in the trivial sector
```
leburgel:
```suggestion
    sector = unit(sectortype(mpo)) # in this example the dominant correlation functions are in the trivial sector
```
Just to not make it sound like this is always the case.

src/algorithms/transfer_spectrum.jl
```diff
@@ -0,0 +1,179 @@
...
+
+- `above::InfiniteMPS`: the state for the "above" leg of the mixed transfer matrix.
+- `below::InfiniteMPS=above`: the state for the "below" leg; defaults to the pure transfer matrix of `above`.
+- `alg`: the eigensolver algorithm specification, resolved per sector via [`MatrixAlgebraKit.select_algorithm`](@extref MatrixAlgebraKit.select_algorithm).
+  This can be a KrylovKit algorithm instance (used verbatim for every sector), a `MatrixAlgebraKit.DefaultAlgorithm` or `NamedTuple` bundling keyword arguments,
+  or `nothing` (the default) to construct the eigensolver from the keyword arguments below.
+
+# Keyword arguments
+
+- `num_vals = 20`: the number of eigenvalues to compute. This can either be a single `Int`, which is used for every sector of the transfer space,
```
leburgel: Not the right place for this and not really worth it probably, but I wouldn't mind calling this `howmany`, in line with the `eigsolve` argument name :)

src/algorithms/transfer_spectrum.jl
```diff
@@ -0,0 +1,179 @@
...
+  or an `AbstractDict`/iterable of `sector => count` pairs to restrict the computation to specific sectors and request a different number of values per sector.
+- `oversampling = 10`: additive margin on the Krylov dimension beyond the number of values requested in a sector (see `krylovdim` below).
+- `oversampling_factor = 1`: proportionality factor between the Krylov dimension and the number of values requested in a sector (see `krylovdim` below).
+- `krylovdim`: the Krylov dimension of the eigensolver. Unless given explicitly, this is chosen adaptively per sector as
+  `max(Defaults.krylovdim, ceil(Int, oversampling_factor * num_vals) + oversampling)`, where `num_vals` is the number of values requested in that sector.
+- `kwargs...`: further keyword arguments (e.g. `tol`, `maxiter`) are forwarded to `MatrixAlgebraKit.default_algorithm` to build the eigensolver.
+  It is an error to pass eigensolver keyword arguments when `alg` is an algorithm instance.
```
leburgel:
```suggestion
  Passing eigensolver keyword arguments when `alg` is an algorithm instance is not allowed, and will result in an error.
```

src/algorithms/transfer_spectrum.jl
```diff
@@ -0,0 +1,179 @@
...
+    angles_approx[findall(angles_approx .== -1)] .= 1       # ∈ ]-1, 1]
+
+    return angles_approx .* π                               # ∈ ]-π, π]
+end
+
+"""
+    marek_gap(above::InfiniteMPS; sector=nothing, kwargs...)
+
+Compute the gap ```ϵ``` for the asymptotics of the transfer matrix, as well as the Marek gap
+```δ``` as a scaling measure of the bond dimension, along with the associated angle ```θ```.
```
leburgel:
```suggestion
Compute the gap `ϵ` for the asymptotics of the transfer matrix, as well as the Marek gap
`δ` as a scaling measure of the bond dimension, along with the associated angle `θ`.
```
Also, we should rename `marek_gap` to something more self-explanatory if we can, at some point.

src/algorithms/transfer_spectrum.jl
```diff
@@ -0,0 +1,179 @@
...
+
+By default this returns a `TensorKit.SectorDict` mapping each sector of the transfer spectrum
+to its `(ϵ, δ, θ)` triple. Passing a specific `sector` returns only that sector's triple. The
```
leburgel:
```suggestion
to its `(ϵ, δ, θ)` triplet. Passing a specific `sector` returns only that sector's triplet. The
```

leburgel (approved):
Few minor comments, but looks good to me.

docs/src/examples/quantum1d/5.haldane-spt/figure-1.png
lkdvos: Yeah, this is the non-injective one so it is more or less numerical noise, the main point is that the dominant value is degenerate, which unfortunately was more visible on the previous plot because it accidentally got wrapped around the branch cut, but I don't really feel like playing whack-a-mole with the RNG...

src/algorithms/transfer_spectrum.jl
lkdvos: changed it anyways :)

PR #454, opened 2026-07-10
[WIP] Basis Update Galerkin integrators

lkdvos:
This is a (very WIP) implementation of the BUG integrators by the Lubich group. There are two main motivations for this:

1. For imaginary time evolution (or non-hermitian systems), TDVP has a backwards time step that becomes unstable for dissipative systems. The BUG integrators only have forwards time steps, therefore are more robust to this.
2. The parallel version, while lower in accuracy, has the opportunity of more easily using available computational resources, which is particularly relevant on compute clusters.

I still need to clean this up and verify most of the code, but at least it looks like the tests are passing, so if nothing else this is an accidentally correct integrator.


Jutho:
I am happy to review when you are ready with this.

lkdvos:
I'm still testing quite a bit of things, I have to admit that I have only really looked slightly at the non-parallel BUG version just yet, I'll ping you once I've gone over it myself. I don't really want you to have to review my AI-code 😉 

I'll probably also split this PR into first just the BUG anyways, so we can go from there.

PR #455, opened 2026-07-10
perf: DMRG(2) with smarter adaptive tolerances

### LINKED ISSUE #311: U(2) DMRG performance unexpectedly slower than U(1) ITensor version

lkdvos:
This PR features a somewhat large refactor of the DMRG implementations, and finally fixes #311 hopefully in a more systematic way.

The main idea is that I want to use loose bounds on the krylov solvers, since currently everyone has the wrong assumption that ITensors is a lot faster than MPSKit, which is just due to the default settings.
Nevertheless, while for ITensorMPS you just have to know and read the docs to figure out that if you want to actually converge something for a non-gapped state, I wanted to keep some guarantees on stability and convergence rates.
Therefore, this PR introduces a bunch more knobs to determine adaptive tolerance and krylov settings based on the information that is available:

1. The tolerance of the krylov solver depends on the global convergence, and oversolving locally (especially for finite systems) has very little benefit.
2. The tolerance of the krylov solver should not go far below the truncation error incurred in adaptive DMRG schemes, since that error will dominate the eigenvector precision anyways.
3. The gap of the local effective system can be predicted by investigating the decay rate of the previous sweep, which is theoretically predicted to be geometric for Lanczos solvers. Using this, we can make an educated guess for how gapped the system is and adapt accordingly.

---

A secondary part of this PR is just me refactoring the DMRG implementations into a more readable and consolidated form, which I think overall just improves the code quailty.

---

### Benchmark summary

Backing up the claims in this PR, I've attached a number of results from investigations I did. The headline is that this new adaptive local eigensolver is an across-the-board improvement over the current main.
Doing a large amount of hyperparameter tests, it seems like the main point is that the results aren't too sensitive to the actual values of the various tolerance factors etc, but the general statement is still true that too high of a precision on the local solvers is not useful at the start of a run, but essential towards the end, specifically for gapless systems. The main thing is that the overhead of switching sides is actually quite large at higher bond dimensions (the SVD, hamiltonian building and environment transfers), such that the extra iterations in the krylov solvers are well worth it because that work has to be done anyways.

I think I'm happy with the current settings, which seems to at least balance having reasonably fast sweeps in the easy regimes and no stagnation in the harder regimes, so for me this PR is now ready.

<details>

<summary> Benchmark plots </summary>

heisenberg1_trivial
heisenberg1_2_trivial
hubbard_u1

</details>

### DIFF
--- a/src/algorithms/fixedpoint.jl
+++ b/src/algorithms/fixedpoint.jl
@@ -1,32 +1,23 @@
 # wrapper around KrylovKit.jl's eigsolve function
 
-"""
-    fixedpoint(A, x₀, which::Symbol; kwargs...) -> val, vec
-    fixedpoint(A, x₀, which::Symbol, alg) -> val, vec
-
-Compute the fixed point of a linear operator `A` using the specified eigensolver `alg`. The
-fixedpoint is assumed to be unique.
-"""
-function fixedpoint(A, x₀, which::Symbol, alg::Lanczos)
-    vals, vecs, info = eigsolve(A, x₀, 1, which, alg)
-
-    info.converged == 0 &&
-        @warnv 1 "fixed point not converged after $(info.numiter) iterations: normres = $(info.normres[1])"
-
-    return vals[1], vecs[1]
-end
-
-function fixedpoint(A, x₀, which::Symbol, alg::Arnoldi)
-    TT, vecs, vals, info = schursolve(A, x₀, 1, which, alg)
-
-    info.converged == 0 &&
-        @warnv 1 "fixed point not converged after $(info.numiter) iterations: normres = $(info.normres[1])"
-    size(TT, 2) > 1 && TT[2, 1] != 0 && @warnv 1 "non-unique fixed point detected"
-
-    return vals[1], vecs[1]
-end
-
 function fixedpoint(A, x₀, which::Symbol; kwargs...)
     alg = KrylovKit.eigselector(A, scalartype(x₀); kwargs...)
     return fixedpoint(A, x₀, which, alg)
+end
+
+"""
+    fixedpoint(A, x₀, which::Symbol; kwargs...) -> val, vec, info
+    fixedpoint(A, x₀, which::Symbol, alg) -> val, vec, info
+
+Compute the fixed point of a given linear operator `A` with initial guess `x₀`.
+The dominant eigenvector is assumed to be unique.
+"""
+function fixedpoint(A, x₀, which::Symbol, alg::Lanczos)
+    vals, vecs, info = eigsolve(A, x₀, 1, which, alg)
+    return vals[1], vecs[1], info
+end
+function fixedpoint(A, x₀, which::Symbol, alg::Arnoldi)
+    TT, vecs, vals, info = schursolve(A, x₀, 1, which, alg)
+    size(TT, 2) > 1 && TT[2, 1] != 0 && @warnv 1 "non-unique fixed point detected"
+    return vals[1], vecs[1], info
 end
--- a/src/algorithms/groundstate/dmrg.jl
+++ b/src/algorithms/groundstate/dmrg.jl
@@ -87,18 +87,15 @@ function local_update!(
         fixedpoint(H_effective, ac_old, :SR, alg_eigsolve)
     end
 
-    # convergence error: how much the eigensolve moved the center, `1 - |⟨old|new⟩|`.
-    ϵ_conv = abs(1 - abs(dot(ac_old, AC′)) / (norm(ac_old) * norm(AC′)))
-
     # 3. gauge
     ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge!(ψ, site, direction, AC′, alg.alg_gauge; normalize = true)
 
     # 4. bookkeeping: measured contraction factor per matvec, kept a strict contraction in (0, 1)
     decay_rate = clamp((first(info.normres) / ϵ_local)^(1 / max(1, info.numops)), 1.0e-3, 0.999)
 
-    @debug "DMRG local update" site direction numops = info.numops normres = first(info.normres) krylovdim = alg_eigsolve.krylovdim maxiter = alg_eigsolve.maxiter tol = alg_eigsolve.tol decay_rate ϵ_local ϵ_conv
+    @debug "DMRG local update" site direction numops = info.numops normres = first(info.normres) krylovdim = alg_eigsolve.krylovdim maxiter = alg_eigsolve.maxiter tol = alg_eigsolve.tol decay_rate ϵ_local
 
-    return ψ, ϵ_local, ϵ_trunc, decay_rate, ϵ_conv
+    return ψ, ϵ_local, ϵ_trunc, decay_rate
 end
 
 """
@@ -167,9 +164,6 @@ function local_update!(
         (newA2center, info)
     end
 
-    # convergence error: how much the eigensolve moved the two-site center, `1 - |⟨old|new⟩|`.
-    ϵ_conv = abs(1 - abs(dot(ac2, newA2center)) / norm(newA2center))
-
     # 2. gauge: truncated SVD split back into single-site tensors and install;
     #           the discarded weight is the truncation error
     ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge2!(ψ, pos, direction, newA2center, alg.alg_gauge; normalize = true)
@@ -177,9 +171,9 @@ function local_update!(
     # 3. bookkeeping: measured contraction factor per matvec, kept a strict contraction in (0, 1)
     decay_rate = clamp((first(info.normres) / ϵ_local)^(1 / max(1, info.numops)), 1.0e-3, 0.999)
 
-    @debug "DMRG2 local update" pos direction numops = info.numops normres = first(info.normres) krylovdim = alg_eigsolve.krylovdim maxiter = alg_eigsolve.maxiter tol = alg_eigsolve.tol decay_rate ϵ_local ϵ_trunc ϵ_conv
+    @debug "DMRG2 local update" pos direction numops = info.numops normres = first(info.normres) krylovdim = alg_eigsolve.krylovdim maxiter = alg_eigsolve.maxiter tol = alg_eigsolve.tol decay_rate ϵ_local ϵ_trunc
 
-    return ψ, ϵ_local, ϵ_trunc, decay_rate, ϵ_conv
+    return ψ, ϵ_local, ϵ_trunc, decay_rate
 end
 
 # Per-algorithm sweep geometry: single-site DMRG updates all `length(ψ)` sites (endpoints once,
@@ -202,21 +196,19 @@ function find_groundstate!(
 
     Tr = real(scalartype(ψ))
     n = _num_updates(alg, ψ)
-    ϵ_locals = ones(Tr, n)      # local gradient norms (drive the adaptive eigensolver tolerance)
-    ϵ_global = maximum(ϵ_locals)
-    ϵ_convs = ones(Tr, n)       # local convergence errors (update magnitude; drive the stop test)
-    ϵ_conv = maximum(ϵ_convs)
+    ϵ_locals = ones(Tr, n)      # local Galerkin errors (drive both the eigensolver tol and the stop test)
+    ϵ_global = maximum(ϵ_locals) # sweep-wide Galerkin error, the convergence measure
     ϵ_truncs = zeros(Tr, n)     # local truncation error
     decay_rates = zeros(n)      # local observed decay rate of eigensolver
     fwd, bwd = _sweep_ranges(alg, ψ)
 
     LoggingExtras.withlevel(; alg.verbosity) do
-        @infov 2 loginit!(log, ϵ_conv, expectation_value(ψ, H, envs))
+        @infov 2 loginit!(log, ϵ_global, expectation_value(ψ, H, envs))
         for iter in 1:(alg.maxiter)
             @timeit timeroutput "sweep" begin
                 # left-to-right
                 for pos in fwd
-                    ψ, ϵ_locals[pos], ϵ_truncs[pos], decay_rates[pos], ϵ_convs[pos] =
+                    ψ, ϵ_locals[pos], ϵ_truncs[pos], decay_rates[pos] =
                         local_update!(
                         pos, Val(:right),
                         ψ, H, alg, envs,
@@ -228,7 +220,7 @@ function find_groundstate!(
 
                 # right-to-left
                 for pos in bwd
-                    ψ, ϵ_locals[pos], ϵ_truncs[pos], decay_rates[pos], ϵ_convs[pos] =
+                    ψ, ϵ_locals[pos], ϵ_truncs[pos], decay_rates[pos] =
                         local_update!(
                         pos, Val(:left),
                         ψ, H, alg, envs,
@@ -238,26 +230,26 @@ function find_groundstate!(
                     ϵ_global = maximum(ϵ_locals)
                 end
             end
-            ϵ_conv = maximum(ϵ_convs)
+            ϵ_global = maximum(ϵ_locals)
 
             ψ, envs = @timeit timeroutput "finalize" alg.finalize(
                 iter, ψ, H, envs
             )::Tuple{typeof(ψ), typeof(envs)}
 
-            if ϵ_conv <= alg.tol
+            if ϵ_global <= alg.tol
                 @infov 4 timeroutput
-                @infov 2 logfinish!(log, iter, ϵ_conv, expectation_value(ψ, H, envs))
+                @infov 2 logfinish!(log, iter, ϵ_global, expectation_value(ψ, H, envs))
                 break
             end
             if iter == alg.maxiter
                 @infov 4 timeroutput
-                @warnv 1 logcancel!(log, iter, ϵ_conv, expectation_value(ψ, H, envs))
+                @warnv 1 logcancel!(log, iter, ϵ_global, expectation_value(ψ, H, envs))
             else
-                @infov 3 logiter!(log, iter, ϵ_conv, expectation_value(ψ, H, envs))
+                @infov 3 logiter!(log, iter, ϵ_global, expectation_value(ψ, H, envs))
             end
         end
     end
-    return ψ, envs, ϵ_conv
+    return ψ, envs, ϵ_global
 end
 
 function find_groundstate(ψ, H, alg::Union{DMRG, DMRG2}, envs...; kwargs...)
--- a/src/algorithms/groundstate/dmrg.jl
+++ b/src/algorithms/groundstate/dmrg.jl
@@ -150,19 +150,7 @@ function local_update!(
     Heff = @timeit timeroutput "AC2_hamiltonian" AC2_hamiltonian(pos, ψ, O, ψ, envs)
 
     kind = direction === Val(:right) ? :ACAR : :ALAC
-    ac2 = normalize!(AC2(ψ, pos; kind))
-    # Two-site Galerkin gradient `ϵ_local`, the direct analogue of single-site `calc_galerkin`: the
-    # part of `Heff * ac2` that points off the current left-canonical manifold, obtained by
-    # projecting onto the LEFT null space of `AL[pos]` only. Projecting onto BOTH the left and right
-    # null spaces would keep only the (null, null) bond-expansion corner — the directions a
-    # truncated SVD discards — which collapses to the truncation floor after each local solve and
-    # so spuriously signals convergence after a single sweep.
-    #
-    # `Heff * ac2` only feeds this gradient; it must NOT seed the eigensolver. Seeding with
-    # `Heff * ac2` is power-iteration intuition (amplifies the largest-magnitude eigenvector), which
-    # is wrong for `:SR` (smallest algebraic): it de-weights the ground state and, when the
-    # effective ground eigenvalue is 0, annihilates it exactly so `:SR` converges to the smallest
-    # nonzero eigenvalue. Seed with the plain center `ac2` instead (as single-site DMRG's `ac_old`).
+    ac2 = AC2(ψ, pos; kind)
     AC2′ = normalize!(Heff * ac2)
     project_complement!(AC2′, ψ.AL[pos])
     ϵ_local = norm(AC2′)
--- a/src/algorithms/groundstate/dmrg.jl
+++ b/src/algorithms/groundstate/dmrg.jl
@@ -151,16 +151,20 @@ function local_update!(
 
     kind = direction === Val(:right) ? :ACAR : :ALAC
     ac2 = normalize!(AC2(ψ, pos; kind))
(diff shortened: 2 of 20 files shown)


lkdvos:
Here I've added some results on the comparisons for the sweeptimes, but also on the energy reached after a given walltime, in an attempt to not just measure what the fastest way is to get 10 sweeps done.
What's interesting is that we now consistently outperform ITensorMPS, even without symmetries, but that probably we will have to investigate the single-site version a bit more, since that seems to not perform the way I'd expect.

I'd also say that it seems really hard to make generalizations about the actual walltimes it takes to get to a given energy, since the faster sweeps get compensated with the necessity for more sweeps. In general, this is not faster everywhere, but I think it at least feels faster because the sweeps take a little less time, and it makes a naive benchmark no longer favor the ITensor version that much.


sweeptime_heisenberg_s1
energy_heisenberg_s1
energy_heisenberg_s1_2
sweeptime_heisenberg_s1_2

src/algorithms/fixedpoint.jl
```diff
@@ -1,32 +1,23 @@
...
-Compute the fixed point of a linear operator `A` using the specified eigensolver `alg`. The
-fixedpoint is assumed to be unique.
+Compute the fixed point of a given linear operator `A` with initial guess `x₀`.
+The dominant eigenvector is assumed to be unique.
 """
 function fixedpoint(A, x₀, which::Symbol, alg::Lanczos)
     vals, vecs, info = eigsolve(A, x₀, 1, which, alg)
-
-    info.converged == 0 &&
-        @warnv 1 "fixed point not converged after $(info.numiter) iterations: normres = $(info.normres[1])"
```
borisdevos, line 14: How come you remove these non-convergence warnings? Do they appear elsewhere if you have a high enough verbosity?

src/algorithms/groundstate/dmrg.jl
```diff
@@ -178,84 +135,129 @@ function DMRG2(;
...
+                # right-to-left
+                for pos in bwd
+                    ψ, ϵ_locals[pos], ϵ_truncs[pos], decay_rates[pos], ϵ_convs[pos] =
+                        local_update!(
+                        pos, Val(:left),
+                        ψ, H, alg, envs,
+                        ϵ_global, ϵ_truncs[pos], decay_rates[pos],
+                        timeroutput
+                    )
+                    ϵ_global = maximum(ϵ_locals)
```
borisdevos, line 228: What's the role of the local and global errors now? It seems that `ϵ_conv` is the only one that matters now

src/algorithms/groundstate/dmrg.jl
```diff
@@ -64,82 +66,40 @@ function DMRG(;
...
-            end
-            ϵ = maximum(ϵs)
+    # 3. gauge
+    ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge!(ψ, site, direction, AC′, alg.alg_gauge; normalize = true)
 
-            ψ, envs = @timeit timeroutput "finalize" alg.finalize(
-                iter, ψ, H, envs
-            )::Tuple{typeof(ψ), typeof(envs)}
+    # 4. bookkeeping: measured contraction factor per matvec, kept a strict contraction in (0, 1)
+    decay_rate = clamp((first(info.normres) / ϵ_local)^(1 / max(1, info.numops)), 1.0e-3, 0.999)
```
borisdevos, line 94: Is it impossible here for `ϵ_local` to be exactly 0? Or is that part of why the debug log has been added?

src/states/orthoview.jl
```diff
@@ -264,19 +274,67 @@ be a [`TruncatedAlgorithm`](@extref MatrixAlgebraKit.TruncatedAlgorithm) to trun
...
+
+@doc """
+    gauge!(ψ, pos, direction, AC, [alg]; normalize = false) -> ψ, ϵ
+
+Direction-dispatching wrapper around [`left_gauge!`](@ref) / [`right_gauge!`](@ref): gauge an
+updated center tensor `AC` at site `pos` and install it into `ψ`, shifting the gauge center past
+`pos` to the right for `direction = Val(:right)` and to the left for `Val(:left)`. `alg` and
+`normalize` are forwarded unchanged, and the truncation error `ϵ` is returned as-is.
+"""
+function gauge!(ψ::AbstractFiniteMPS, pos::Int, ::Val{Dir}, AC, alg = Defaults.alg_orth(); kwargs...) where {Dir}
```
borisdevos, line 305: Not that it really matters, but the way `gauge!` dispatches to certain directions is different to the recently added `changebond!`, which truly dispatches on `Val{:left}` or `Val{:right}`. Here, you compare symbols at runtime to then dispatch to functions with different names. Was this a deliberate choice?

src/states/finitemps.jl
```diff
@@ -343,8 +343,15 @@ Base.@propagate_inbounds function Base.getindex(ψ::FiniteMPS, i::Int)
     end
 end
 
-# TODO: check where gauge center is to determine efficient kind
-AC2(psi::FiniteMPS, site::Int) = psi.AC[site] * _transpose_tail(psi.AR[site + 1])
+function AC2(psi::FiniteMPS, site::Int; kind = :ACAR)
```
borisdevos, line 346: Same comment as below (oops) on the dispatch design choice.

src/utility/dynamictols.jl
```diff
@@ -77,4 +79,149 @@ function _updatetol(alg::DefaultAlgorithm, tol::Real)
...
+instantiate_algorithm(alg, args...) = alg
+
+# a `DynamicTol` wrapper is resolved to a concrete solver by tightening its tolerance with the
+# global gradient norm (the per-sweep, tol-only "legacy" behaviour, now driven per site).
+function instantiate_algorithm(
+        alg::DynamicTol, decay_rate::Real, g_local::Real, g_global::Real, eps_trunc::Real
+    )
+    tol = clamp(g_global * alg.tol_factor, alg.tol_min, alg.tol_max)
+    return _updatetol(alg.alg, tol)
+end
```
borisdevos: This behavior is very different to what the user would previously expect from `DynamicTol`, no? I think someone can write `DMRG(; alg_eigsolve = DynamicTol(Lanczos()))`, and then expect line 65's formula to still hold (using 1/sqrt(iter)).

src/algorithms/groundstate/dmrg.jl
```diff
@@ -178,84 +135,129 @@ function DMRG2(;
...
+    project_complement!(AC2′, ψ.AL[pos])
+    project_complement_right!(AC2′, _transpose_tail(ψ.AR[pos + 1]))
+    ϵ_local = norm(AC2′)
+
+    # 1. local two-site update
+    alg_eigsolve = instantiate_algorithm(alg.alg_eigsolve, decay_rate, ϵ_local, ϵ_global, ϵ_trunc)
+    newA2center, info = @timeit timeroutput "AC2_eigsolve" begin
+        _, newA2center, info = fixedpoint(Heff, HAC2, :SR, alg_eigsolve)
+        (newA2center, info)
+    end
```
borisdevos: Since H * AC2 is used as an initial vector to the fixed point equation, this one application should also count to the number of matrix-vector multiplications, but this isn't being logged. Maybe this 1 matvec doesn't matter in the grand scheme of things though 🤷

borisdevos (commented):
This is amazing work, Lukas! The benchmarks also look very promising. 

I was just wondering whether we should be more verbose about how `AdaptiveKrylov` is now the default everywhere. Also, this is not being compared explicitly in our tests by name, though I'm not sure what the criteria would be.

The refactoring indeed makes the code much easier to follow! The title of the PR could also added "performance" arguably ;)

You should definitely let others review this PR as well, since I'm no expert really, but hopefully I brought up some things with substance.

src/algorithms/groundstate/dmrg.jl
lkdvos: In principle it is not impossible, although it is very unlikely, as this is computed from the norm of something projected onto a nullspace, so for any floating point precision you'd expect at least order `eps`. Nevertheless, the clamp gracefully handles that too: 
```julia
julia> clamp(1/0.0, 1e-2, 1)
1.0

julia> clamp((1/0.0)^3, 1e-2, 1)
1.0
```

src/algorithms/fixedpoint.jl
lkdvos: They do not - the problem is that I deliberately don't want to consider the tolerance alone as a convergence criterion. For example, the main reason why we couldn't just replicate the ITensor default settings is that you couldn't use the `krylovdim` + `maxiter` combination to bound the work without spitting out these warnings.
The entire point of the adaptive scheme is basically to figure out good `krylovdim` and `maxiter` and `tol` settings such that everything behaves and one of these is hit.

I think this is all part of a more general paradigm shift where in nested iterative schemes (outer sweep, inner krylov solver), it actually does make sense to not converge the inner solvers completely, and we don't want to spam the console for that.

Similarly, (but slightly on a separate note) for DMRG, it is actually very common to simply set a maximal number of iterations, and reaching that does not necessarily mean that a warning should be printed. If there is no way to turn these warnings off, that becomes annoying.

src/algorithms/groundstate/dmrg.jl
lkdvos: The `ϵ_global` is used in the adaptive scheme, and signifies more or less the norm of the gradient projected onto the tangent space. On the other hand, the `ϵ_conv` gives how much the tensors have drifted.
I really wanted to avoid to use this latter quantity, but that was what DMRG2 was using before, and the problem is that the former can become zero even if the state is not converged, as it becomes zero for any eigenstate.
This broke the excitation tests, because `FiniteExcited` basically adds a penalty term to the ground state, which does not change the eigenvectors, and therefore the algorithm immediately terminates.
A different solution would be to just add a `miniter` as well, which I'm definitely also open to

src/states/orthoview.jl
lkdvos: Ish - for `changebond!` the functions with different names didn't exist yet, so there it seemed unnecessary to first introduce the function names and then use the `Val` dispatch, while here I wanted to use a flag to dispatch, but already had the function names. It's mostly a matter of trying to not be too backwards-breaking?

src/states/finitemps.jl
lkdvos: This is in line with the function that already existed for the infinite case:
https://github.com/QuantumKitHub/MPSKit.jl/blob/4e0e89191f98e53ecd6482db6d5320f87350f40c/src/states/infinitemps.jl#L237-L245

The main difference is that the return type does not change for different values of this keyword argument, so the `Val` dispatch is not really necessary and we can just use runtime to distinguish between the two. The `left_gauge` and `right_gauge` functions on the other hand return their arguments in different orders, so to avoid type-instability I added the `Val` dispatch. In hindsight, I guess at the `gauge!` level this is not really relevant since it just returns the state, so I agree that it might be reasonable to change that to a keyword argument too, to take some strain from the compiler.

src/utility/dynamictols.jl
lkdvos: Good point, I somehow forgot about that since my intention was to completely supersede the `DynamicTol` implementation. I'll make sure to restore this though, thanks!

src/algorithms/groundstate/dmrg.jl
lkdvos: It definitely does, we are now doing on the order of 5-10 of them so it is 10-20%. However, the debug information is mostly for inspecting the `eigsolve` behavior, since that is also what is configured with the adaptive tolerances (krylovdim=3 would end up doing 4 matvecs if we count the initial one), so I think debug-wise this is probably fine?

lkdvos:
Thanks a lot @borisdevos for taking the time to go over this, the feedback is definitely appreciated!

Some more things I would love inputs on:

- Looking at the benchmarks right now, it is actually not entirely clear that these looser defaults are really helping. Of course, if the metric is to get fast sweeps this makes them a lot faster, but if the metric is to get to the groundstate faster, this doesn't seem to really be the case, especially in the gapped case or for higher bond dimensions. I might redo an actual hyperparameter sweep in the current implementation and see if I can actually get some more quantitative data on this.
- I really don't know what to do with the convergence criterion. The galerkin error seems more principled to me, but especially in the twosite context I'm not even sure if I'm defining this correctly, since the correct thing might be to project onto the twosite left nullspace? Maybe @Jutho can say something more about this. Ultimately I'm happy to in the future work towards making this configurable, but I don't want to tackle that in this PR.
- I have a bunch more tricks up my sleeve to make this faster, which mostly have to do with the way the environments are constructed and stored, which arguably will make a much larger difference across the board than this hyperparameter tweaking, so I don't want to spend **too** much time dwelling on this. I'm probably going to try and add some of these benchmark results to the documentation, where for the comparison with ITensor I will just use the "fixed krylov" settings, which is also what TenPy settled on for their benchmarks.
- The `conv` vs `galerkin` errors currently have a different scale - the former is `1 - dot(x, y)` while the latter is `norm = sqrt(dot(...))`. However, if I naively square root the former the behavior doesn't really fix my excitations immediately converging issue...

src/algorithms/groundstate/dmrg.jl
```diff
@@ -64,82 +66,40 @@ function DMRG(;
...
-                    @timeit timeroutput "AC_eigsolve" begin
-                        Hac = AC_hamiltonian(pos, ψ, H, ψ, envs)
-                        _, AC′ = fixedpoint(Hac, ψ.AC[pos], :SR, alg_eigsolve)
-                    end
-
-                    # 3. gauge (QR center-move or truncated SVD, selected by `alg_gauge`)
-                    @timeit timeroutput "gauge" left_gauge!(ψ, pos, AC′, alg.alg_gauge; normalize = true)
-                end
+    # convergence error: how much the eigensolve moved the center, `1 - |⟨old|new⟩|`.
+    ϵ_conv = abs(1 - abs(dot(ac_old, AC′)) / (norm(ac_old) * norm(AC′)))
```
Jutho: Are the norms of ac_old and AC' necessary? Aren't they 1 automatically?

src/algorithms/groundstate/dmrg.jl
Jutho: Is it useful to have H * AC2 instead of just AC2 as a starting vector? For transfer matrix fixed points, this should be a strictly better initial guess, but I am a bit worried for Hamiltonians which are such that the ground state energy is (close to) zero, and so H * AC2 might actually project out the most relevant contribution to the eigenvector when you are close to convergence.

src/algorithms/groundstate/dmrg.jl
```diff
@@ -178,84 +135,129 @@ function DMRG2(;
...
+function local_update!(
+        pos, direction,
+        ψ, O, alg::DMRG2, envs,
+        ϵ_global, ϵ_trunc, decay_rate,
+        timeroutput
+    )
+    Heff = @timeit timeroutput "AC2_hamiltonian" AC2_hamiltonian(pos, ψ, O, ψ, envs)
+
+    kind = direction === Val(:right) ? :ACAR : :ALAC
+    ac2 = normalize!(AC2(ψ, pos; kind))
```
Jutho: This comment probably belongs elsewhere, but it currently seems like `AC2_hamiltonian` doesn't need any direction or kind info, `AC2` needs a `kind` kwarg, and then `gauge2` below needs a `direction` kwarg. Is there a convenient non-breaking way to make this a bit more consistent? I am in particular confused about how `AC2_hamiltonian` knows whether it is the Hamiltonian for `pos:pos+1` or `pos-1:pos`, and if there is just a fixed choice, why is this not used the same for `AC2`? Also, I would then think that, depending on direction, the position index for `AC2_hamiltonian` would need to be `pos` or `pos-1`.

Jutho: Ok, I think I start to see how it works. I assume all of these functions always use `pos:pos+1`, but then it's about how they compute the relevant quantities. Never mind this comment then.

src/algorithms/groundstate/dmrg.jl
```diff
@@ -178,84 +135,129 @@ function DMRG2(;
...
+
+# Per-algorithm sweep geometry: single-site DMRG updates all `length(ψ)` sites (endpoints once,
+# interior twice), whereas two-site DMRG2 updates the `length(ψ) - 1` bonds. `_num_updates`
+# gives the number of per-update bookkeeping slots and `_sweep_ranges` the forward/backward
+# index ranges; everything else in the sweep is shared.
+_num_updates(::DMRG, ψ) = length(ψ)
+_num_updates(::DMRG2, ψ) = length(ψ) - 1
+
+_sweep_ranges(::DMRG, ψ) = (1:(length(ψ) - 1), length(ψ):-1:2)
+_sweep_ranges(::DMRG2, ψ) = (1:(length(ψ) - 1), (length(ψ) - 2):-1:1)
```
Jutho, line 185: It doesn't really matter, but this is a bit inconsistent between `DMRG` and `DMRG2`, I would have expected:


```suggestion
_sweep_ranges(::DMRG2, ψ) = (1:(length(ψ) - 2), (length(ψ) - 1):-1:2)
```

Also, for both `DMRG and `DMRG2`, it is indeed true that the first (`1`) and last site (`length(ψ)` for `DMRG` and `length(ψ)-1` for `DMRG2`) only needs to be visited in either forward or backward, e.g. backwards only goes to site 2, and then the next forward starts again at site 1. However, in the very last iteration, you do want to also visit site 1 after the backward sweep. (In most cases this doesn't matter, since anyway if Dl * d = Dr, there is actually no point in optimizing that site, but there might be problems where d, the physical dimension is very large, and where this would have an effect).

lkdvos: While it is definitely possible to include site 1 in the last sweep, I don't really see how this makes any real change, except for where the tolerance is checked. If we do that 1 extra update, doesn't that then mean that now site 2 might need updating?
I definitely might be missing something, but somehow I always kind of assumed that this is completely arbitrary, and we could also just check convergence after every local update and stop at whatever site reaches that?

src/algorithms/groundstate/dmrg.jl
lkdvos: Oh you are right, I was definitely thinking of the transfer matrix case.

src/algorithms/groundstate/dmrg.jl
Jutho: Ok yes, I assume it is quite arbitrary and doesn't really matter for optimization / DMRG. I guess I am more thinking of TDVP/splitting integrator.

src/algorithms/groundstate/dmrg.jl
Jutho: Does it make sense for `ϵ_global = norm(ϵ_locals)`. I think the total projected norm is in fact the 2-norm of the local ones. 

As for using `ϵ_conv` and the issue you describe, this sounds like a deeper problem. If `ϵ_global = 0`, then all local eigenvalue problems should also be satisfied by their initial guess, and should return immediately. And if they don't, as in the case you describe where you add a penalty term but still use the ground state as initial guess, that is really relying on some numerical noise to mix in new eigenvector components in the Krylov subspace, and it does not sound very robust. For finding excited states in this way, I think you really want to explicitly perturb the state away from the ground state as initial guess.

src/utility/dynamictols.jl
```diff
@@ -77,4 +79,149 @@ function _updatetol(alg::DefaultAlgorithm, tol::Real)
...
+    tol = clamp(max(trunc_tol, conv_tol), alg.tol_min, alg.tol_max)
+
+    # the measured decay rate must be a genuine contraction in (0, 1); a non-positive or ≥ 1
+    # rate signals an uninitialized or hard/stalling bond → fall back to the largest budget.
+    ρ = clamp(decay_rate, 0.0, 1.0)
+    if !(0 < ρ < 1)
+        krylovdim = iszero(ρ) ? alg.krylovdim_min : alg.krylovdim_max
+        maxiter = iszero(ρ) ? alg.iter_min : alg.iter_max
+        return (T ? Lanczos : Arnoldi)(; alg.orth, krylovdim, maxiter, tol, alg.eager, alg.verbosity)
+    end
```
Jutho, line 217: With `decay_rate = clamp((first(info.normres) / ϵ_local)^(1 / max(1, info.numops)), 1.0e-3, 0.999)` above, can `ρ` ever fall outside (or even reach) the limits 0.0 or 1.0?

lkdvos: yes, the very first iteration I don't have any history so I set it to 0 to detect this. I'll add a comment here to clarify!

src/algorithms/groundstate/dmrg.jl
lkdvos: I think what is subtle about this is that the local errors are computed in a non-simultaneous way, so really only the one that got computed last is fully correct with the current state, since we are updating as we go along the sweep. That's also why I'm using `maximum` here rather than taking the norm. A different reason that is more pragmatic is that I want something that is size-independent, as otherwise the various tolerances and settings would have to scale inversely with system size. This is less of a problem in the infinite settings, but here I think it makes more sense to have a 50-site and a 500-site have the same local tolerances, not a global one?

Jutho: Yes, in particular the size dependence is also something that occurred to me after writing this comment, so I think `maximum` is a suitable choice.

lkdvos:
I've updated the implementation and the PR description, benchmark results at the top.

The important changes made here:
- using galerkin error also for twosite convergence criterion works, but probably we should make some of this configurable in the future
- `FiniteExcited` should seed from perturbed states to not make that stagnate

PR #460, opened 2026-07-14
DMRG3S: strictly single-site DMRG subspace expansion (Hubig et al. 2015)

VinceNeede:
## Summary

Implements DMRG3S — strictly single-site DMRG with subspace expansion — from Hubig, McCulloch, Schollwöck & Wolf, [Phys. Rev. B 91, 155115 (2015)](https://doi.org/10.1103/PhysRevB.91.155115). Unlike CBE (`alg_expand`), which enriches the bond *before* the local eigensolve, DMRG3S enriches it *after*, using a perturbation built from the just-optimized tensor, the local Hamiltonian, and the environment. This lets single-site DMRG escape local minima caused by missing quantum-number sectors that CBE-style pre-expansion can't reach.

## This PR also targets TDVP, pending design agreement

DMRG only for now. TDVP support belongs in this same PR, but I'd like design agreement on the interface below first, since it directly shapes how it plugs into TDVP too. Docs are also on hold for the same reason — no point documenting an interface that might still change.

## What's here

- Refactored `find_groundstate!`'s left-to-right and right-to-left sweeps from two separate, near-duplicate loops into a single loop over `direction in (:right, :left)`, dispatching on `Val`. No behavioral change; this just made room for the new expansion step without doubling its code.
- `alg_post_expand` field on `DMRG`, dispatching a new `post_expand!` step after the eigensolve, alongside the existing pre-eigensolve `alg_expand`. Default (`NoExpand()`) reproduces current behavior exactly.
- `DMRG3S`, with a composable `NoiseSchedule` for the mixing factor (`ExponentialDecay`, `Warmup`, `∘` for combining them).
- `alg_expand` and `alg_post_expand` are independent and can be used together or separately.
- A warning fires if either expansion mechanism is active without a truncating `alg_gauge`, since the bond would otherwise grow unboundedly each sweep.
- Tests: a bond-growth smoke test alongside the existing CBE tests, and a reproduction of the paper's Sec. VII A example (plain single-site DMRG gets stuck at E ≈ -6.35, DMRG3S recovers E ≈ -8.68).

## Open design questions for maintainers

1. **`alg_post_expand` sentinel**: I used a dedicated `NoExpand()` type rather than `nothing` (which `alg_expand` uses). This lets `NoExpand` carry its own `post_expand!` method for dispatch, but it's an inconsistency with `alg_expand`'s convention worth flagging.
2. **Expand+gauge+neighbor-pad are fused into one `post_expand!` call** rather than kept separate, because splitting them left the MPS briefly inconsistent between steps (caught via a real bug during development — happy to share details). This does mean `post_expand!` needs `alg_gauge` threaded in as a parameter, coupling it to the gauge step more tightly than `alg_expand` is.

## TODO before merge

- [x] Docs
- [x] TDVP support (Decided to add it in another PR)
- [x] Decide on Q1–Q2 above


lkdvos:
Thanks for this, I had a brief look and I think it already looks quite well.
I'll try and make some time for a proper review either tomorrow or the day after, but mostly I'm wondering if it might make sense to wait for #455, as that has a similar kind of refactor of the DMRG machinery to avoid duplication.

Considering the expansion algorithm itself - without having looked in a lot of detail so definitely feel free to tell me I'm wrong - I think it could be nice to avoid having both an `alg_gauge` and an `alg_postexpand`, and simply merge the two together. As always, naming things is really the hardest part, but it feels a bit clumsy to have the if-guard in the generic local step, and to have to pass 2 algorithms to the `perturb!` body.
Do you think it might be possible to simply put the `alg_gauge` inside of the `DMRG3S`, and then keep only one of these around that we dispatch on? I'm not sure how blasphemous it would be to call these both `alg_gauge`...
We might also try and discuss more carefully about the invalid state though, the `FiniteMPS` struct and the interaction with the environments is quite subtle, and it might still be that it is still fixable to separate the two cleanly as well.

For the TDVP implementation, I definitely agree to flush this one out first, and honestly it's probably easier to tackle and review that in a separate PR.

Considering your other question, I'm definitely happy to add the `NoExpand`, and since we are in the middle of a breaking cycle, I would be happy to change the `alg_expand` to use that too.

VinceNeede:
Hi @lkdvos,

Pushed a sketch of merging `alg_gauge`/`alg_post_expand` along the lines you suggested. The dispatch side is in place: `NoExpand`/`DMRG3S` both wrap an inner gauge algorithm, `gauge!`/`gauge2!` dispatch on whichever one `alg_gauge` is, and `local_update!` ends up with a single unconditional `gauge!(...)` call — no if-guard, no second algorithm argument. The noise schedule advances once per outer iteration via `_update_alg_gauge`, threaded through `find_groundstate!`.

I haven't wired up `DMRG`'s constructor yet, so this isn't runnable end-to-end through the public API — that's next, and here's the direction I'm planning to take it.

Since `DMRG`'s `trscheme`/`alg_svd`/`alg_orth` also determine the inner gauge, a user could in principle construct `DMRG3S` with its own gauge already set (`DMRG3S(noise, schedule, some_gauge)`) and pass a different `trscheme` to `DMRG`, giving two conflicting sources for the same thing. My plan is to make `trscheme`/`alg_svd`/`alg_orth` keyword arguments of `NoExpand`/`DMRG3S` themselves, building the inner gauge there — rather than have them accessible from `DMRG`'s constructor directly. That way each gauge object is self-contained, and the conflict can't arise in the first place. Let me know if you'd rather keep `trscheme` etc. as `DMRG`-level kwargs instead (with some other rule for resolving the conflict), otherwise I'll go ahead and wire it up this way.

src/algorithms/groundstate/dmrg.jl
```diff
@@ -203,6 +203,7 @@ function find_groundstate!(
     LoggingExtras.withlevel(; alg.verbosity) do
         @infov 2 loginit!(log, ϵ_global, expectation_value(ψ, H, envs))
         for iter in 1:(alg.maxiter)
+            alg_gauge = _update_alg_gauge(alg.alg_gauge, iter, ϵ_global)
```
lkdvos: A comment here but this involves the other parts of this code:

It might be beneficial to consider passing `iter` to the `local_update!` function, such that this alg update can be handled inside that function, similar to how that is handled for the dynamic eigensolver settings.
I very much like the warmup and decay options, so it might even make sense to also unify that into there anyways (in a follow-up PR for sure though, let's try and keep things moving).

The main reason I'm suggesting this is that it is slightly unfortunate to have to pass `alg_gauge`, even though that is already contained in `alg`, so it feels slightly more natural to me to pass `iter`.

src/algorithms/post_expand/dmrg3s.jl
```diff
@@ -0,0 +1,113 @@
...
+end
+
+DMRG3S(noise, schedule) = DMRG3S(noise, schedule, nothing)
+
+function _update_alg_gauge(alg::DMRG3S, iter, ϵ)
+    noise = alg.schedule(alg.noise, iter, ϵ)
+    return iszero(noise) ? NoExpand(alg.alg_gauge) : DMRG3S(noise, alg.schedule, alg.alg_gauge)
+end
+
+function _get_combiner(::Type{T}, V1, V2) where {T}
```
lkdvos: I think this function is the same as `fuser`, which already exists somewhere in our utility.jl file I think. You can recover the fused space via `only(domain(F))` I think.

src/algorithms/post_expand/dmrg3s.jl
```diff
@@ -0,0 +1,113 @@
...
+
+function gauge!(pos::Int, ::Val{:right}, ψ::AbstractFiniteMPS, H, envs, AC, alg::DMRG3S; normalize = true)
+    El = leftenv(envs, pos, ψ)
+    Hi = H[pos]
+    α = alg.noise
+    T = promote_type(scalartype(ψ), scalartype(Hi))
+    V = right_virtualspace(AC)
+    combiner, Vpert = _get_combiner(T, V, right_virtualspace(Hi))
+
+    @plansor pert[-1 -2; -3] := α * El[-1 1; 2] * AC[2 3; 4] * Hi[1 -2; 3 5] * combiner[4, 5; -3]
```
lkdvos: I think this contraction might be the same as our `AC_projection` contractions, and it could be beneficial to reuse that - the main reason is that we also support density-matrix MPS, which have 2 physical legs and would break this contraction, and reusing that function automatically fixes that.

lkdvos (commented):
Thanks for rebasing onto that other work!

Considering the DMRG constructor, I think I agree that we should exclude being able to set conflicting schemes, but unless I'm missing something that would already not be possible - the gauging algorithm is either a (truncated) orth algorithm or DMRG3S, and in the latter case there is only a single (truncated) orth algorithm in the DMRG3S.
If at all possible, I would like to design the DMRG constructor to be able to take the `trscheme` still, and have that routed through automatically, or error if both a trscheme and DMRG3S are provided. My experience is that while these "algorithm structs" are very convenient as a developer, it is somewhat harder as a newcomer to the package to keep track of everything, and a simple keyword-argument approach, which possibly does not expose all settings but handles the common cases, helps alleviate some of these issues.

src/algorithms/post_expand/dmrg3s.jl
VinceNeede: I already took a look at `AC_projection`, and from my understanding, it requires to supply all the enviroments, both left and right, here instead only one is required at a time. I'm not sure it can be adapted to this use case

lkdvos: yes, but the combiner can be thought of as an environment in this case, so while the code routing is a bit more complex, the contraction is the same as this one:

https://github.com/QuantumKitHub/MPSKit.jl/blob/5bbab08d1c96e5ba4cc43ba30487681e7e4d770b/src/algorithms/derivatives/mpo_derivatives.jl#L69-L70

src/algorithms/groundstate/dmrg.jl
VinceNeede: The reason I decided not to pass iter and instead pass alg_gauge is that alg_gauge is only updated once per sweep. Moving the update inside local_update! would mean calling it unnecessarily at every site. I could make the struct mutable to avoid the extra allocations, but there would still be the call overhead. Another option would be to have the scheduler update at each site instead of once per iteration — in the original formulation, it was actually supposed to monitor the change in energy before and after truncation and use that to adapt the noise.

Alternatively, we could update alg (the DMRG struct) in place each sweep instead of threading alg_gauge through as a separate argument — local_update! would then just read alg.alg_gauge directly, which avoids the extra parameter without needing to call the update per site. Either DMRG or DMRG3S would need to become mutable for this to work.

src/algorithms/post_expand/dmrg3s.jl
VinceNeede: I tried using `fuser`, but it is restricted to `IndexSpace` only, while `right_virtualspace(H[pos])` returns something like `BlockTensorKit.SumSpace{GradedSpace{ProductSector{Tuple{TensorKitSectors.PlanarTrivial, Z2Irrep}}, Tuple{Int64, Int64}}}`
Could we extend the method? or is the isomorphism not uniquely defined in other cases?

src/algorithms/groundstate/dmrg.jl
lkdvos: I would be very surprised if this ever shows up in a profiler though, since all these structs can be kept immutable, they should all end up being stack-allocated. As a result, creating a new one or changing out some settings is virtually free, and presumably takes order nanoseconds. Of course to be completely sure we can measure this, but I don't think this is ever dominant compared to the actual contractions.

For the same reason I'd want to try and keep these algorithms as non-mutable structs, because that basically means they don't have to be allocated and often creating a new algorithm is better than mutating an existing one.

src/algorithms/post_expand/dmrg3s.jl
lkdvos: Oh I forgot that this is restricted, feel free to indeed just extend it by removing that type restriction :)

src/algorithms/post_expand/dmrg3s.jl
VinceNeede: I have tried using `MPO_AC_Hamiltonian`, but the assumption of this line breaks:
https://github.com/QuantumKitHub/MPSKit.jl/blob/5bbab08d1c96e5ba4cc43ba30487681e7e4d770b/src/algorithms/derivatives/mpo_derivatives.jl#L71
For a proper environment, the resulting MPS does have a single block, but in the case of a combiner this breaks. For an hamiltonian of bond dimension 3, the right environment has blocks $1 \times 3 \times 1$ while the combiner has $1\times 3 \times 3$, which give `y` as a $1\times1\times3$ blocked.
I also tried `JordanMPO_AC_Hamiltonian`, but it requires indexing the right environment
https://github.com/QuantumKitHub/MPSKit.jl/blob/5bbab08d1c96e5ba4cc43ba30487681e7e4d770b/src/algorithms/derivatives/hamiltonian_derivatives.jl#L83
And this does not work for `combiner`, because getindex only works witth zero or 1 non trivial dimensions https://github.com/QuantumKitHub/BlockTensorKit.jl/blob/215f9283301cc4d964d5819754320d6c4a25a064/src/vectorspaces/sumspaceindices.jl#L63-L76

I don't think I'm in a good position to safely change either of these, so I'd rather get your take on how you'd want this handled before I try anything myself. Is there a way you'd recommend getting the raw, uncollapsed contraction out of this framework, or is this a case where the current abstraction just isn't meant to support it and I should go another route entirely?

src/algorithms/groundstate/dmrg.jl
VinceNeede: I'll try and run a benchmark for sanity check as soon as I have time

VinceNeede:
# Update
- `DMRG` constructor has been updated to accept `alg_gauge`. The inner algorithm used in the svd/ortho step is constructed as before, ensuring back compatibility. `trscheme` is initially defaulted to `nothing` to prevent supplying the svd/ortho algorithm twice, it is then repristinated to the previous default as `trscheme = something(trscheme, notrunc())`. The docstring has been updated to explain the behavior and `DMRG3S` role.
- `fuser` type restriction has been removed, and is now used to construct the combiner. The `@plansor` contraction has been reordered to mimic `MPO_AC_Hamiltonian`, even though it still can't be used directly for the above mentioned problems.
# TODO
- [x] Move `update_alg_gauge` inside `local_update!`, verifying no significative change in performance/allocations is present
- [x] Understand how `MPO_AC_Hamiltonian` can be used for the contraction step

src/algorithms/post_expand/dmrg3s.jl
lkdvos: Thanks for checking and explaining why it fails, I definitely also missed that.
Looking into this in a little more detail, I guess we probably don't want to get a 
$1 \times 1 \times 3$ object out anyways, since you want to concatenate it with the MPSTensor right afterwards.
In particular, it should be more efficient to immediately make the fuser that collapses these levels together. I would have to look at the exact implementation requirements, but I think you can call `TensorKit.oplus` on the fused spaces to combine them. (Unfortunately, that indeed now means we can no longer reuse the `fuser` function, but that's fine).

Let me know if this makes sense, I definitely might be missing something, and I'm happy to explain more of what I'm thinking of.
The idea would definitely be to try and reuse the `MPO_AC_Hamiltonian`, not the `JordanMPO` one.

lkdvos:
Thanks for the update! I commented on the reuse of the AC_projection, hopefully that makes sense, and I think it is safe to assume that the alg_gauge computation per site is not hurting performance, indeed after these two this looks ready to go for me.
Already thanks for putting in this effort, this is a very nice PR.

src/algorithms/post_expand/dmrg3s.jl
VinceNeede: Hi @lkdvos,

I was trying to use `oplus` to flatten a leg, but ran into a loss of precision in doing so (my second DMRG3S test, which checks proper subspace exploration, started failing). Tracked it down to `unitary`/`isomorphism` producing a non-unitary map when the domain is a `SumSpace` and the codomain is its flattened (`oplus(...)`) counterpart — even when every sector's dimension matches exactly between the two. Minimal example:

```julia
using TensorKit, BlockTensorKit, LinearAlgebra

V1 = Rep[U₁](0 => 2, 1 => 1)
V2 = Rep[U₁](0 => 2, 1 => 1)
Vsum = V1 ⊞ V2
Vflat = oplus(Vsum)

u = unitary(ComplexF64, Vflat, Vsum)
@show norm(u' * u - id(Vsum))          # 1.73..., should be ~0
@show norm(only(u * u') - id(Vflat))   # 1.73..., should be ~0
```

Is this a known limitation, or should I open a proper issue (and if so, here or on `TensorKit`)? Happy to give more detail / try a fix there.

Thanks!

lkdvos: This is definitely not a known limitation and almost certainly is a bug. Feel free to open an issue on BlockTensorKit.jl, where I assume this is originating from, and I'll try and have a look asap.

VinceNeede:
Both two last points have been implemented. I have explicitly update the compatibility of `BlockTensorKit` since it depends on https://github.com/QuantumKitHub/BlockTensorKit.jl/pull/64. I have marked this ready for review. 
Since the formatter is still not executing, I have run it locally:
```bash
vince@fedora:~/MPSKit.jl$ git log --oneline -1
660b14e (HEAD -> vb-dmrg3s, origin/vb-dmrg3s) add _get_combiner function and update gauge! to use MPO_AC_Hamiltonian for perturbation calculation
vince@fedora:~/MPSKit.jl$ git runic main
runic did not modify any files
```

src/algorithms/groundstate/dmrg.jl
```diff
@@ -87,8 +132,10 @@ function local_update!(
         fixedpoint(H_effective, ac_old, :SR, alg_eigsolve)
     end
 
+    alg_gauge = _update_alg_gauge(alg.alg_gauge, iter, ϵ_global)
+
     # 3. gauge
-    ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge!(ψ, site, direction, AC′, alg.alg_gauge; normalize = true)
+    ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge!(site, direction, ψ, O, envs, AC′, alg_gauge; normalize = true)
```
lkdvos:
```suggestion
    ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge!(ψ, site, direction, O, envs, AC′, alg_gauge; normalize = true)
```

Just adding this since I think everywhere else this is the argument order. Looking at it, I have to admit that it's not the most consistent with the surrounding code, so I'd be fine with switching it, as long as it is the same everywhere, feel free to pick one :)

src/algorithms/groundstate/dmrg.jl
```diff
@@ -137,15 +184,15 @@ function DMRG2(;
     alg_eigsolve′ = alg_eigsolve isa NamedTuple ?
         Defaults.alg_eigsolve(; adaptive = true, alg_eigsolve...) : alg_eigsolve
     # two-site DMRG always truncates the enlarged bond back down, so the gauge is a truncated SVD
-    alg_gauge = MatrixAlgebraKit.TruncatedAlgorithm(alg_svd, trscheme)
+    alg_gauge = NoExpand(MatrixAlgebraKit.TruncatedAlgorithm(alg_svd, trscheme))
```
lkdvos: Do we still require the `NoExpand` wrapper, or do you think we can simply interpret DMRG3s as expansion and all other algorithms as regular gauging?

src/algorithms/groundstate/dmrg.jl
```diff
@@ -162,9 +209,11 @@ function local_update!(
         (newA2center, info)
     end
 
+    alg_gauge = _update_alg_gauge(alg.alg_gauge, iter, ϵ_global)
+
     # 2. gauge: truncated SVD split back into single-site tensors and install;
     #           the discarded weight is the truncation error
-    ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge2!(ψ, pos, direction, newA2center, alg.alg_gauge; normalize = true)
+    ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge2!(pos, direction, ψ, O, envs, newA2center, alg_gauge; normalize = true)
```
lkdvos:
```suggestion
    ψ, ϵ_trunc = @timeit timeroutput "gauge" gauge2!(ψ, pos, direction, O, envs, newA2center, alg_gauge; normalize = true)
```

src/algorithms/post_expand/dmrg3s.jl
```diff
@@ -0,0 +1,165 @@
...
+    return iszero(noise) ? NoExpand(alg.alg_gauge) : DMRG3S(noise, alg.schedule, alg.alg_gauge)
+end
+
+# similar to `fuser` but the contracted leg is flattened
+function _get_combiner(::Type{TorA}, V1, V2) where {TorA}
+    Vprod = V1 ⊗ V2
+    return isomorphism(TorA, oplus(fuse(Vprod)), Vprod)
+end
+
+function gauge!(pos::Int, ::Val{:right}, ψ::AbstractFiniteMPS, H, envs, AC, alg::DMRG3S; normalize = true)
```
lkdvos:
```suggestion
function gauge!(ψ::AbstractFiniteMPS, pos::Int, ::Val{:right}, H, envs, AC, alg::DMRG3S; normalize = true)
```

src/algorithms/post_expand/dmrg3s.jl
```diff
@@ -0,0 +1,165 @@
...
+    B = _transpose_tail(ψ.AR[pos + 1])
+    AR = _transpose_front(catcodomain(B, zeros(T, Vpert ← domain(B))))
+
+    normalize && normalize!(C)
+    ψ.AC[pos] = (AL, C)
+    ψ.AC[pos + 1] = (C, AR)
+    return ψ, ϵ
+end
+
+function gauge!(pos::Int, ::Val{:left}, ψ::AbstractFiniteMPS, H, envs, AC, alg::DMRG3S; normalize = true)
```
lkdvos:
```suggestion
function gauge!(ψ::AbstractFiniteMPS, pos::Int, ::Val{:left}, H, envs, AC, alg::DMRG3S; normalize = true)
```

src/algorithms/post_expand/post_expand.jl
lkdvos, line 1: Same comment here, can we avoid this by simply using the `inner_gauge` algorithm directly?

src/algorithms/post_expand/dmrg3s.jl
```diff
@@ -0,0 +1,165 @@
...
+
+(s::Warmup)(noise, iter, ϵ) = iter ≤ s.iters ? noise : zero(noise)
+
+"""
+    DMRG3S(noise, schedule::NoiseSchedule)
+
+Gauge algorithm wrapper that, at every site update, injects a Hamiltonian-derived
+perturbation of the just-optimized tensor before gauging — the "strictly single-site DMRG with
+subspace expansion" scheme of Hubig, McCulloch, Schollwöck & Wolf,
+[Phys. Rev. B 91, 155115 (2015)](https://doi.org/10.1103/PhysRevB.91.155115). This lets
```
lkdvos: Would you mind adding this to the bibliography file, and citing in the same style as e.g. the VUMPS or QuasiparticleAnsatz docstrings?

lkdvos (commented):
I left some minor things to resolve, will merge afterwards:

- [x] the `gauge!` function is slightly inconsistent in its argument order, I think currently both `psi` as first and third argument exists, could you make that the same order everywhere?
- [x] Can we still simplify by getting rid of the `NoExpand` algorithm wrapper?

src/algorithms/groundstate/dmrg.jl
VinceNeede: Sure, I was actually copying `changebond` signature, but I actually prefer `\psi` first, in in-place methods I prefer to put the thing being modified as first argument, so I'll change this asap

src/algorithms/groundstate/dmrg.jl
VinceNeede: `NoExpand` is not **strictly** required, but I believe that without it `alg_gauge` could mean two different things, both the algorithm used during the gauge, and an expansion algorithm that also happen to shift the gauge. It could be eliminated and simply dispatch on `Any` by default, and specialize methods for `DMRG3S`. Of course this is more of a design/interface point, so I'll be happy to go with whichever you prefer

src/algorithms/post_expand/post_expand.jl
VinceNeede: GitHub is not showing me the line you are referring to, is this related to the `NoExpand` point?

lkdvos: yes, it is the same

src/algorithms/groundstate/dmrg.jl
lkdvos: I think my main motivation is to keep it slightly more backwards-similar-looking (not truly backwards compatible since this was a breaking change anyways), and to avoid having too many wrappers, so I have a slight preference for specializing on `DMRG3S` and keeping the rest `Any`.

lkdvos (approved):
Thanks for all the work, will merge as soon as tests pass!

VinceNeede:
Thanks for the review! I've implemented the last two points we discussed. I also ran the formatter locally since it's still blocked  here, let me know if the CI flags anything I missed, or if you'd like any further changes.

Really want to say thanks for the time and care you put into this review. I learned a lot about the internals of MPSKit going through it with you, and I appreciate you taking a somewhat unconventional PR from someone new to the codebase this seriously.

src/algorithms/post_expand/post_expand.jl
```diff
@@ -0,0 +1,9 @@
+_update_alg_gauge(alg, iter, ϵ) = alg
+
+set_alg_gauge(::Nothing, inner_gauge) = inner_gauge
+alg_gauge(alg) = alg
+
+gauge!(ψ::AbstractFiniteMPS, pos::Int, direction, H, envs, AC, alg_gauge; normalize::Bool = false) =
+    gauge!(ψ, pos, direction, AC, alg_gauge; normalize)
+gauge2!(ψ::AbstractFiniteMPS, pos::Int, direction, H, envs, AC2, alg_gauge; normalize::Bool = false) =
+    gauge2!(ψ, pos, direction, AC2, alg_gauge; normalize)
```
lkdvos, line 9: probably need to check if these methods are required anymore: line 4 is a duplicate (which is the CI failure), and the others seem to be remnants from the previous argument order, which should no longer be required :)

VinceNeede: I forgot to specialize `alg_gauge(::DMRG3S)`, that's why tests were failing. I think opting for this method and using methods dispatch is the more clear way, otherwise we could delete these methods but the `dmrg` main file would become less readable since if cases on `alg_gauge` type should be inserted. Not sure if I'm missing something trivial here.

lkdvos:
Hopefully now the format tests are fixed as well, will merge as soon as we get the green light from the tests :)

src/algorithms/post_expand/dmrg3s.jl
```diff
@@ -0,0 +1,169 @@
...
+"""
+struct ExponentialDecay{T <: Real} <: NoiseSchedule
+    decay_rate::T
+    threshold::T
+end
+ExponentialDecay(decay_rate, threshold) = ExponentialDecay(promote(decay_rate, threshold)...)
+ExponentialDecay(decay_rate; threshold = zero(decay_rate)) = ExponentialDecay(decay_rate, threshold)
+
+function (s::ExponentialDecay)(noise, iter, ϵ)
+    decayed = noise * s.decay_rate^iter
```
Jutho, line 51: I am a bit confused by this. It seems like `_update_alg_gauge` is called after every iteration, and then the `alg_gauge` is replaced with a new `DMRG3S` instance, with the updated noise. But in that case, don't you just want to multiply the noise with `s.decay_rate`, without the power. You will get the power automatically from the sequential calls after every iteration, unless I am misreading the code.

Jutho: Never mind, I see. `_update_alg_gauge` is always called on the initial `alg.alg_gauge`, not on the current `alg_gauge`, so this is correct.

VinceNeede: Yes, the naming (`alg_gauge`/`_update_alg_gauge`) probably isn't the clearest, open to suggestions if you have one. To confirm the mechanics: `_update_alg_gauge` is called from `local_update!` on every site, not once per outer iteration, and each call always derives from the original `alg.alg_gauge` stored on the struct — not from the decayed instance returned by the previous call. So `decay_rate^iter` recomputes the correct amplitude for the sweep from scratch each time; without the power, the noise would decay every site instead of every sweep, since there's no accumulated state being threaded through.

PR #467, opened 2026-07-20
Configurable backend and per-solve scratch allocator

lkdvos:
Every local update of an MPS algorithm contracts a handful of intermediate tensors that die before the next update.
This PR stops handing those to the garbage collector and serves them from a dedicated allocator instead, and while at it makes the contraction `backend` a setting on every algorithm.

## Two settings, in two different places

**`backend` is a field on the algorithm**, defaulting to `Defaults.backend()`.
It is stateless, and `TensorOperations.DefaultBackend` is a placeholder that resolves to an implementation from the types of the tensors involved, so it is already the right choice for a GPU-backed state.

**The allocator is not a field.** Which one is appropriate depends on two things the algorithm cannot know:

- *Where the tensors live.* Host memory can be served by an allocator that bypasses Julia's memory manager; a device-backed state needs its scratch space on the device. This is only known once there is a state in hand.
- *Whether it will be shared between tasks.* `BufferAllocator` is a bump buffer with a mutable offset, so it is only safe when a single task owns it. Concurrent work gets a `ManualAllocator` instead, which holds no state at all.

So `MPSKit.default_allocator(state, scheduler)` is asked once at the start of each solve, and the answer serves every local update of that solve.
This is what settles the concern from the first version of this PR: **the algorithms stay thread-safe, and reusing one across a parallel parameter sweep is fine** — there is no scratch space on the algorithm to share.
Anything MPSKit does not recognise falls back on `DefaultAllocator`, which allocates through the storage type itself and is therefore correct on any device.

Buffering can be turned off with `MPSKit.Defaults.set_buffering!(false)`, which trades the speed back for lower memory use.
It is a compile-time preference, so Julia has to be restarted.

## What it buys

Two-site DMRG ground state, spin-1 Heisenberg, `L = 100`, `χ = 128`, 5 sweeps, single Julia thread, MKL with 16 threads:

| metric (steady state) | `DefaultAllocator` | `BufferAllocator` | change |
|---|---:|---:|---:|
| wall time | 65.5 s | 50.4 s | **−23%** |
| GC time | 24.2 s (37%) | 10.3 s (20.5%) | **−57%** |
| allocated | 110.85 GiB | 39.85 GiB | **−64%** |
| GC pauses | ~311 | ~137 | −56% |
| ground-state energy | −138.940086126678 | −138.940086126678 | Δ = 0 |

Of the ~111 GiB a sweep churns through, ~71 GiB is recyclable effective-operator scratch.
The residual ~40 GiB is Krylov basis vectors, the gauge-step SVD, environment updates and result tensors, which this allocator does not manage and which is therefore the floor.

Smaller measurements from the same machine (`@allocated`, so read as ratios): finite TDVP 513 → 325 MiB, `BUG` 886 → 792 MiB, CBE-DMRG 1263 → 1170 MiB.
`BUG` gains least because much of its cost sits in the basis-augmentation step, which is a factorization rather than an effective-operator contraction.

## Scope

Wired: `DMRG`/`DMRG2`, `IDMRG`/`IDMRG2`, `TDVP`/`TDVP2`, `BUG`, `VUMPS`, `VOMPS`, `GradientGrassmann`, `DMRG3S`, the bond-expansion algorithms, `ChepigaAnsatz2`, `FiniteExcited`, `exact_diagonalization`, dynamical DMRG, and the `approximate`/`leading_boundary` variants of the above.
`calc_galerkin` takes them too, so the Galerkin error no longer bypasses the buffer.

Not wired, and left for a follow-up: `fidelity_susceptibility` and the quasiparticle excitations, neither of which has an algorithm object carrying a backend today.

## Notes for review

- The effective-operator structs (`MPODerivativeOperator`, the `JordanMPO_*` operators, `ProjectionDerivativeOperator`) now carry the backend and allocator their matvec uses. Threading the 37 `@plansor` calls in the Jordan operators is what makes any of this pay: before that, the plumbing was inert for `MPOHamiltonian` and the two allocators were bit-identical in allocations.
- The allocator is obtained at the *entry point* of a solve and threaded down. Obtaining it anywhere else silently loses the reuse — it went wrong three times during review (per IDMRG iteration, per site in `changebond!`, and not at all in `calc_galerkin`), each time without any test noticing, since the answers are unaffected.
- Steps with no effective-operator contraction (a plain `gauge!`, `RandExpand`'s `changebond!`) accept the allocator and drop it, so a sweep can pass it unconditionally.
- Test coverage lives next to the algorithms it belongs to rather than in a file of its own: the finite `DMRG`/`timestep` testsets check that reusing one algorithm object reproduces the answer, and the infinite `VUMPS`, `GradientGrassmann` and `TDVP` testsets are parametrized over the serial and dynamic schedulers, since that is what picks the allocator. `test/misc/allocator.jl` covers the selector itself, and `test/gpu/{cuda,amd}/allocator.jl` covers only what a real device can settle — that a device-backed state is never handed a host allocator.


lkdvos:
### Benchmark: `DefaultAllocator` vs `BufferAllocator`

End-to-end DMRG2 ground-state run measuring the effect of reusing a `BufferAllocator` across the sweep, driven by `benchmark/allocator_gc.sbatch` / `benchmark/allocator_gc_run.jl`.

**Setup:** spin-1 Heisenberg chain (`Trivial`), `L=100`, `χ=128`, 5 fixed sweeps, `krylovdim=20`, non-adaptive eigensolver (identical work for both allocators). Single Julia thread (so GC is single-threaded and its wall time is a clean number), MKL with 16 threads. `julia 1.12.6`.

| metric (steady-state) | `DefaultAllocator` | `BufferAllocator` | change |
|---|---:|---:|---:|
| wall time | 65.5 s | 50.4 s | **−23%** |
| GC time | 24.2 s (37%) | 10.3 s (20.5%) | **−57%** |
| allocated | 110.85 GiB | 39.85 GiB | **−64% (−71 GiB)** |
| GC pauses | ~311 | ~137 | −56% |
| full collections | 22 | 9 | −59% |
| ground-state energy | −138.940086126678 | −138.940086126678 | Δ = 0 |

Of the ~111 GiB a sweep churns through, **~71 GiB is recyclable effective-operator contraction scratch** — with `BufferAllocator` it now comes from the reused buffer instead of the GC heap. The residual ~40 GiB (Krylov basis vectors, the gauge-step SVD, environment updates, result tensors) is not managed by the operator allocator and is therefore identical for both — that's the floor.

<details>
<summary>Raw output</summary>

```
model=Heisenberg S=1  L=100  χ=128  sweeps=5  krylovdim=20  reps=3
julia=1.12.6  julia_threads=1  gc_threads=1  BLAS_threads=16
DefaultAllocator | rep 1/3 |    66.79 s | GC   25.48 s ( 38.1%) |   110.85 GiB |    320 pauses |  23 full
DefaultAllocator | rep 2/3 |    65.62 s | GC   24.24 s ( 36.9%) |   110.85 GiB |    310 pauses |  22 full
DefaultAllocator | rep 3/3 |    65.41 s | GC   24.14 s ( 36.9%) |   110.85 GiB |    313 pauses |  22 full
BufferAllocator  | rep 1/3 |    50.91 s | GC   10.34 s ( 20.3%) |    39.85 GiB |    135 pauses |   9 full
BufferAllocator  | rep 2/3 |    50.37 s | GC   10.33 s ( 20.5%) |    39.85 GiB |    137 pauses |   9 full
BufferAllocator  | rep 3/3 |    50.38 s | GC   10.32 s ( 20.5%) |    39.85 GiB |    137 pauses |   9 full
energy check: |Δ| = 0.00e+00
```
</details>

src/algorithms/derivatives/hamiltonian_derivatives.jl
```diff
@@ -401,35 +421,37 @@ function prepare_operator!!(
...
 
-    return JordanMPO_AC2_Hamiltonian{O1, O2, O3, O4′}(II, IC, ID, CB, CA, AB, AA, BE, DE, EE)
+    return JordanMPO_AC2_Hamiltonian{O1, O2, O3, O4′}(II, IC, ID, CB, CA, AB, AA, BE, DE, EE, backend, allocator)
 end
 
 # Actions
 # -------
 function (H::JordanMPO_AC_Hamiltonian)(x::MPSTensor)
+    backend, allocator = H.backend, H.allocator
     y = ismissing(H.A) ? zerovector(x) : H.A(x)
```
borisdevos, line 433: As far as I can tell, this continuing block's action doesn't go through an `@plansor` with backends or allocators passed (this particular line dispatches to line 77 of `mpo_derivatives.jl`). I think this will occur in every algorithm which calculates a Galerkin error, as there through the `projection operator * x` you end up here. This isn't an issue within the actual eigsolve calls which matter in e.g. DMRG, so I'm not sure how severe this is

src/algorithms/derivatives/hamiltonian_derivatives.jl
```diff
@@ -12,42 +12,52 @@ const _HAM_MPS_TYPES = Union{
 Efficient operator for representing the single-site derivative of a `MPOHamiltonian` sandwiched between two MPSs.
 In particular, this operator aims to make maximal use of the structure of the `MPOHamiltonian` to reduce the number of operations required to apply the operator to a tensor.
 """
-struct JordanMPO_AC_Hamiltonian{O1, O2, O3} <: DerivativeOperator
+struct JordanMPO_AC_Hamiltonian{O1, O2, O3, Bk, Al} <: DerivativeOperator
```
borisdevos: Genuine question, how much would compilation benefit from specifying `Bk` to be an `AbstractBackend` at this level already? And the same for the two-site version. This is done for the `PrecomputedDerivative` struct.

src/algorithms/groundstate/dmrg.jl
```diff
@@ -24,8 +24,12 @@ given `trscheme`).
 ## Fields
 
 $(TYPEDFIELDS)
+
+!!! warning
+    The default allocator uses a buffer that is created at algorithm construction time.
+    As a result, it is **not thread-safe** to reuse the algorithm across multiple concurrent solves.
```
leburgel: I think this is enough of a warning for people doing parameter sweeps, I don't think it's worth spending the time to try and circumvent the issue altogether. What seems more important to me is some explanation on how to use a different (potentially thread-safe) allocator for the people that want to do this.

While this is not really the place for that, when checking the docs I noticed that the keyword-argument constructor for the DMRG algorithm that people would actually use to instantiate an algorithms isn't actually documented anywhere. The same goes for the other algorithms like `VUMPS`. It would be really nice to have a docstring for the kwarg constructor, which even mentions the default values used for all the settings. Especially for the DMRG docstring, it would be good to have this before the whole discussion on expansion and gauging algorithms that is there now.

So, either here or in a follow-up, it would be great to have complete docstrings for the kwarg-based algorithms constructors which indicate the default settings already. This would also make it immediately visible that the default allocater is a buffer allocator, in addition to the warning added here.

src/algorithms/derivatives/hamiltonian_derivatives.jl
leburgel: I guess this should be automatically resolved once the `leading_boundary` implementations also get backend and allocator support.

leburgel (commented):
Looks great to me. Aside from the comments I am quite curious: do you have any idea of how this affects the compilation time?

src/algorithms/groundstate/dmrg.jl
lkdvos: I actually got started on some of that, as well as bringing all of it in a uniform format in #449, which then kind of got out of hand to be an enormous PR. I might try and consolidate some parts of this into separate PRs, and if not just push these updates through ASAP because I think our docs really could be improved dramatically by this.

src/algorithms/derivatives/hamiltonian_derivatives.jl
lkdvos: I don't think it actually matters, although you are right that it might be good to be consistent. Putting that in the type def is mostly just a type restriction, in the sense that it gives explicit errors if you were to put something in that field that is not an `AbstractBackend`, but I don't think the compiler can really use that information, since Julia tends to specialize anyways.

It might become relevant if we were to try and start despecializing some things, but even then I can't really tell if this would help or hurt the compilation time or runtime.

src/algorithms/derivatives/hamiltonian_derivatives.jl
lkdvos: Oh I think you are right, only the prepared ones get the allocator/backend support. I'll make sure to update this, although I would really like to simplify this mess of code at some point...

lkdvos (commented):
I'm not sure this affects compilation times that much, in the sense that of course we have to compile stuff multiple times for different backends and allocators, but since that was previously not possible this isn't really a regression.

I'm still playing around a bit with some compilation time stuff too though, for example I think the higher-level algorithm implementations don't actually need to be specialized which might recover some of this.

lkdvos:
Update here: I'm still working on what I want to do with the design here, and for now I'm thinking that this might be overengineered right now, so I'm going to try and slim this down slightly.

lkdvos:
Alright, update again here, I think this is now more or less settling on a design I'm happy with. I kept the backends selectable, being a hook for making changes, but actually decided against making the allocator a configurable thing, mostly because it is just too easy to accidentally do something wrong with that. I think it is fair to always take the bumper where possible, and this now also handles the GPU case, so the main thing left for me to do is clean up the tests and rebase. Also note that this now means that parameter sweeps in parallel would work perfectly fine, and would probably have way better performance because the GC needs to run less often.
I'll try to do some more before/after tests to actually see if this matters.

src/algorithms/groundstate/idmrg.jl
```diff
@@ -172,26 +161,30 @@ function localupdate_step!(
...
+        it.backend, allocator,
     )
 end
 
 function localupdate_step!(
         it::IterativeSolver{<:IDMRG2}, state
     )
     alg_eigsolve = adapt_solver(it.alg_eigsolve; iter = state.iter, g_global = state.ϵ)
+    # the sweep is serial, so a single allocator serves all local updates
+    allocator = default_allocator(state.mps, SerialScheduler())
```
borisdevos: Is there a reason why only in `DMRG(2)` the allocator is passed to the "per-site" function, while here (and in some other places) this is rebuilt every site?

src/algorithms/groundstate/dmrg.jl
```diff
@@ -118,15 +111,14 @@ function DMRG(;
...
 
-
 function local_update!(
         site, direction,
         ψ, O, alg::DMRG, envs,
         ϵ_global, ϵ_trunc, decay_rate,
-        iter, timeroutput
+        iter, timeroutput, allocator
     )
     ϵ_local = calc_galerkin(site, ψ, O, ψ, envs)
```
borisdevos: `calc_galerkin` still doesn't take backends/allocators, is this on purpose?

src/algorithms/derivatives/projection_derivatives.jl
```diff
@@ -25,7 +25,7 @@ end
 function AC2_hamiltonian(
         site::Int, below, operator::ProjectionOperator, above, envs;
         prepare::Bool = true,
-        backend::AbstractBackend = DefaultBackend(), allocator = BufferAllocator()
+        backend::AbstractBackend = DefaultBackend(), allocator = DefaultAllocator()
```
borisdevos, line 45: I can't get to the relevant lines from here, but the `h(x)` actions for these operators don't have allocators or backends passed along, only within this `AC(2)_Hamiltonian` call through `prepare_operator!!`. Lines 38-51 btw

src/algorithms/ED.jl
```diff
@@ -22,7 +22,6 @@ equivalent to dense eigenvectors.
 - `alg=Defaults.alg_eigsolve(; dynamic_tols=false)`: the diagonalization algorithm to use,
   see [`KrylovKit.eigsolve`](@extref).
 - `backend=DefaultBackend()`: backend for tensor contractions and index manipulations.
```
borisdevos: `Defaults.backend()` is currently `DefaultBackend()`, but that might change, so it's probably better to write it as the former.

src/algorithms/groundstate/idmrg.jl
lkdvos: This is actually not rebuilt every site (the primitive below here does a full sweep over all sites in the unitcell), but the main reason is just that I missed it. I think to actually correctly fix this I would prefer to try and rewrite the IDMRG implementation in the same shape as the DMRG one, but it is getting hard to either make PRs orthogonal and not too large and to decouple these kinds of decisions.

Thanks for noting this in any case, I'll try and "fix" this in a better way.

lkdvos:
I think this is now actually in a merge-able state. I'm kind of happy with taking the allocators out of the hands of configuration, and it seems like it doesn't complicate the code too much, so I think this is acceptable

borisdevos (approved):
Looks good! As of now the docs preview isn't showing the two page changes, but it should build correctly when merging. I look forward to killing all my laptop's memory 😄

PR #468, opened 2026-07-20
feat: BUG integrator

lkdvos:
This PR adds the BUG integrator as an alternative to TDVP. It's a consolidation of part of #454.
The main purpose is to have a stable integrator to be able to handle (large) imaginary or complex time steps in a way that is more stable than standard TDVP since it does not involve a backwards integration.

@Jutho I would be very happy to hear your opinions on this, I'm still not completely sure I've done everything correctly and many parts still confuse me, so a second set of eyes would be greatly appreciated. 🙃


docs/src/changelog.md
```diff
@@ -21,8 +21,19 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 ### Added
 
+- `BUG` time-evolution algorithm: a symmetric second-order Basis-Update & Galerkin integrator for
+  finite MPS. Unlike `TDVP` it has no backward-in-time substep (stable for imaginary-time evolution),
```
Jutho: Christian Lubhich would of course not approve of this description, as TDVP is the differential equation / framework and BUG is an alternative integrator to the projector-splitting integrator. But yes, names and conventions are hard.

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,190 @@
+"""
+$(TYPEDEF)
+
+Single-site, symmetric second-order time-evolution algorithm for **finite** MPS, based on the
+Basis-Update & Galerkin (BUG) integrator, an unconventional robust integrator for dynamical low-rank
+approximation.
+
+Unlike [`TDVP`](@ref), BUG advances both the basis (K-step) and the core (Galerkin C-step) tensors
+*forward* in time, with no backward-in-time substep. This makes it a natural choice for
+imaginary-time / dissipative evolution, where the backward core step of TDVP can become unstable.
```
Jutho: Definitely here it might be useful to say something lie

"where the backward core step of the conventional projector-splitting TDVP integrator can become unstable."

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,190 @@
...
+always kept and only newly-appended directions are cut. The bond therefore *grows* to track the
+entanglement under a `truncerror` tolerance; a hard rank cap (`truncrank`) can additionally shrink it.
+
+!!! warning
+    With `trscheme = notrunc()` the augmented basis is kept at full rank every half-sweep, so the
+    bond grows unboundedly (up to the local Hilbert-space dimension) — this is **not** a fixed-rank
+    integrator. To cap the bond dimension at `D`, pass `trscheme = truncrank(D)`: because the old
+    directions carry singular value `≥ 1`, a rank-`D` cut keeps them and drops the newly appended
+    directions, so a bond already at `D` stays there (a fixed-rank-`D` step). A bond currently below
+    `D` can still grow up to `D` as new directions are admitted.
```
Jutho: Do you want to say that, with `truncrank(D)` where `D` is the current dimension, there is no evolution at all? I am probably misinterpreting this.

src/algorithms/timestep/tdvp.jl
```diff
@@ -11,10 +11,11 @@ the enlarged bond back down (selecting the truncated-SVD gauge). The expansion i
 state-preserving, as required for a consistent time evolution.
 
 !!! note
-    Real-time evolution preserves the norm: neither the bond expansion nor the truncation
-    renormalizes, so the state norm reflects the accumulated truncation error. Imaginary-time
-    evolution instead renormalizes at every step, like a ground-state search. CBE is only
-    available for finite MPS.
+    By default the norm is preserved: neither the bond expansion nor the truncation
```
Jutho: Typo I think?

```suggestion
    By default the norm is not preserved: neither the bond expansion nor the truncation
```

src/algorithms/timestep/tdvp.jl
```diff
@@ -163,9 +167,9 @@ function timestep!(
             imaginary_evolution
         )
```
Jutho, line 172: I am a bit confused by the handling of the right most tensor, and the structure of this second loop. You evolve the rightmost tensor for a time step `dt/2` in between the two loops, and then once more for a time step `dt/2` in this second loop, but with a possible `changebond!` call in between. What does that do for `i == length(ψ)`? 


This code structure is definitely somewhat different then what I am used to, namely

1) left-to-right-sweep from i=1 to L-1
   * evolve Ac[i] for dt/2
   * split updated Ac[i] into AL[i] and some C
   * evolve this C on bond i:i+1 backward
   * absorb this backward evolved C in site to form old Ac[i+1]
2) evolve Ac[L] for time dt
3) right-to-left-sweep for i=L-1 to 1
   * split Ac[i+1] into C and AR
   * evolve this C on bond i:i+1 backward for time dt/2
   * absorb this C with AL[i] to form old Ac[i]
   * evolve this Ac[i] forward for dt/2
 
I see how it most likely ends up being the same, but is there some overhead in evolving (and constructing AC_hamiltonian) for site Ac[L] twice, instead of just once? And what does `changebond!` for `i==L` do? Does it just return immediately?

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,190 @@
...
+        imaginary_evolution::Bool = false, normalize::Bool = false
+    )
+    # symmetric 2nd-order: a dt/2 left→right half-sweep composed with its dt/2 mirror
+    L = length(ψ)
+    h = dt / 2
+
+    # left→right half-sweep (root = last site): freeze the bases as `[AC[1], AR[2], …, AR[L]]` and
+    # carry `AC_old`/`transport` forward
+    ARs = ψ.AR[2:end]
+    pushfirst!(ARs, ψ.AC[1])
```
Jutho: What happens if you just ask for `ARs = ψ.AR[1:end]` ? For a normalized MPS, that should be the same, but I don't know how a non-normalized state is handled in that case?

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,190 @@
...
+    transport = isomorphism(scalartype(ψ), left_virtualspace(ψ, 1) ← left_virtualspace(ψ, 1))
+    AC_old = ARs[1]
```
Jutho: `isomorphism` called this way doesn't know about the storage type, and I think you really want to start it off as an identity (which is what isomorphism produces, but still, code clarity and such):

```suggestion
    AC_old = ARs[1]
    transport = id!(similar(AC_old, left_virtualspace(ψ, 1) ← left_virtualspace(ψ, 1)))
```

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,190 @@
...
+# copying version
+function timestep(
+        ψ::AbstractFiniteMPS, H, time::Number, timestep::Number,
+        alg::BUG, envs::AbstractMPSEnvironments...;
+        imaginary_evolution::Bool = false, normalize::Bool = false, kwargs...
+    )
+    isreal = (scalartype(ψ) <: Real && !imaginary_evolution)
+    ψ′ = isreal ? complex(ψ) : copy(ψ)
+    if length(envs) != 0 && isreal
+        @warn "Currently cannot reuse real environments for complex evolution"
```
Jutho, line 196: I assume calling `complex` is not supported on the environments?

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,190 @@
...
+    return ψ, envs
+end
+
+# copying version
+function timestep(
+        ψ::AbstractFiniteMPS, H, time::Number, timestep::Number,
+        alg::BUG, envs::AbstractMPSEnvironments...;
+        imaginary_evolution::Bool = false, normalize::Bool = false, kwargs...
+    )
+    isreal = (scalartype(ψ) <: Real && !imaginary_evolution)
```
Jutho: Does `scalartype(H)` work? 
```suggestion
    isreal = (scalartype(ψ) <: Real && !(imaginary_evolution && scalartype(H) <: Real))
```

src/algorithms/timestep/bug.jl
lkdvos: That doesn't really work, because the normalization would be in `C[0]`, and additionally that breaks the `WindowMPS` support, which should now actually just work

src/algorithms/timestep/bug.jl
lkdvos: Not really, and it is somewhat of a pain to define it, because the bookkeeping of the environments does weird things with that. Looking at this a bit more, (we have the same pattern for TDVP), the environments will probably be recomputed anyways if we switch to complex, since the object identity of the tensors is what is used for the environments to know if they have to be recomputed. I'll open an issue for this to tackle this separately.

src/algorithms/timestep/tdvp.jl
lkdvos: The `changebond!` is an expansion step, so in the `:left` direction it would expand `AR[L] -> [AR[L]; VR*X]` and `AL[L-1] -> [AL[L-1] 0]`.

I do agree that evolving the last site twice is probably not what we want, and has some needless overhead. I'll look into changing that (in a separate PR as well though).

src/algorithms/timestep/bug.jl
lkdvos: The thing I'm trying to warn for is that there are two modes of defining `trscheme` in this package, for a variety of reasons. In the expansion schemes, most of the time the `trscheme` is counted with respect to the expanded directions, so `truncrank(4)` would mean "add at most 4 to the current D", and not "make the current D at most 4". On the other hand, the twosite schemes really have `truncrank(D)` to mean "make the current D at most 4", so I wanted to kind of distinguish between that? I'll try and rephrase though, I agree this is not the greatest way of making this clear.

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,190 @@
+"""
+$(TYPEDEF)
+
+Single-site, symmetric second-order time-evolution algorithm for **finite** MPS, based on the
+Basis-Update & Galerkin (BUG) integrator, an unconventional robust integrator for dynamical low-rank
+approximation.
```
Jutho: What is symmetric? You mean time-reversal symmetric in the same was as the projector splitting integrator? That is definitely not true. I am also quite sure that it is not a second order integrator, at least not the one that is described in the reference you mention or its extension to TTN in https://arxiv.org/pdf/2201.10291, or the parallel one in https://arxiv.org/pdf/2304.05660. They have a second order scheme in https://arxiv.org/abs/2402.08607, but that requires to grow the bond dimension by a factor of 4.

lkdvos: I'm slightly confused because I think I actually measured this.
I definitely agree that I'm cutting some corners in the implementation, but I thought the forwards + backwards sweep here actually is equivalent to the midpoint scheme in the last reference you mention (and what I'm referring to with symmetric).
The bond dimension growth is indeed a factor 4 in that case, (doubling once for each sweep) except that I'm allowing intermediate truncations between the two sweeps to happen, (these are the corners I'm cutting), so in that sense if the truncation is too severe the second-order properties are lost?

I will definitely concede that this isn't rigorous at all, and I'll rerun some tests, but I really recall getting second-order results out of this (at least from a practical side)

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,190 @@
...
+        return ψ, transport, AC_old
+    end
+
+    # 2. move gauge right
+    oldbasis, C₀ = left_gauge(AC₀, alg.alg_orth)                        # old AL - C in new frame
+    AC_next = _mul_front(C₀, ARs[site + 1])
+
+
+    # 3. reproject onto new basis
+    newbasis, _ = left_gauge(catdomain(oldbasis, AC), alg.alg_gauge)    # augmented basis
```
Jutho: Can `alg.alg_gauge` truncate here? I don't think that is something you want to do like this, nor that this is wat was intended in their papers. This would be quite bad, since if you truncate to the old dimension, I think this will just lead to `newbasis = oldbasis`. Indeed, `oldbasis` has orthonormal columns, whereas the extra columns are those of a center site, so they have singular values attached and are furthermore not orthogonal to the old basis. This is not very rigorous, but essentially I think the old basis will have a weight that is strictly larger than one, whereas whatever new information is in the additional columns has a weight strictly smaller than one.

Jutho: I thought I understood how the algorithm works, but I am still confused. However, I don't think that what you are doing here is correct.

lkdvos: Would it be fair to say that I can't interleave truncation and evolution, because in reality I am truncating while the right- or left- half of the schmidt basis is still "at the wrong time", and I should therefore first evolve everything, then truncate, which will regauge everything?

The reason I wanted to avoid this is that in that case the gauge center effectively ends up at the starting side, so the whole symmetric back-and-forth looks like it requires an additional sweep of gauging.

Looking at this a bit more though, it might be that I could get away with this simply by changing the order:
1. truncate the old AC
2. transport that to the new frame and evolve
3. move gauge
4. reproject onto new basis.

Let's discuss this though, to see if I didn't miss anything here

docs/src/changelog.md
lkdvos: I'll keep it a bit more brief in the changelog, since here this refers to code structs, and make sure to update the doc(strings) more with projector-splitting language :)

lkdvos:
Alright, after a round of review and additional discussions, (thanks @Jutho for the insights!), I think I've now addressed some of the issues that were here and added some more convenience parts to the implementation.

The first major thing is that I now no longer truncate the augmented basis, which is indeed incorrect. From the small tests I ran this definitely did improve the accuracy, so this was definitely a needed step.

To consider where the truncation actually has to happen, I agree that the original formulation of the algorithm would mean to first evolve left-to-right, and then truncate right-to-left. Since I kind of wanted to keep some of the left-right symmetry here, I made a small alteration that I think is both more accurate as well as more efficient, but I would be happy to get a second opinion on this. The train of thought goes something like this: imagine that on the reverse pass, instead of first truncating right to left, then regauging to the right, and then starting the new evolution run from right-to-left again, I truncate site L-1, and then immediately start the evolution, as if no truncation happened `1:L-2`. This means I don't have to regauge back up the chain, and additionally the evolution sees a less truncated state. Then I do the same for site `L-2`, all the way down to `1`.
I'm pretty sure this change in order keeps everything consistent, of course up to truncation errors but it seems at least sound in my head?

I also added some timeroutput support to be able to track where the time is actually spent (almost entirely in the exponentiator, of course), but this can be useful for debugging.

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,209 @@
+"""
+$(TYPEDEF)
+
+Single-site, symmetric second-order time-evolution algorithm for **finite** MPS, based on the
+Basis-Update & Galerkin (BUG) integrator, an unconventional robust integrator for dynamical low-rank
+approximation.
+
+Unlike [`TDVP`](@ref), BUG advances both the basis (K-step) and the core (Galerkin C-step) tensors *forward* in time, with no backward-in-time substep.
+This makes it a more natural choice for imaginary-time (dissipative) evolution, where the backward core step of the conventional projector-splitting [`TDVP`](@ref) can become unstable for large timesteps.
```
Jutho:
```suggestion
This makes it a more natural choice for imaginary-time (dissipative) evolution, where the backward core step of the conventional projector-splitting integrator [`TDVP`](@ref) can become unstable for large timesteps.
```

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,209 @@
+"""
+$(TYPEDEF)
+
+Single-site, symmetric second-order time-evolution algorithm for **finite** MPS, based on the
+Basis-Update & Galerkin (BUG) integrator, an unconventional robust integrator for dynamical low-rank
+approximation.
```
Jutho:
```suggestion
Single-site time-evolution algorithm for **finite** MPS, based on the
Basis-Update & Galerkin (BUG) integrator, an unconventional robust integrator for dynamical low-rank approximation with an observed second-order convergence.
```

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,209 @@
...
+        AL_old, C₀ = left_orth(AC₀; alg)
+        return AL_old, C₀, AR_next, zero(real(scalartype(ψ)))
+    end
+
+    # `Vᴴ` goes into the neighbour rather than into `C₀`, so the local update happens in the truncated
+    # bond space; in-place, as the installs below invalidate the `AC` cache anyway
+    AL_old, C₀, Vᴴ, ϵ = svd_trunc!(AC₀, alg)
+    AR_next = _mul_front(Vᴴ, AR_next)                    # old right block, rotated into the cut bond
+    ψ.AC[site] = (AL_old, C₀)
+    ψ.AC[site + 1] = (C₀, AR_next)
```
Jutho, line 79: I don't fully understand this paradigm. Why put back `C₀` twice?

src/algorithms/timestep/bug.jl
```diff
@@ -0,0 +1,209 @@
...
+    )
+end
+
+function timestep!(
+        ψ::AbstractFiniteMPS, H, t::Number, dt::Number, alg::BUG,
+        envs::AbstractMPSEnvironments = environments(ψ, H, ψ);
+        imaginary_evolution::Bool = false, normalize::Bool = false,
+        timeroutput::TimerOutput = DISABLED_TIMER
+    )
+    # symmetric 2nd-order: a dt/2 left→right half-sweep composed with its dt/2 mirror
```
Jutho: I still don't think it is symmetric in the ODE sense. It is not just about going from left to right, and then going from right to left and doing exactly the same operation. Instead, you want to do the inverse operation with a negative time step.

src/states/orthoview.jl
```diff
@@ -182,6 +182,78 @@ function Base.setindex!(
...
+update: `set_AL_AC!` writes the left-isometric `AL` at `site` and the center tensor `AC` at
+`site + 1`, `set_AC_AR!` writes the right-isometric `AR` at `site` and the center tensor `AC` at
+`site - 1`. The gauge center therefore ends up at `site + 1` and `site - 1` respectively.
+
+These are the pendants of `ψ.AC[site] = (AL, C)` / `ψ.AC[site] = (C, AR)` for algorithms that
+already know the *next* center tensor: they avoid materializing a bond tensor purely to keep `ψ`
+well-defined, and — unlike installing the canonical tensor on its own — they never leave `ψ`
+without a gauge center.
+
+The isometry of `AL`/`AR` is not verified.
```
Jutho:
```suggestion
The isometric nature of `AL`/`AR` is not verified.
```

Or "isometricity"? 🙃

src/algorithms/timestep/tdvp.jl
```diff
@@ -120,23 +124,23 @@ end
...
-        imaginary_evolution::Bool = false
+        imaginary_evolution::Bool = false, normalize::Bool = false
     )
 
     # sweep left to right
     for i in 1:(length(ψ) - 1)
         # 1. optionally expand the bond ahead of the local update (CBE)
         isnothing(alg.alg_expand) ||
-            changebond!(i, Val(:right), ψ, H, alg.alg_expand, envs; normalize = imaginary_evolution)
+            changebond!(i, Val(:right), ψ, H, alg.alg_expand, envs; normalize)
```
Jutho, line 138: It seems like there is some possibility of unifying the building blocks between TDVP and BUG. Maybe this is very similar to `_cut_bond!`?  And `_evolve_center!` might also be used here? However, feel free to keep this to another PR and/or contributor 😄 .

Jutho:
Ok, I think this looks fine, up to a few suggestions. I could probably go over it once more in even more detail, but I don't have time for this right now, so feel free to merge.

src/algorithms/timestep/bug.jl
lkdvos: This is just a quirk of the system, we only have setters for `AC` and I need this signature to avoid doing any of the automatic/correct poisoning of the left/right gauged tensors. It shouldn't actually be doing anything.

src/algorithms/timestep/tdvp.jl
lkdvos: Good point, although I will indeed keep that for a separate PR 😉

lkdvos:
Will merge this as soon as tests pass :)

PR #475, opened 2026-07-25
Refactor MPSTensor construction to fix type piracy

### LINKED ISSUE #471: `zeros` initialization of `MPSTensor` not working
According to this constructor:
https://github.com/QuantumKitHub/MPSKit.jl/blob/547c9130e8f191b32a14feeffc70f8ca739bd686/src/states/abstractmps.jl#L41-L54
It should be possible to construct an `MPSTensor` filled with `zero`, but that is currently not working, since the method `zeros!(:::AbstractTensorMap)` does not appear to be defined

```julia
julia> MPSKit.MPSTensor(zeros, Float64, 2, 10, 10)
ERROR: UndefVarError: `zeros!` not defined in `MPSKit`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] (AbstractTensorMap{T, S, 2, 1} where {S, T})(f::Function, T::Type, P::ComplexSpace, Vₗ::ComplexSpace, Vᵣ::ComplexSpace)
   @ MPSKit ~/.julia/packages/MPSKit/NXsUL/src/states/abstractmps.jl:50
 [2] (AbstractTensorMap{T, S, 2, 1} where {S, T})(f::Function, T::Type, d::Int64, Dₗ::Int64, Dᵣ::Int64)
   @ MPSKit ~/.julia/packages/MPSKit/NXsUL/src/states/abstractmps.jl:74
 [3] top-level scope
   @ REPL[12]:1

help?> TensorKit.zeros!
  No documentation found.

  Binding TensorKit.zeros! does not exist.
```
I've opened the issue here, but this could be very well up streamed to `TensorKit` to define the `zeros!` method

VinceNeede:
Closes #471.

Not really my place to be doing bigger refactors yet (this is only my second PR, after my
DMRG3S implementation), but the discussion in #471 pointed pretty directly at a concrete fix,
so I wanted to give it a try rather than just patching `zeros!` → `zerovector!` and leaving
the type piracy in place.

## What this does

- Removes the old `MPSTensor(f, T, P, Vₗ, Vᵣ)` / `MPSTensor(::UndefInitializer, ...)` /
  `MPSTensor(d, Dₗ, Dᵣ)` constructors, rather than just fixing the `zeros!` call inside them.
  As @lkdvos noted, these were type piracy (`MPSTensor` is a `const` alias for
  `AbstractTensorMap`, not a type MPSKit owns), and the original issue was really just a
  symptom of that.
- Introduces a small `MPSMapSpace{S, Sₚ}` struct that holds `(Vₗ, P, Vᵣ)`, so the
  `Vₗ ⊗ P ← Vᵣ` leg convention lives in exactly one place instead of being repeated in every
  constructor/accessor.
- Extends `Base.rand`/`Base.randn`/`Base.zeros` for `MPSMapSpace`, giving essentially the
  `zeros(T, Vl * P, Vr)`-style interface you mentioned as a preferred option, e.g.:

  ```julia
  julia> zeros(Float64, MPSMapSpace(ℂ^10, ℂ^2, ℂ^10))
  ```

  These are added as genuine `Base.rand`/`Base.randn`/`Base.zeros` methods (not a new type
  alias constructor), so there's no piracy here — `MPSMapSpace` is a type this package owns.

## What this does *not* do

I left out a dedicated `mpstensor(...)` convenience function for now. `MPSTensor` was never
exported, and grepping both MPSKit (src, test, docs, examples) and PEPSKit.jl for
`MPSTensor(`/`MPSTensor.(` turns up zero call sites outside the old constructor definitions
themselves, so I don't think there's a hidden dependency on the old constructor shapes. Given
that, it felt safer to start minimal and let you decide whether a convenience wrapper on top
of `MPSMapSpace` is actually wanted, rather than guessing at a signature nobody's using yet.

Happy to add `mpstensor` (or rename `MPSMapSpace` to something else, if `Vl * P` composite
notation is what you'd rather standardize on) if that's preferred — this is very much open
to being redirected.

### DIFF
--- a/src/states/finitemps.jl
+++ b/src/states/finitemps.jl
@@ -258,7 +258,7 @@ function FiniteMPS(
     end
 
     # construct MPS
-    tensors = @. f(elt, MPSMapSpace(Vspaces[1:(end - 1)], Pspaces, Vspaces[2:end]))
+    tensors = @. f(elt, Vspaces[1:(end - 1)] ⊗ Pspaces ← Vspaces[2:end])
     return FiniteMPS(tensors; normalize, overwrite = true)
 end
 function FiniteMPS(
--- a/src/states/finitemps.jl
+++ b/src/states/finitemps.jl
@@ -258,7 +258,7 @@ function FiniteMPS(
     end
 
     # construct MPS
-    tensors = MPSTensor.(f, elt, Pspaces, Vspaces[1:(end - 1)], Vspaces[2:end])
+    tensors = @. f(elt, MPSMapSpace(Vspaces[1:(end - 1)], Pspaces, Vspaces[2:end]))
     return FiniteMPS(tensors; normalize, overwrite = true)
 end
 function FiniteMPS(
--- a/src/states/abstractmps.jl
+++ b/src/states/abstractmps.jl
@@ -13,77 +13,22 @@ const MPSBondTensor{S} = AbstractTensorMap{T, S, 1, 1} where {T}
 const GenericMPSTensor{S, N} = AbstractTensorMap{T, S, N, 1} where {T} # some functions are also defined for "general mps tensors" (used in peps code)
 const MPSTensor{S} = GenericMPSTensor{S, 2} # the usual mps tensors on which we work
 
-"""
-    MPSTensor([f, eltype], d::Int, left_D::Int, [right_D]::Int])
-    MPSTensor([f, eltype], physicalspace::Union{S,CompositeSpace{S}}, 
-              left_virtualspace::S, [right_virtualspace]::S) where {S<:ElementarySpace}
-
-Construct an `MPSTensor` with given physical and virtual spaces.
-
-### Arguments
-- `f::Function=rand`: initializer function for tensor data
-- `eltype::Type{<:Number}=ComplexF64`: scalar type of tensors
-
-- `physicalspace::Union{S,CompositeSpace{S}}`: physical space
-- `left_virtualspace::S`: left virtual space
-- `right_virtualspace::S`: right virtual space, defaults to equal left
-
-- `d::Int`: physical dimension
-- `left_D::Int`: left virtual dimension
-- `right_D::Int`: right virtual dimension
-"""
-function MPSTensor(
-        ::UndefInitializer, T, P::Union{S, CompositeSpace{S}}, Vₗ::S, Vᵣ::S = Vₗ
-    ) where {S <: ElementarySpace}
-    TT = tensormaptype(S, 1 + (P isa S ? 1 : length(P)), 1, T)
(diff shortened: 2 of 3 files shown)


lkdvos (commented):
Thanks for getting started with this!

I think my main question here is about whether it makes sense to have `MPSMapSpace` as a dedicated struct, or we could just use `TensorMapSpace{S,N,1}` directly and get rid of this functionality.
I understand that it can be convenient to not have to think too much about the exact shape of the tensors, but on the other hand it is also useful to be more upfront about the underlying structure of these tensors, and it is slightly less code to maintain.

I'd be curious to hear what other people think too, @VictorVanthilt, @borisdevos, @leburgel ?

src/states/finitemps.jl
```diff
@@ -258,7 +258,7 @@ function FiniteMPS(
     end
 
     # construct MPS
-    tensors = MPSTensor.(f, elt, Pspaces, Vspaces[1:(end - 1)], Vspaces[2:end])
+    tensors = @. f(elt, MPSMapSpace(Vspaces[1:(end - 1)], Pspaces, Vspaces[2:end]))
```
VinceNeede:
```suggestion
    tensors = @. f(elt, Vspaces[1:(end - 1)] ⊗ Pspaces ← Vspaces[2:end])
```
@lkdvos do you mean something like this directly? This was my original idea, but I didn't love hardcoding this specific `TensorMapSpace` construction at every call site — I thought a centralized struct would make the convention easier to maintain in one place, which is why I went with `MPSMapSpace` instead.

That said, I don't have a strong opinion — happy to go with whichever you and the others prefer.

lkdvos: Yeah exactly. I agree it is nice to have a centralized part for this, but the struct really is just a `TensorMapSpace` which has to reimplement some of the logic that already is there, so there might be an argument for the fact that `TensorMapSpace` already **is** this centralized convention and we don't need to reinvent that here.

VictorVanthilt:
Just using `TensorMapSpace` makes more sense for me, I don't see what the `MPSMapSpace` adds. In a sense its a piracy in disguise of `TensorMapSpace` which is what we were trying to avoid in the first place.

VinceNeede:
Thank you both for the input, @lkdvos and @VictorVanthilt!

My idea was to keep the mapping `(Vl, P, Vr) -> Vl ⊗ P ← Vr` in a single place, but I agree it isn't really needed — it's already implicit elsewhere (e.g. the `MPSTensor` definition itself as a 2 <- 1 map), and given that `TensorMapSpace` already provides this centralized convention, reimplementing part of it in `MPSMapSpace` doesn't add enough to be worth it. I'll go ahead and switch to using `TensorMapSpace` directly.

lkdvos (approved):

PR #479, opened 2026-07-27
feat: standalone `linsolve` primitives

lkdvos:
This PR adds dedicated routines for `linsolve`, i.e. solving `Ax = b` with MPO/MPS.
This was already partially implemented in the `propagator` and dynamical DMRG stuff, but the goal is to make it a bit more of a first-class citizen and not hardcode the specific linear problems to solve.
Can also be used for time-stepping purposes, doing things like trapezoid rules.

There definitely are some design choices that I'm happy to have discussions about, considering names etc, but one particular choice I made here is to **not** overload KrylovKit's version.
The reasoning behind this is that the `envs` were a bit awkward to get in there anyways, and KrylovKit does not have an in-place version to begin with. It does have the annoying side-effect of nameclashes, since I definitely want to export this, so maybe a different name could be useful?

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -32,6 +32,16 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
   Unlike `TDVP` it has no backward-in-time substep (stable for imaginary-time evolution),
   and passing a truncating `trunc` enables rank-adaptivity (the bond dimension grows and shrinks
   automatically to track entanglement).
+- `linsolve`/`linsolve!` solve the linear system `(a₀ + a₁·A)·x = b` for a finite MPS `x`, with `A`
+  an MPO/MPOHamiltonian and `b` an MPS.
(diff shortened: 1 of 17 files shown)


Jutho:
It might be good (or not good because to broad) to consider the names of all the high level methods, and try to make them more consistent and streamlined. This would of course be massively breaking, but on the other hand, renaming a method is probably the easiest type of breaking change in terms of deprecations, and the old name can stay around for a very very long time. So what do we currently have:

* `find_groundstate`
* `excitations`
* `leading_boundary`
* `linsolve`
* `exact_diagonalization`
* `propagator`
* `approximate`

and maybe

* `changebonds`

The current naming is certainly all over the place 😄 .

Jutho:
That being said, I do think the current situation is already much nicer than some of the other tensor network packages, which often use the specific algorithm as a method name, and certainly also have a wild variety of names around.

PR #484, opened 2026-07-28
fix: FiniteMPS `+`/`-` mixes gauges when the lazy cache is re-derived (#473)

### LINKED ISSUE #473: FiniteMPS +/- fails to cancel parallel operands from different tensor networks (returns wrong state)
## Summary

`Base.:-` / `Base.:+` for `FiniteMPS` returns a **wrong state** when the two operands are (nearly) proportional vectors that are represented by *different tensor networks* (different gauge / bond bases).
Instead of cancelling, the result comes out with roughly the norm of `a + b`.

The simplest way to hit it: `gs` is an eigenstate of `H`, so `H * gs == E0 * gs`, hence `E0*gs - H*gs` should be the zero vector.
Instead `norm(E0*gs - H*gs)` returns `≈ 2*norm(gs)*E0`.

I ran into this while implementing an MPS linear solver, where `z*x - H*x` (with `x ≈ c*gs`) blew up to roughly `|z*x| + |H*x|` instead of cancelling.

## Minimal reproducer (public API only)

```julia
using MPSKit, TensorKit

# a tiny transverse-field Ising Hamiltonian, built from Pauli matrices
function tfim(L; g = 1.0)
    X = TensorMap(ComplexF64[0 1; 1 0], ℂ^2 ← ℂ^2)
    Z = TensorMap(ComplexF64[1 0; 0 -1], ℂ^2 ← ℂ^2)
    @tensor ZZ[-1 -2; -3 -4] := Z[-1; -3] * Z[-2; -4]
    lattice = fill(ℂ^2, L)
    return FiniteMPOHamiltonian(lattice, (i, i + 1) => -ZZ for i in 1:(L - 1)) +
        FiniteMPOHamiltonian(lattice, (i,) => -g * X for i in 1:L)
end

L = 6
H = tfim(L)
gs, envs, = find_groundstate(FiniteMPS(rand, ComplexF64, L, ℂ^2, ℂ^16), H, DMRG(; verbosity = 0))
E0 = real(expectation_value(gs, H, envs))

a = E0 * gs         # E0 * gs
b = H * gs          # H  * gs  ==  E0 * gs   (gs is an eigenstate of H)
c = a - b           # must be the zero vector

mps_norm  = norm(c)
true_diff = norm(E0 * convert(TensorMap, gs) - convert(TensorMap, H * gs))  # dense cross-check

println("norm(a)           = ", norm(a))
println("norm(b)           = ", norm(b))
println("norm(a - b)  MPS  = ", mps_norm)   # <-- returned by MPSKit
println("norm(a - b)  TRUE = ", true_diff)  # <-- correct value (dense)
```

Output:

```
norm(a)           = 7.296229810558757
norm(b)           = 7.296229810558758
norm(a - b)  MPS  = 14.592459621116891     # ≈ 2*norm(a); i.e. it computed a + b
norm(a - b)  TRUE = 3.1957415666152057e-12
```

`convert(TensorMap, a - b)` is also `≈ 14.59`, so the returned MPS *state* is genuinely wrong, not just its `norm`.

## Cross-check: the operands are individually correct

`H * gs` and `E0 * gs` are the same vector, verified densely:

```julia
norm(convert(TensorMap, E0 * gs) - convert(TensorMap, H * gs))   # ~1e-12
```

`norm`, `dot`, and scalar `*` are all individually correct (they match the dense values); only the `+` / `-` combination is wrong.

## Isolation

- **It is the `+` / `-` operation.** `dot(a, b)`, `norm(a)`, `norm(b)` and `(-1) * b` all match their dense values exactly; only the sum/difference is wrong.
- **Not symmetry/planar-specific.** Reproduces identically for plain `ComplexSpace` (`ℂ^d`, shown above) *and* for planar `GradedSpace{PlanarTrivial}` (`ℙ^d`).
- **Trigger = proportional operands from different tensor networks.** The bug appears precisely when the two operands are (nearly) parallel vectors expressed through different gauges/bond bases and should nearly cancel:
  - `E0*gs - H*gs` (parallel, different networks) → **wrong**.
  - `(c*gs) - (H*gs)` for any coefficient `c` → wrong by a constant additive `≈ 2*norm(H*gs)`, independent of `c` (the result equals `a + b`, not `a - b`).
  - `(H*gs) - (H*gs)` (identical network) → correct (`~1e-15`).
  - `psi - H*psi` for a *random* (non-parallel) `psi` → correct.
  - `psi - FiniteMPS(convert(TensorMap, psi))` (re-canonicalized to the same Schmidt basis) → correct.
- **Smallest size:** `L = 4` (with `L = 3` it happens to be fine); reproduces for all `L ≥ 4`.
- **Secondary symptom:** the result also depends on the lazy gauge-cache state of the operands at call time (e.g. whether they were `convert`ed/gauged before the subtraction), which points at the same root cause. The fresh-call behaviour shown above is deterministic.

## Where it lives

`Base.:+(::FiniteMPS, ::FiniteMPS)` at `src/states/finitemps.jl:477` (and `-` at `:551`, which is defined as `ψ₁ + (-1 * ψ₂)`). The direct-sum construction there uses `left_orth!` / `right_orth!`; the cancellation is lost when the two operands' virtual bases differ, which suggests the recombination/orthogonalization step does not correctly account for the cross-overlap of the operands.

## Environment

- MPSKit `v0.13.13`, TensorKit `v0.17.1`
- Julia `1.12.6`, Linux
- Base commit `547c9130` (`QuantumKitHub/MPSKit.jl`). The reproducer uses only public API; the `FiniteMPS` arithmetic / `norm` / `dot` code paths are unmodified relative to this commit, so it reproduces on a clean checkout.

lkdvos:
Fixes #473.

`Base.:+`/`-` on `FiniteMPS` returned a wrong state whenever the operands were near-parallel vectors carried by different tensor networks: `norm(E₀ * gs - H * gs)` came out as `2 * norm(gs) * E₀` instead of ~0, i.e. the result was `a + b`.

The addition kernel is not at fault — it reproduces the dense sum to ~1e-15 for `L = 2..8` with mismatched bond dimensions.
The bug is in the lazy gauging.
`CView.getindex` moved the gauge center by factorizing `ACs[j]` and writing the result into `ALs[j]`/`Cs[j+1]` (resp. `Cs[j]`/`ARs[j]`) *unconditionally*, including when those entries were already cached.
Different code paths install different factorizations — DMRG writes `AL, C` from a truncated SVD via `left_gauge!`, while the lazy sweep re-derives them from a positive Householder QR — so the recomputed tensor differs from the cached one by a bond unitary.
In the reproducer it is exactly a sign flip: `‖AL[1]_before − AL[1]_after‖ = 2√2` on a `‖AL‖ = √2` tensor.

`Base.:+` read `ψ.AL[1]`, then read `ψ.AL[2]`, which triggered the sweep that replaced `ALs[1]`.
The result therefore combined a stale `AL[1]` with a fresh `AL[2]`, flipping the sign of one operand's left block.
The same staleness was latent in every consumer that reads several gauge tensors across a center move, `dot` included.

## Changes

* `CView.getindex` now skips the factorization when the target entries already exist, so a cached tensor is never silently replaced. This also avoids redundant factorizations.
* `Base.:+` settles both operands' gauge centers at the seam and snapshots the tensors it needs before building anything, so it only ever combines tensors from a single canonical form. It also gained the boundary-virtual-space check that `FiniteMPO`'s `+` already had.
* `FiniteMPS`/`FiniteMPO` addition now promotes the scalar type, so `FiniteMPS{Float64} + FiniteMPS{ComplexF64}` works instead of erroring. This needs a new `Base.similar(ψ::FiniteMPS, ::Type{S<:Number})` (built on a small `similar_scalartype` utility) and widens `Base.:+(ψ₁::MPS, ψ₂::MPS) where {MPS <: FiniteMPS}` to accept two different `FiniteMPS` types.

## Tests

The new `test/states/arithmetic.jl` covers all three levels of the bug: that reading `AL[2]` leaves the already-cached `AL[1]`/`C[1]` untouched, that `ψ - flip_gauge!(copy(ψ))` vanishes, and the issue's own `norm(E₀ * gs - H * gs) ≈ 0`.
It also adds the dense cross-checks of `±` for states that were missing entirely, across planar, U(1) and SU(2) spaces, plus the scalar-type promotion cases for `FiniteMPS` and `FiniteMPO`.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -60,6 +60,15 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
   combined tensors belonging to two different gauges. The same staleness was latent in every
   consumer that reads several gauge tensors across a center move, `dot` included.
   ([#473](https://github.com/QuantumKitHub/MPSKit.jl/issues/473), [#484](https://github.com/QuantumKitHub/MPSKit.jl/pull/484))
+- Addition of single-site operands. `FiniteMPS + FiniteMPS` asserted `length > 1`, and
+  `FiniteMPO + FiniteMPO` threw a space error, because both split the chain into a left and a
+  right block and fuse them at the seam — of which there is none at length 1. Such an operand has
+  no internal bond to fuse, so the sum is now simply the sum of the two tensors.
+  ([#484](https://github.com/QuantumKitHub/MPSKit.jl/pull/484))
+- `convert(TensorMap, ::FiniteMPO)` on a single-site MPO stripped the left virtual leg of `mpo[1]`
+  and the right virtual leg of `mpo[end]` and contracted the two — which at length 1 is the *same*
+  tensor, so it returned `O * O` on twice the physical space instead of `O`.
+  ([#484](https://github.com/QuantumKitHub/MPSKit.jl/pull/484))
 
 ### Performance
 
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -21,6 +21,9 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 ### Added
 
+- Addition of `FiniteMPS`/`FiniteMPO` with different scalar types, through a new
+  `Base.similar(ψ, ::Type{S})` for `S <: Number` on `FiniteMPS`. ([#484](https://github.com/QuantumKitHub/MPSKit.jl/pull/484))
+
 ### Changed
 
 - `environments` now follows a single positional contract for every state and operator kind:
@@ -48,6 +51,15 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 ### Removed
 
 ### Fixed
+
+- `Base.:+`/`-` on `FiniteMPS` returned a wrong state for near-parallel operands carried by
+  different tensor networks, e.g. `norm(E₀ * gs - H * gs)` coming out as `2 * norm(gs) * E₀`
+  instead of ~0. The lazy gauge sweep in `CView.getindex` re-derived `AL`/`C` entries that were
+  already cached, and since different code paths install different factorizations (a truncated SVD
+  from DMRG vs. a positive QR from the sweep) the replacement differed by a bond unitary, so `+`
+  combined tensors belonging to two different gauges. The same staleness was latent in every
+  consumer that reads several gauge tensors across a center move, `dot` included.
+  ([#473](https://github.com/QuantumKitHub/MPSKit.jl/issues/473), [#484](https://github.com/QuantumKitHub/MPSKit.jl/pull/484))
 
 ### Performance
 
--- a/src/operators/mpo.jl
+++ b/src/operators/mpo.jl
@@ -122,7 +122,9 @@ function Base.:+(mpo1::FiniteMPO{<:MPOTensor}, mpo2::FiniteMPO{<:MPOTensor})
         right_virtualspace(mpo1, N) == right_virtualspace(mpo2, N)
 
     halfN = N ÷ 2
-    A = storagetype(eltype(mpo1))
+    # the fusers carry the promoted scalar type, so every contraction below follows suit
+    T = promote_type(scalartype(mpo1), scalartype(mpo2))
+    A = TensorKit.similarstoragetype(storagetype(eltype(mpo1)), T)
 
     # left half
     F₁ = isometry(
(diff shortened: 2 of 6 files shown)


VictorVanthilt:
Just curious, when did you run into issues with 1-site MPOs?

lkdvos:
```julia
@testset for L in 1:6
    # ...
end
```

I knew this was a problem, (I wrote the assertions for it), and figured might as well fix this here)

PR #486, opened 2026-07-29
Try using DefaultAllocator instead of BufferAllocator

kshyatt:
I think `BufferAllocator` doesn't work with the GPU arrays...


lkdvos:
Let me hold off on this for a moment while I figure out #467, I'm going to try and have a best-of-both-worlds version where this is handled automatically.

kshyatt:
Great, feel free to just nuke this then :D

PR #491, opened 2026-07-31
`dot` shortcut for `FiniteMPS`

borisdevos:
Slightly cheaper this way, and matters somewhat in expectation value calls.


lkdvos:
Do you feel like it might make sense to additionally just replace that in the expectation_value calls directly?

lkdvos (approved):

Issue #459, opened 2026-07-14
Performance with long range interactions is really poor


AndreasChristophilopoulos:
I had some code that I decided to migrate from TenPy, but unfortunately performance is far worse for some reason. The relevant part of the Julia code is this:
```
disordered_distance(i, j, dx) = abs(j - i + dx[j] - dx[i])

function build_rydberg_hamiltonian(L, Δ, Ω, σ, t, V, seed)
    # Random displacements per site, uniform in [-sigma/2, sigma/2]
    rng = MersenneTwister(seed)
    dx = (rand(rng, L) .- 0.5) .* σ

    Vphys = ℂ^2
    physical_spaces = fill(Vphys, L)

    σx = TensorMap([0 1; 1 0], Vphys ← Vphys)
    σz = TensorMap([1 0; 0 -1], Vphys ← Vphys)
    Id = id(Vphys)
    Nop = (Id + σz) / 2
    Sp = TensorMap([0 1; 0 0], Vphys ← Vphys)
    Sm = TensorMap([0 0; 1 0], Vphys ← Vphys)

    lattice = fill(Vphys, L)

    on_site_terms = [(i,) => -Δ * Nop + Ω * σx for i in 1:L]
    interaction_terms = []
    for i in 1:L, j in 1:L 
        if i < j
            r = disordered_distance(i,j,dx)
            Vij = V/r^6
            tij = t/r^3
            push!(interaction_terms, (i,j) => Vij*Nop⊗Nop + tij * ( Sp ⊗ Sm + Sm⊗Sp)) 
        end
    end
    H = FiniteMPOHamiltonian(lattice, on_site_terms..., interaction_terms...) 

    return H, physical_spaces
end

# Sample values 
Δ = 1.0; Ω = 1.0; V = 1.0; t=1.0;
L = 25
chi_max = 50
psi = FiniteMPS(L, ComplexSpace(2), ComplexSpace(chi_max))

H, physical_spaces = build_rydberg_hamiltonian(L, Δ, Ω, sigma, t, V, seed)

psi, envs, dmrg_info = find_groundstate!(psi, H, DMRG(; maxiter=10))

```

As you can see this implements a Hamiltonian with long range interactions. For comparison, the TenPy code model definition is:
```
class RydbergModel(CouplingMPOModel):
    
    def init_sites(self, model_params):
        # symmetry settings
        conserve = model_params.get('conserve', None, str)
[...]

PR #493, opened 2026-07-31
Tests for Hamiltonian derivatives and long-range models

borisdevos:
This one took a while to debug since I don't really know how these Jordan MPOs work. I now understand a little better 😄 

It turns out two things have to occur simultaneously for this to have been noticed. First, the scalartype of the Hamiltonian and the state must mismatch, such that the constructor of the AC Hamiltonian doesn't get fast-forwarded to the inner constructor but lands on the fallback one with the typo. Second, the Hamiltonian needed to not have the ending block while also have the finishing one. Since I was playing around with something semi-long range, I came across this, which makes sense. Nearest-neighbor Hamiltonians will always have B. 

Coincidentally, we never caught this in the tests because we don't look at long-range Hamiltonians. So I added that to the tests, along with fairly extensive tests on the derivatives themselves directly. Some might be overkill, but better safe than sorry. Locally the performance of these long-range Hamiltonians is notably terrible (see #459)

I also discovered the changelog, so I decided to add to that :)


test/algorithms/groundstate.jl
```diff
@@ -291,6 +291,46 @@ end
     end
 end
 
+@testset "Long-range Hamiltonian with real scalartype" verbose = true begin
+    # A real `H` optimised over a complex MPS sends the derivative operators through their
```
lkdvos: I'm not sure I'm understanding what this means to be honest, would you mind explaining?

borisdevos: I think I explained it elsewhere better, but basically when the scalartype of the Hamiltonian and the state aren't equal at the real/complex level (I didn't test at the level of precision), `JordanMPO_AC_Hamiltonian` will go through the fallback outer constructor.

Two things had to happen at the same time. First, there needs to be an onsite term so the D-block exists as well as an ending block B (to catch the typo), and second they need to have a different scalartype to the environments. The D-block doesn't inherit the type of these environments in this particular case, as can be seen in: https://github.com/QuantumKitHub/MPSKit.jl/blob/0ff11c35da63d33f177eb2846b84503e3cab2bd8/src/algorithms/derivatives/hamiltonian_derivatives.jl#L56

So the typed inner constructor doesn't work, and it goes through the untyped one which wants to convert accordingly, which then errored due to the typo.

I think I wrote this poorly in the comments, so I can change this if it helps to remember why those tests exist. 

Edit: I had definitely confused B with D in my initial response here, I fixed them so it's correct now.

borisdevos:
The typo actually got fixed in #467 so this PR is just adding tests to the fallback constructor

PR #497, opened 2026-08-03
docs: vendor the VitePress config; serve static assets from public/

### LINKED ISSUE #477: Markdown titles in docstrings not rendering properly in published docs pages
It seems the markdown (sub)titles used in the docstrings no longer render properly since #445. They now just show as plain text. The default Documenter.jl theme also didn't do much with these, but at least it rendered markdown titles in bold face, which was usually enough to indicate some structure in the docstrings.

It would be nice if this could be restored using the new DocumenterVitePress rendering.

[EDIT] Aparently this issue is only present on Linux (confirmed on Ubuntu and Mint), while things render fine on MacOS or Windows.

### LINKED ISSUE #478: Documentation search can't target html identifiers in the `lib` page
Before #445, searching for common names such as `FiniteMPS` would give results both [in the manual page](https://quantumkithub.github.io/MPSKit.jl/v0.13.13/man/states/#FiniteMPS) as well as [in the library page](https://quantumkithub.github.io/MPSKit.jl/v0.13.13/lib/lib/#MPSKit.FiniteMPS). On `dev`, it seems the search function can't target any of the html tags from the library page, which is quite annoying. Searching `FiniteMPS` now just leads to the manual page, even though the corresponding [link to the library page](https://quantumkithub.github.io/MPSKit.jl/dev/lib/lib#MPSKit.FiniteMPS) still seems to work. This is especially annoying when searching for things that only occur in the library page but have a proper docstring. For example, searching `AC2_hamiltonian` doesn't give anything useful, while the corresponding tag should just be there on the library page.

Even when the search window shows the right docstring on the library page highlighted in the search box, clicking further just takes you to the very top of the library page...

lkdvos:
Split out of #449 so it can be reviewed on its own; independent of the documentation content.

`MarkdownVitepress` does not expose `themeConfig.search` or `markdown.config`, so `docs/src/.vitepress/config.mts` is vendored from the DocumenterVitepress template with the MPSKit hooks added on top. Since a vendored copy silently rots, `make.jl` checksums the upstream template and warns when it changes, naming the versions our copy is a faithful superset of.

VitePress serves `src/public/` verbatim at the site root, which is what the logo and favicon need; `src/assets/` is for files Documenter processes. Moved accordingly.

## What this does about #477 and #478

Two independent things were going wrong.

**Structure was invisible in docstrings.** Documenter rewrites markdown headings inside a docstring into bold-only paragraphs (`recursive_heading_to_bold!`), so there is no `<h*>` left to style. A markdown-it rule in `config.mts` tags exactly those paragraphs and the CSS gives them a heading-like treatment — spacing, size, and a light rule beneath — so the structure reads even where bold does not.

**Bold did not render as bold on Linux.** VitePress' reset sets `font-synthesis: style`, permitting a synthesised oblique but not a synthesised weight. That is fine wherever the resolved font has a real bold face, as `-apple-system` and `Segoe UI` do. But the DocumenterVitepress stack names three families it never loads (`Barlow`, `Inter var experimental`, `Inter var`), and on Linux fontconfig satisfies those *and* the platform names after them with a single regular face — every one of the 15 names in that stack resolves to `DejaVuSans.ttf` "Book" here. So `font-weight: 700` computes correctly and still paints at regular weight.

This PR allows weight synthesis and **does not touch the font stack**, so nothing changes on platforms that already resolve a real bold face. It is deliberately the smallest change that fixes the reported symptom, and it is what I intend to propose upstream — at which point this rule can be deleted.

There is a deeper issue behind it, kept separate on purpose: VitePress self-hosts Inter with a real 100–900 weight axis, but the DocumenterVitepress stack names the family names VitePress used *before* [vuejs/vitepress#3694](https://github.com/vuejs/vitepress/pull/3694) renamed the face to plain `Inter`, and omits plain `Inter`. Every DocumenterVitepress site therefore downloads a font it never uses. Naming it would give real bold faces rather than synthesised ones, but would also change body text on every platform — a look-and-feel decision for upstream, not something to fold into a bug fix.

## Unverified

The VitePress build cannot run on my machine (rollup's native module needs a newer glibc), so I have not seen this rendered. The mechanism checks out at every layer I could inspect, but confirmation that bold now reads as bold on Linux needs CI or a preview deploy — ideally from @leburgel, who reported #477 on Ubuntu/Mint.


leburgel:
Docstring headers look bold to me now, and I can target library entries with the search like I wanted!

PR #498, opened 2026-08-03
docs: move the changelog to the repo root

lkdvos:
Split out of #449 so it can be reviewed on its own; independent of the documentation content.

A changelog is the first thing a contributor looks for in the repository, so keep the canonical copy at `CHANGELOG.md` where GitHub renders it, and have `make.jl` copy it into `src/` at build time for the rendered page. The generated copy is gitignored so it cannot drift from the root file.


lkdvos:
Dropping the changelog migration — the changelog stays at `docs/src/changelog.md`.

#499 has been updated so that `CONTRIBUTING.md` and the PR template point contributors at that path rather than a root `CHANGELOG.md`. The branch `docs/changelog-root` is left in place if this is ever worth revisiting.

PR #504, opened 2026-08-05
Allow calculating entanglement of left virtual boundary of finite MPS

borisdevos:
I was contemplating changing the argument also to `bond` to refer to the actual cut location instead of `site`, so there's no confusion with left/right, but I decided ultimately against it and just added some clarification in the relevant docstrings.

### DIFF
--- a/src/algorithms/toolbox.jl
+++ b/src/algorithms/toolbox.jl
@@ -7,7 +7,8 @@ MPS state or directly from an entanglement spectrum as obtained from
 [`entanglement_spectrum`](@ref).
 
 When called on an MPS with an integer `site`, the entropy is computed across the
-entanglement cut to the right of site `site`. For `InfiniteMPS`, omitting `site` returns a
+entanglement cut to the right of site `site` (or equivalently, to the left of bond `site`).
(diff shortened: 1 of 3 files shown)


src/algorithms/toolbox.jl
```diff
@@ -7,7 +7,8 @@ MPS state or directly from an entanglement spectrum as obtained from
 [`entanglement_spectrum`](@ref).
 
 When called on an MPS with an integer `site`, the entropy is computed across the
-entanglement cut to the right of site `site`. For `InfiniteMPS`, omitting `site` returns a
+entanglement cut to the right of site `site` (or equivalently, to the left of bond `site`).
```
lkdvos: I'm not so convinced by this wording, mostly because bond `site` assumes you are not counting the bond left of the first site, which doesn't work for `WindowMPS`, which is precisely the case where this entanglement entropy would make the most sense to compute :)

lkdvos (commented):
Other than the minor comment, looks good to me, thanks!

lkdvos (approved):

PR #508, opened 2026-08-12
Fix `Multiline` (`MultilineMPS`/`MultilineMPO`) correctness and API inconsistencies

borisdevos:
[Relevant edit deeper in conversation]

I found a cluster of bugs that all traced back to the same root cause, namely `Multiline{T}` was inconsistent about what it is. Sometimes it was treated as a 2D array indexed by `(row, col)`, and sometimes as a 1D sequence of `T`-typed lines, this while different parts of the code picked different conventions without checking consistency. The real bugs were:

- `MultilineMPO * MultilineMPS` had never worked (straight up method error), and fixing that trivially still left both a shape bug (`length` counting `rows*cols` while `map`/`collect` only ever produced `rows` elements) and a silently wrong row-shift bug (`fidelity 0` where it should be `1`).
- `expectation_value(::MultilineMPS, ::MultilineMPO, ...)` had a fallback method that silently returned something meaningless (`prod` instead of `sum`, no row shift, discarded `envs` which was noted through a todo) for any `MultilineMPO` line type it wasn't explicitly guarded against. Related to this, there was half-support for Hamiltonian or finite MPO multilines even though they're meaningless. Also the environments weren't being correctly converted, so they were being recalculated every row.
- `isfinite`, `changebonds(::MultilineMPO, ::SvdCut)`, and `axes(m, 2)` all threw outright.

A summary of the changes I made:

"Semantics":
- `Multiline{T}` now consistently treats `length`/`eltype`/`iterate`/`m[i]` as referring to the lines (matching the periodic rows), while `size`/`axes`/`eachindex` refer to the `(nrows, ncols)` lattice shape. These two views are intentionally different (`length(m) != prod(size(m))`). I also documented this in docstrings.
- Fixed multiplication: both were broken. The row-shifting is implemented and documented.
- Deleted the `expectation_value` fallback. Changed the infinite MPS-MPO expectation value to correctly account for environments.
- Fixed `isfinite`, `changebonds(::MultilineMPO/::MultilineMPS, ::SvdCut)`, and `axes(m, i)`.
- Some grassmann code needed a change related to the 1d vs 2d view thing above.
- Added the missing instance-level `spacetype`/`sectortype`/`storagetype` for multiline. I needed this for the custom show (see below).

Construction restrictions:
- `MultilineMPO` only accepts `InfiniteMPO` lines now, enforced at constructor level. This is deliberate, since `Multiline` is also used via converts in environments, quasiparticles etc.
- Removed the `AbstractMatrix` constructor that silently built finite-line `MultilineMPO`s, and the dead code in `orthoview.jl` that existed only to support finite MPS multilines.

Display:
- `MultilineMPS`/`MultilineMPO` custom show: added a `summary` which looks at the 2d shape, and a `show` that renders each row using that row's own existing `show` method. Here below an example of how it looks now:

Here the multiline MPS:
<details>

```
2×2 MultilineMPS(ComplexF64, Vect[IsingAnyon]) with maximal dimension 10.0:
row 1:
2-site InfiniteMPS(ComplexF64, Vect[IsingAnyon]) with maximal dimension 10.0:
| ⋮
| (:σ => 7)
├─[2]─ (:σ => 1)
│ (:I => 5, :ψ => 5)
├─[1]─ (:σ => 1)
│ (:σ => 7)
| ⋮

  ⋮
row 2:
2-site InfiniteMPS(ComplexF64, Vect[IsingAnyon]) with maximal dimension 10.0:
| ⋮
| (:I => 5, :ψ => 5)
├─[2]─ (:σ => 1)
│ (:σ => 7)
├─[1]─ (:σ => 1)
│ (:I => 5, :ψ => 5)
| ⋮
```
</details>

And here a multiline MPO:

<details>

```
2×2 MultilineMPO(ComplexF64, Vect[IsingAnyon]) with maximal dimension 1.4142135623730951:
row 1:
2-site InfiniteMPO(ComplexF64, Vect[IsingAnyon]) with maximal dimension 1.4142135623730951:
| ⋮
| (:σ => 1)
┼─[2]─ (:σ => 1)
│ (:σ => 1)
┼─[1]─ (:σ => 1)
│ (:σ => 1)
| ⋮

  ↓  (row 1 maps onto row 2)
row 2:
2-site InfiniteMPO(ComplexF64, Vect[IsingAnyon]) with maximal dimension 1.4142135623730951:
| ⋮
| (:σ => 1)
┼─[2]─ (:σ => 1)
│ (:σ => 1)
┼─[1]─ (:σ => 1)
│ (:σ => 1)
| ⋮

  ↓  (row 2 maps onto row 1)
```

</details>

Tests:
- New `test/operators/multilinempo.jl`: exact-fidelity tests for the row-shift convention in both `MultilineMPO * MultilineMPS` and `MultilineMPO * MultilineMPO`, using a permutation MPO whose fixed point is known exactly.
- `test/algorithms/statmech.jl`: multi-row `expectation_value` regression test. Also a test that the Hamiltonian loophole above still correctly throws.
- `test/states/multilinemps.jl`: added instance-level `isfinite(ψ)` (only the type-level form was covered before).

Documentation:
- `docs/src/man/states.md`: added the row-shift convention, what's currently supported (infinite lines only), and subtleties.
- `docs/src/man/operators.md`: new `MultilineMPO` section covering the same convention, and emphasising the MPO\*MPO non-composability caveat.
- Updated docstrings on `Multiline`, `MultilineMPO`, `MultilineMPS`, and `expectation_value`.
- Added to the changelog, as the PR template wished of me 😄 .

Some things I didn't address because I'm not sure what to do with them:
- VectorInterface stuff is still broken for `MultilineMPS`, simply because `InfiniteMPS` itself doesn't implement `zerovector`/`scale`.
- `norm(ψ) == sqrt(nrows)` for a `MultilineMPS` of individually-normalized rows is unchanged. This goes through `dot`/`inner` sum over rows. I don't know if this is what you want.

## Checklist

- [x] Runic formatter is run


docs/src/man/operators.md
```diff
@@ -274,3 +274,22 @@ a collection (direct sum) of spaces, one for each row/column.
 ```@example operators
 left_virtualspace(H_ising, 1), right_virtualspace(H_ising, 1), physicalspace(H_ising, 1)
 ```
+
+## MultilineMPO
+
+Much like a [`MultilineMPS`](@ref) is a repeating set of `InfiniteMPS` lines, a [`MultilineMPO`](@ref) is a repeating set of `InfiniteMPO` lines.
+This is typically the row-to-row or column-to-column transfer matrix of a 2D classical partition function, or a boundary MPO in the context of PEPS.
```
leburgel: "boundary MPO" isn't really a thing I think. Also, a PEPS norm network is really a classical partition function. So you could say "including for example the norm network corresponding to a PEPS overlap."

docs/src/man/operators.md
```diff
@@ -274,3 +274,22 @@ a collection (direct sum) of spaces, one for each row/column.
...
+See the `MultilineMPS` section of the [states](@ref um_states) page for the row-shift convention relating a `MultilineMPO` to the `MultilineMPS` it acts on: row `i` of the operator maps row `i` of the state onto row `i + 1`.
+
+```@example operators
+mpo_multi = MultilineMPO([mpo, mpo])
+```
+
+Only `InfiniteMPO` lines are supported; neither Hamiltonians nor finite MPOs are allowed currently.
```
leburgel: Why not allow finite MPOs? I understand the exclusion of multiline Hamiltonians, but multiline finite MPOs make perfect sense for finite partition functions. If there is currently no support for this, I don't mind just throwing a method error with a nice message saying that there's no support yet.

I've personally actually used this here and there already, where I just used it as a vector of finite MPOs and wrote my own functionality for it (for example [here](https://gist.github.com/leburgel/2d07b915d2133429acfc1155da32ff02)). Since the objects themselves make sense, it might be a bit inconvenient to now disallow constructing it altogether.

docs/src/man/operators.md
```diff
@@ -274,3 +274,22 @@ a collection (direct sum) of spaces, one for each row/column.
...
+This is enforced by the constructors rather than by the type itself, so `MultilineMPO([H, H])` for a Hamiltonian `H` throws a `MethodError` right at the construction site, instead of producing an object that only fails once used.
+
+!!! note "`*` between `MultilineMPO`s does not compose the way you might expect"
+    Because of the `+1` row shift, multiplying two `MultilineMPO`s produces an operator that spans two rows of the lattice: `(O1 * O2)` maps row `i` onto row `i + 2`.
+    As a consequence, `(O1 * O2) * ψ != O1 * (O2 * ψ)` in general, for any definition of the product.
+    The two sides are genuinely different operations (one fused two-row step, versus two sequential one-row steps).
+    Treat `O1 * O2` as a "fused double row", not as an operator that can be composed further.
```
leburgel: I don't think this makes sense at all, we should really have that `(O1 * O2) * ψ != O1 * (O2 * ψ)`. I think there might be some misunderstanding between us about what this multiplication is supposed to do.

docs/src/man/states.md
```diff
@@ -170,5 +170,35 @@ col = 2
...
+i.e. applying `O` shifts every row up by one.
+This convention is not just a property of `*`: it is baked into `environments`, `expectation_value`, and the derivative machinery as well, all of which pair `ψ[row + 1]` against `O[row]`.
+It is also why `O * ψ` requires `size(O, 1) == size(ψ, 1)`, rather than the row counts of operator and state being independent.
+
+### What is currently supported
+
+Only lines that are themselves infinite are supported: a `MultilineMPS` is always built out of
+[`InfiniteMPS`](@ref) lines, never [`FiniteMPS`](@ref) or [`WindowMPS`](@ref) ones.
+This matches the intended use case of rows/columns in an infinite 2D network.
+The constructors reject finite lines.
```
leburgel: Again, I'm not sure if this actually makes sense.

borisdevos:
After having discussed with @leburgel, I realised I didn't fully appreciate/understand what multiline actually represents, so recent changes rectify some mistakes I made, and tried to explain these niche things more cleanly in the docs/docstrings.

- I now allow finite lines at type and constructor level, but there's still no support for these at the algorithmic level. Hamiltonians are still correctly rejected.
- Multiplying multiline things never made sense, so I removed these methods and added the one that does make sense.
- Some orthoview changes I previously removed, plus fixing bounds checking on these views.
- I took the liberty of not calling the value that you would calculate with multiline an expectation value, but rather what it is, a dominant eigenvalue. So I introduced a new function `dominant_eigenvalue`, explained why it's separate from `expectation_value` and when these actually collide. The tricky thing here was that the VUMPS code worked for both MPOs and Hamiltonians, so there's a funky check there to see whether expectation values or dominant eigenvalues are being logged.
- Docs and docstrings clarifications.

docs/src/changelog.md
```diff
@@ -72,16 +76,38 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
   not the requested correlator. ([#489](https://github.com/QuantumKitHub/MPSKit.jl/pull/489))
 - TimerOutputs 1.x is now required. The timing tables printed at `verbosity > 3` use the new
   layout (tree guides, heat bars) and additionally report per-section GC time.
+- `Multiline` (and therefore `MultilineMPS`/`MultilineMPO`) now consistently treats
+  `length`/`eltype`/`iterate`/`m[i]` as referring to the individual lines it stores
+  (`length(m) == nrows`), while `size`/`axes`/`eachindex` refer to the `(nrows, ncols)` lattice
+  shape.
```
lkdvos, line 82: I think this is the change I have the most questions about. I like that you state that it is consistent, even though it is clearly inconsistent between most of these methods 😆 (`length(x) == prod(size(x))` is now explicitly broken).

Am I understanding this correctly that you are effectively claiming this is interpreted as a vector of lines, with a little bit of syntactic sugar to make `m[i, j]` work and then have `size` follow the syntactic sugar?
With that in mind, I think most of the code is actually not calling `m[i,j]` at all, since for MPS that does not really mean anything (you'd need the `.AC` or something anyways), and for MPO I think we are mostly doing things by first unpacking the rows and only then indexing the sites.
Would it also work to actually pull this through all the way, just remove the lattice shape from the equation entirely?

src/algorithms/expval.jl
```diff
@@ -181,26 +191,68 @@ function expectation_value(ψ::FiniteQP, mpo::FiniteMPO)
...
-        ψ::MultilineMPS, O::MultilineMPO{<:InfiniteMPO},
+# fallback
+function expectation_value(ψ::AbstractMPS, mpo::AbstractMPO, envs...)
+    return dot(ψ, mpo, ψ) / dot(ψ, ψ)
+end
+
+# Dominant eigenvalues
+# --------------------
+"""
+    dominant_eigenvalue(ψ, O, [environments]) -> λ
```
lkdvos, line 204: Can you maybe explain a bit better what is going on here? I am not really following most of the discussion here to be honest. Why is this not `⟨ϕ|O|ϕ⟩`?
From what I can tell it really is that, in the sense that because you take the product over the transfer of all of the rows, you get exactly the overlap after taking it through an entire unit cell, so I don't really understand what is going on here.

I think I might get that it is not an expectation value in the sense that it is not the expectation value of an operator in the statmech sense, for which indeed you'd use `dot(psi, O, phi)` instead, but if `expectation_value(psi, O) = dot(psi, O, psi)` by definition, I don't immediately see the benefit of making this a separate function.

src/states/multilinemps.jl
```diff
@@ -1,15 +1,20 @@
 # MultilineMPS
 # ------------
-const MultilineMPS = Multiline{<:InfiniteMPS}
+#TODO: add support for finite MPS
+const _MPSs = Union{InfiniteMPS, FiniteMPS}
```
lkdvos, line 4: Is this `AbstractMPS`?

src/states/orthoview.jl
```diff
@@ -278,25 +278,28 @@ Base.axes(psi::CView{<:AbstractFiniteMPS}) = map(n -> 0:(n - 1), size(psi))
...
-        true
-    else
-        checkbounds(Bool, CView(first(psi.parent.data)), b)
-    end
+    return checkbounds(Bool, _lineview(psi), b)
 end
+_lineview(psi::ACView) = ACView(first(parent(psi.parent)))
+_lineview(psi::ALView) = ALView(first(parent(psi.parent)))
+_lineview(psi::ARView) = ARView(first(parent(psi.parent)))
+_lineview(psi::CView) = CView(first(parent(psi.parent)))
```
lkdvos, line 302:
```suggestion
    return checkbounds(Bool, parent(psi), a) &&
        checkbounds(Bool, @inbounds psi[a], b)
end
```

Does this also work?

src/states/orthoview.jl
```diff
@@ -278,25 +278,28 @@ Base.axes(psi::CView{<:AbstractFiniteMPS}) = map(n -> 0:(n - 1), size(psi))
...
+    return (Base.OneTo(length(parent(psi.parent))), 0:length(first(parent(psi.parent))))
 end
 
-#the checkbounds for multiline objects needs to be changed, as the first index is periodic
-#however if it is a Multiline(Infinitemps), then the second index is also periodic!
+const MultilineOrthoView{S} = Union{ACView{S}, ALView{S}, ARView{S}, CView{S}}
+
+# the row index is always periodic
+# for infinite lines the column index is periodic as well
+Base.checkbounds(::Type{Bool}, ::MultilineOrthoView{<:Multiline{<:InfiniteMPS}}, a, b) = true
```
lkdvos, line 291: See comment below, which then might make this obsolete?

src/utility/multiline.jl
```diff
@@ -102,6 +118,7 @@ end
 
 VectorInterface.add!!(x::Multiline, y::Multiline, α::Number, β::Number) = add!(x, y, α, β)
 
+# FIXME? is it intentional that a nontrivial multilinemps of normalised rows never has norm 1?
```
lkdvos, line 121: I'm actually wondering if this should be `prod` instead of `sum` now...

src/utility/multiline.jl
```diff
@@ -117,6 +134,7 @@ site_type(::Type{Multiline{S}}) where {S} = site_type(S)
 bond_type(::Type{Multiline{S}}) where {S} = bond_type(S)
 site_type(st::Multiline) = site_type(typeof(st))
 bond_type(st::Multiline) = bond_type(typeof(st))
-TensorKit.sectortype(::Type{Multiline{T}}) where {T} = sectortype(T)
-TensorKit.spacetype(::Type{Multiline{T}}) where {T} = spacetype(T)
-TensorKit.storagetype(::Type{Multiline{T}}) where {T} = storagetype(T)
+for ftype in (:spacetype, :sectortype, :storagetype)
+    @eval TensorKit.$ftype(::Type{Multiline{T}}) where {T} = $ftype(T)
+    @eval TensorKit.$ftype(m::Multiline) = $ftype(typeof(m))
```
lkdvos, line 139: I somehow was under the impression that the latest versions of TensorKit already do this forwarding, is that wrong?

docs/src/changelog.md
borisdevos: It's consistent for me in the sense that when working with a multiline MPS what really makes it different is the layers of MPSs, so it made sense to me that length and eltype should refer to the layers. Also, `m[i, j]` is actually not for iMPS because we don't even allow slicing it (as you say, it doesn't make sense), though it is supported for iMPOs. 

So yeah in a nutshell, I am indeed trying to make multiline behave like an `AbstractVector` at the level of iterating and index slicing correctly over its length, but it's not fully a subtype of that because you still want 2d cartesian indexing over something like `m.AL`. The initial motivation for this was that some `map` (among others) on a multiline MPS would previously make an nrows*ncols vector with only nrows defined entries.

Do you mean by removing the lattice shape to just treat multiline as a vector of lines? This is certainly doable, though every `size(mpsi, 2)` would need to be replaced by `length(mpsi[1])`, among others which rely on the 2d structure. If you think this is better, then might as well make `Multiline{T} <: AbstractVector{T}`, right?

src/algorithms/expval.jl
borisdevos: The motivation for this was indeed semantics, trying to stress the fact that it's a stat-mech method, and there the way to reason this eigenvalue is really from applying transfer operators onto a state which lives on one side of the partition function, say the top. To really consider an overlap, which is the standard way of interpreting an expectation value, you'd need to sandwich your transfer operators from both sides, and the fixed point in the bottom is generally not that from the top. That's kind of what I tried explaining in the note in this docstring.

Really, from an identity point of view, you're right, but it's simply not what's being calculated, and I wanted to avoid a collision of meanings behind the same "expectation value" by just giving meaning to the more correct "dominant eigenvalue"

src/states/multilinemps.jl
borisdevos: No, because I want to leave out window MPS

docs/src/changelog.md
lkdvos: I think if you want to change it and make it consistent, might as well make it behave like a vector of lines, indeed, and possibly just add a new name for `size(mps, 2)`. It just really looks iffy to me to have both cartesian 2d indexing as well as linear indexing but have both of them not compatible, and I'm definitely fine with keeping `m.AL` alive, as that doesn't interfere with this, since it outputs a separate `ALView`, for which we can then separately decide to make it behave like a `Matrix` instead of a `Vector`.
TLDR, I think technically the current and previous approach are piracy, in the sense that we are stealing the `AbstractArray` interface and then incorrectly implementing it :)

src/states/multilinemps.jl
lkdvos: Why? 😉

src/algorithms/expval.jl
lkdvos: I'm still confused by this, it really is what is being calculated? I think my main point here is that this seems like you are introducing a lot of extra code mostly for pedantics, which I am sympathetic to but I fail to see a lot of benefits here. In general, most of our code uses the operator type to distinguish the meanings, and since `expectation_value` means `dot / norm`, which is precisely what we want in both cases, I am not really sure why this is so confusing?

docs/src/changelog.md
leburgel: I prefer just making it behave like a vector, so `length(m) == prod(size(m)) == size(m, 1)`. We can call what is now `size(m, 2)` something like `width(m)` returning `length(m[1])`. Allowing two-dimensional indexing for the tensor views is fine. Whether or not we actually make `Multiline{T} <: AbstractVector{T}` is the same to me.

I just really want to be able to call `m[3]` and get the third line. Given this, making it consistently behave like a vector is the only reasonable thing to do.

src/algorithms/expval.jl
leburgel: > Can you maybe explain a bit better what is going on here? I am not really following most of the discussion here to be honest. Why is this not `⟨ϕ|O|ϕ⟩`? From what I can tell it really is that, in the sense that because you take the product over the transfer of all of the rows, you get exactly the overlap after taking it through an entire unit cell, so I don't really understand what is going on here.

This is a bit of a semantics nightmare. I think the main point is that while this is indeed technically correct in the cases we use it for, what we call `expectation_value(expectation_value(ϕ, O)` is only equal to `⟨ϕ|O|ϕ⟩` inside a `leading_boundary` call or after its convergence, in the sense that this is only true when `O` is exactly the MPO that `ϕ` is the right (or top, in our conventional MPS notation) fixed point of. If you put in a different MPO, this produces total nonsense, since the "taking it through an entire unit cell" process relies entirely on the `leading_boundary` convergence criterion being satisfied. I think Boris' point is that it's misleading to call something `expectation_value` when it only produces a result in line with what you expect for exactly one operator, which is already assuming that you've obtained a converged eigenstate. The latter assumption can be fine in practice, but the former very quickly leads to bad mistakes when trying to use this in practice.

In that sense, calling this `dominant_eigenvalue` is less misleading, since it should immediately feel a bit wrong to call this without `ϕ` being an eigenvector of `ϕ`. While this may not be ideal, just calling the current implementation `expectation_value` really seems to imply that this is a sensible thing to do generically, but it just isn't.

To tag onto this, currently MPSKit provides no way at all to evaluate expectation values in the way we would actually need them in stat mech systems. Let me try to explain what I think the issues are concretely. I will already say I don't really know how to solve these, or even if we should do so within the context of MPSKit. Also this is probably really not the place to start this entire discussion. But at least this might bring everyone on the same page.

### What would be a "useful" expectation value to compute

Given some stat mech partition function whose row-to-row transfer operator corresponds to an MPO $O$, what we realistically need to compute as an "expectation value" is the normalized contraction of the partition function network where we insert some local tensor $E$ at site $(r, c)$ on our 2D lattice. If we have the leading eigenvector $\ket{\phi}$ of $O$ and we assume that $O$ is Hermitian, we could write what we want as

$$\langle E \rangle = \frac{\langle \phi | O_E | \phi \rangle}{\langle \phi | O | \phi \rangle},$$

where $O_E$ denotes the MPO where we've replaced the local MPO tensor at site $(r, c)$ by $E$. Just looking at the code in MPSKit, one might be tempted to assume that this can be computed as `expectation_value(expectation_value(ϕ, O_E) / expectation_value(expectation_value(ϕ, O)`. I have naively done this in the past and have lost quite a bit of time figuring out that while this sometimes works, it's actually just wrong in the general case.

### Warmup problem: MPOs don't have to be Hermitian

If we're dealing with a partition function that doesn't admit a Hermitian row-to-row transfer operator, $O$ will not be Hermitian. In that case what we would actually want is

$$\langle E \rangle = \frac{\langle \phi_L | O_E | \phi_R \rangle}{\langle \phi_L | O | \phi_R \rangle},$$

where $\ket{\phi_L}$ and $\ket{\phi_R}$ are the left and right leading eigenvectors of $O$ (or the bottom and top MPS fixed points in our usual notation). We have no support for this currently, while this is a very common thing to have to do. In particular, this always requires computing new environments of a "mixed" top-bottom transfer matrix to even be able to evaluate this. Ideally we should have some interface that actually allows to do this.

However, that's not really even the problem being targeted here.

### Even for Hermitian MPOs, the current multi-line implementation doesn't do this

Even when $O$ is perfectly Hermitian, there's no way to evaluate this using `expectation_value` for an MPO with 3 or more rows. That's where the real issue is. For a Hermitian $O$, one would really expect `expectation_value(ϕ, O_E) / expectation_value(expectation_value(ϕ, O)` to work.

To see why it doesn't, consider the case of 3 rows. Take $O = O_3 \circ O_2 \circ O_1$, consisting of three single-row MPOs, where in line with the usual notation action on a vector to the right is acting on a "top MPS" from below, and composition is stacking MPOs below each other. Since $O$ is Hermitian, we have $O_1^\dagger = O_3$ and $O_2^\dagger = O_2$.

In this case the trouble already kind of starts with what we call $\ket{\phi}$. What we usually do in MPSKit is capture the leading eigenvector of $O$ as a "multi-line" MPS $\ket{\phi} := (\ket{\phi_1}, \ket{\phi_2}, \ket{\phi_3})$ which we determine using a multi-line `leading_boundary` algorithm. At convergence, the actual right leading eigenvector of the full 3-row $O$ is simply $\ket{\phi_1}$, $O\ket{\phi_1} \approx \Lambda \phi_1$, so our expectation value when $O$ is Hermitian actually becomes

$$\langle E \rangle = \frac{\langle \phi_1 | O_E | \phi_1 \rangle}{\langle \phi_1 | O | \phi_1 \rangle},$$

The other "lines" $\ket{\phi_2}$ and $\ket{\phi_3}$ are really more just bookkeeping tools used in the actual algorithm.

The interdependence of the different lines is captured by the convergence condition of the `leading_boundary` run,

$$\begin{align}
O_1 \ket{\phi_1} &\approx \Lambda_1 \ket{\phi_2}\\
O_2 \ket{\phi_1} &\approx \Lambda_2 \ket{\phi_3}\\
O_3 \ket{\phi_3} &\approx \Lambda_3 \ket{\phi_1}\\
\end{align}$$

and $\Lambda = \Lambda_1 \cdot  \Lambda_2 \cdot  \Lambda_3$. It's exactly this kind of "directional" propagation condition (in particular for each column individually) that is checked in the multi-line `calc_galerkin`, and that underlies the current multi-line `expectation_value` implementation.

What is evaluated in the current multi-line `expectation_value` implementation are local contractions of the form
```
    ┌───────AC[i,j]───────┐
    │          │          │ 
    │          │          │
  GL[i,j]────O[i,j]────GR[i,j]  :=  λ[i, j]
    │          │          │
    │          │          │
    └───────ĀC[i+1,j]─────┘
```
which give some local "eigenvalues" $\lambda_{i,j}$ in such a way that at convergence $\Lambda = \prod_{i,j} \lambda_{i,j}$.

Now say we're looking at a $3 \times 3$ system, and we want to compute an expectation value where $E$ is inserted at site $(1, 2)$. If we straightforwardly made this as a multi-line infinite MPO $O_E$ by replacing the tensor $O_{2,2}$ with $E$, and we called `expectation_value(expectation_value(ϕ, O_E)`, we would end up with a factor like
```
    ┌───────AC[1,2]───────┐
    │          │          │ 
    │          │          │
  GL[1,2]──────E───────GR[1,2]
    │          │          │
    │          │          │
    └───────ĀC[2,2]───────┘
```
But this makes absolutely no sense, because since $E$ is inserted in the first row, we don't just get $\ket{\phi_2}$ when applying this first row.

What we actually need to do is propagate $\bra{\phi_1}$ from *below* through the third and second rows, $\bra{\phi_1} O_3 \circ O_2 = \bra{\phi_3}$, until we end up with the local contraction
```
    ┌───────AC[1,2]───────┐
    │          │          │ 
    │          │          │
  G̃L[1,2]──────E───────G̃R[1,2]
    │          │          │
    │          │          │
    └───────ĀC[3,2]───────┘
```
where $\tilde{G}_L$ and $\tilde{G}_R$ are left and right fixed points of the mixed $\langle\phi_3 | O_1 | \phi_1 \rangle$ transfer matrix.

So in this case, "the expectation value of $E$ at site (1, 2)" should return the ratio of 
```
    ┌───────AC[1,2]───────┐
    │          │          │ 
    │          │          │
  G̃L[1,2]──────E───────G̃R[1,2]
    │          │          │
    │          │          │
    └───────ĀC[3,2]───────┘
```
over
```
    ┌───────AC[1,2]───────┐
    │          │          │ 
    │          │          │
  G̃L[1,2]────O[1, 2]────G̃R[1,2]
    │          │          │
    │          │          │
    └───────ĀC[3,2]───────┘
```

And this just has nothing to do with our `expectation_value` implementation, which really assumes a directional propagation of an eigenvector through an MPO. But an eigenvector can't be propagated through a defect, which is precisely what a naive use of `expectation_value` will do.


### Bringing this back to the changes suggested here

If `expectation_value(expectation_value(ϕ, O)` is only an expectation value for one operator $O$, then I would argue it's not really fair to call this `expectation_value`. I would find it much more natural to reserve the method `expectation_value` for something as illustrated above, contracting around a local tensor inserted into the partition function from all directions.

However, this is much more than we need in the `leading_boundary` logging, where we really want to print the eigenvalue $\Lambda$ we're computing.

Taking all of this into account, calling the scalar obtained from the product of a top-to-bottom propagation of a boundary MPS something like `dominant_eigenvalue` seems more fair to me, even beyond pure semantics.

Several years of unresolved poured into one comment, but maybe this helps clear up the confusion @lkdvos @borisdevos.

src/states/multilinemps.jl
leburgel: This is actually valid I think. Even though we probably don't have any support currently, a multi-line window MPO certainly makes sense, so there's no conceptual reason why a multi-line window MPS should be disallowed at the type level.

src/utility/multiline.jl
leburgel: As a physical thing, I don't see how the inner product of two multi-line MPS actually means anything. A multi-line MPS is just a collection of MPSs that by themselves are just valid states. They are related, or at least they should be to be of any use, but I don't see any useful interpretation of their collective norm.

What was this method actually used for? Is this just to use these kinds of objects in linear and eigensolvers, pretending everything is just one bit "abstract" vector like we sometimes do?

src/algorithms/expval.jl
lkdvos: Thanks Lander, this definitely helps me quite a lot.

Trying to parse the discussion into some actionable things here:
- I absolutely agree that `expectation_value` is somewhat misleading as a name, and we should probably work on an interface for actually doing the thing that would give expectation values. I do think some of the primitives you want are already there though, mostly in the form of `dot(psi, O, phi)`, which does indeed require some effort in the multiline context to get the correct values out, but could be a good starting point. With that in mind, would it make sense to just remove the `expectation_value` methods in their current form for Multiline?
- Considering the `dominant_eigenvalue` function, is the main point here just to have something for the grassman `f` and the debug logging? Would it make sense to hold off on adding a new public function until we actually figure out a proper interface for these things? I just want to be careful introducing more names, that's all

src/utility/multiline.jl
leburgel: Yes, type domain forwarding is default, so line 139 can go.

src/algorithms/expval.jl
leburgel: I would be fine with removing the current Multiline expectation value and not exporting dominant_eigenvalue. The only uses I can thing of are indeed free energy and logging. But it would still be good to include this in the examples, since computing a free energy is something people definitely want to do.

src/operators/multilinempo.jl
```diff
@@ -1,50 +1,57 @@
 # MultilineMPO
 # ------------
+#TODO: add algorithm support for finite MPOs
+const _MPOs = Union{InfiniteMPO, FiniteMPO}
```
leburgel: I think this can also be just `AbstractMPO`, even though there's things we don't support in practice.

docs/src/examples/classic2d/1.hard-hexagon/index.md
```diff
@@ -41,7 +41,7 @@ end
 
 One way to study statistical mechanics in infinite systems with tensor networks is by approximating the dominant eigenvector of the transfer matrix by an MPS.
 This dominant eigenvector contains a lot of hidden information.
-For example, the free energy can be extracted by computing the expectation value of the mpo.
+For example, the partition function per site is the dominant eigenvalue of the transfer matrix, from which the free energy follows.
```
leburgel:
```suggestion
For example, the free energy density can be extracted by computing the dominant eigenvalue of the transfer matrix.
```
"partition function per site" sound a bit weird. I would say "partition function value density" to be more correct, but that sounds even weirder, so I think it's less confusing to just not say that. I think this should be clear regardless, no?

docs/src/examples/classic2d/1.hard-hexagon/index.md
```diff
@@ -52,7 +52,7 @@ V = virtual_space(D)
     ψ₀, mpo,
     VUMPS(; verbosity = 0, alg_eigsolve = MPSKit.Defaults.alg_eigsolve(; ishermitian = false))
 ) # use non-hermitian eigensolver
-F = real(expectation_value(ψ, mpo))
+F = real(dominant_eigenvalue(ψ, mpo))
```
leburgel: Well, if we're calling this the free energy, there should really be a `-log(...)`, no?

docs/src/man/algorithms.md
```diff
@@ -329,6 +328,14 @@ use GradientGrassmann.
 leading_boundary
 ```
 
+The associated eigenvalue, i.e. the partition function per unit cell, is obtained with [`dominant_eigenvalue`](@ref).
```
leburgel: I think I would slightly prefer  `leading_eigenvalue`, to be consistent with `leading_boundary`.

Also, here I would actually say "partition function density" instead of "partition function per unit cell". Or at the very least "partition function value per unit cell").

docs/src/man/algorithms.md
```diff
@@ -329,6 +328,14 @@ use GradientGrassmann.
...
+This is deliberately not `expectation_value`: the boundary contracts against a different line than the one the operator acts on, so it is not an overlap.
+For a single-line `InfiniteMPS`/`InfiniteMPO` pair the two coincide and `expectation_value` forwards accordingly.
```
leburgel: I somehow think mentioning this makes it even more confusing. Just saying that for a single-line MPO this is equivalent to `expectation_value` should be fine i think, I would just drop the first sentence.

docs/src/man/operators.md
```diff
@@ -274,3 +273,23 @@ a collection (direct sum) of spaces, one for each row/column.
 ```@example operators
 left_virtualspace(H_ising, 1), right_virtualspace(H_ising, 1), physicalspace(H_ising, 1)
 ```
+
+## MultilineMPO
+
+A [`MultilineMPO`](@ref) is a stack of MPO lines making up the rows of a two-dimensional tensor network.
```
leburgel:
```suggestion
A [`MultilineMPO`](@ref) is a stack of MPO lines making up the rows of the row-to-row transfer operator of a two-dimensional tensor network.
```
Otherwise this sounds very finite inherently.

docs/src/man/operators.md
```diff
@@ -274,3 +273,23 @@ a collection (direct sum) of spaces, one for each row/column.
...
+This is typically the row-to-row or column-to-column transfer matrix of a 2D classical partition function, or a boundary MPO in the context of PEPS.
```
leburgel: I think you can drop this given the edit on the line before, and just say something like
"this network can be a conventional 2D classical partition function, or an effective partition function for example corresponding to the norm of a PEPS."

"boundary MPO" isn't really a thing I think, I don't know what this means precisely.

docs/src/man/operators.md
```diff
@@ -274,3 +273,23 @@ a collection (direct sum) of spaces, one for each row/column.
...
+See the `MultilineMPS` section of the [states](@ref um_states) page for the row-shift convention.
+
+```@example operators
+mpo_multi = MultilineMPO([mpo, mpo])
+```
+
+The type restricts lines to `InfiniteMPO` and `FiniteMPO`, since a `MultilineMPO` is a statistical mechanical transfer operator.
+The quantity extracted from it is a [`dominant_eigenvalue`](@ref) rather than an energy.
```
leburgel: I think you can just drop this, if we make this clear above and from the example.

docs/src/man/operators.md
```diff
@@ -274,3 +273,23 @@ a collection (direct sum) of spaces, one for each row/column.
 ```@example operators
 left_virtualspace(H_ising, 1), right_virtualspace(H_ising, 1), physicalspace(H_ising, 1)
 ```
+
+## MultilineMPO
+
+A [`MultilineMPO`](@ref) is a stack of MPO lines making up the rows of a two-dimensional tensor network.
+This is typically the row-to-row or column-to-column transfer matrix of a 2D classical partition function, or a boundary MPO in the context of PEPS.
+See the `MultilineMPS` section of the [states](@ref um_states) page for the row-shift convention.
```
leburgel: Does "row-shift convention" have to be this specific? Can we just say "indexing convention"?

docs/src/man/states.md
```diff
@@ -155,20 +155,48 @@ A two-dimensional classical partition function can often be represented by an in
...
 ```@example states
-row = 2
-col = 2
-al = state.AL[row, col];
+al = state.AL[2, 1];
 ```
 
+### The row-shift convention
+
+Row `i` of a [`MultilineMPO`](@ref) maps row `i` of the network onto row `i + 1`.
```
leburgel: I think this section is great, but I'd have to look over the details when I have more time. For example here already, withing our `leading_boundary` algorithm we assume that row `i` of a `MultilineMPO` maps row `i` of the `MultilineMPS` to row `i + 1` of the `MultilineMPS`. But row row `i` of a `MultilineMPO just *is* row `i` of the network, so the sentence itself is a bit confusing.

I'll have a go at rewriting this a bit soon.

docs/src/man/states.md
```diff
@@ -155,20 +155,48 @@ A two-dimensional classical partition function can often be represented by an in
...
+Independently of that, the `ncols` sites of the unit cell contribute factors of their own on top.
+[`dominant_eigenvalue`](@ref) accumulates both.
+
+Applying an operator therefore acts on an ordinary [`InfiniteMPS`](@ref), advancing it by one full period:
+
+```julia
+O * ψ == O[end] * (… * (O[2] * (O[1] * ψ)))
+```
+
+### Subtleties
```
leburgel: Indexing and iteration ideally shouldn't involve subtleties, let's stick to a vector-like behavior.

src/algorithms/groundstate/vumps.jl
```diff
@@ -75,21 +75,24 @@ function dominant_eigsolve(
     state = VUMPSState(mps, operator, envs, iter, ϵ, which, timeroutput)
     it = IterativeSolver(alg, state)
 
+    # both `find_groundstate` and `leading_boundary` go here
+    objective = mps isa MultilineMPS ? dominant_eigenvalue : expectation_value
```
leburgel: To make this a bit cleaner, we can make the objective getter dispatch on the type of state and operator to avoid an explicit switch here?

PR #510, opened 2026-08-12
Allow passing `hasconverged` and `shouldstop` kwargs to `OptimKit.optimize` when using `GradientGrassmann`

leburgel:
## Description

Ground state optimizations using the default `GradientGrassmann` settings can run into serious trouble, where the line search returns a zero step size. This either gives some vague errors further down the line, or leads to the optimization continuing to the `maxiter` without ever moving.

Here I added the ability to set the `hasconverged` and `shouldstop` keyword arguments used by `OptimKit.optimize`, to be able to at least stop the optimization when these kinds of issues occur.

## Checklist

- [x] Runic formatter is run


lkdvos (commented):
Do we need to bump the minimal OptimKit version for this?

leburgel:
> Do we need to bump the minimal OptimKit version for this?

Didn't think of this, but yes: it seems this was introduced in OptimKit v0.4.

lkdvos (approved):

PR #511, opened 2026-08-12
Fix`TaylorCluster` for non-square (finite) Hamiltonians

VictorVanthilt:
## Description

The TaylorCluster algorithm's "loopback" algorithm involves removing the rows and columns that are labled by level-labels only containing 1's and "3"'s. Currently the rows and columns were removed using the right-virtualsize of the Hamiltonian as the "3" level.

To be correct for non-square Hamiltonians, we need to remove the columns with a "3" level based on the right-virtualsize and the rows based on the left-virtual size. This PR implements this.

## Checklist

- [x] Runic formatter is run


src/algorithms/timestep/taylorcluster.jl
```diff
@@ -130,21 +130,45 @@ function _taylor_loopback!(mpo, virtual_sz, linds, ::Val{N}, τ::Number) where {
...
-            end
+            slice[:, 1, 1, 1] = slice[:, 1, 1, 1] + factor * slice[:, 1, 1, c]
         end
     end
+    _remove_threelevels!(mpo, virtual_sz, linds)
     return mpo
 end
 
+function _get_threelevel_linds(virtual_sz, linds)
+    row_threelevels_linds = Int[]
```
lkdvos, line 147: Do we have to allocate this before running over it? Is it not easier to just inline this function into the loop?

lkdvos: [edit] I see you need this below

src/algorithms/timestep/taylorcluster.jl
```diff
@@ -130,21 +130,45 @@ function _taylor_loopback!(mpo, virtual_sz, linds, ::Val{N}, τ::Number) where {
...
+    for (i, slice) in enumerate(parent(mpo))
+        # remove the three-level rows
+        row_threelevel_linds = _get_threelevel_linds(virtual_sz[i], linds[i])
+
+        # remove the three-level columns
+        col_threelevel_linds = _get_threelevel_linds(virtual_sz[i + 1], linds[i + 1])
+
+        for I in nonzero_keys(slice)
+            (I[1] in row_threelevel_linds || I[4] in col_threelevel_linds) && delete!(slice, I)
+        end
```
lkdvos: can we check the properties that define the threelevel_linds instead, mapping it to a cartesian index first?

VictorVanthilt: I'm not quite sure I fully understood your comment but I changed the code to first make the set of CartesianIndices that fall in a to-be-deleted row/column. I then loop over the intersection of this set and the nonzero_keys. This avoids a lot of misses in the "for key in nonzero_keys" loop.

lmk if this is not what you meant

VictorVanthilt:
I don't understand this test failure, I can't see what failed from the logs.

lkdvos (approved):

PR #512, opened 2026-08-14
Overhaul on error reporting of algorithms

borisdevos:
[Relevant update deeper in conversation]

## Description

The main motivation started with `TDVP2` not reporting its truncation error. I noticed it could just make use of  `gauge2!` to remove duplication. While doing this, I realised multiple parts in the code didn't report their error, or didn't clarify clearly what the error actually means. In particular for the time evolution code which isn't variational, it made me realise that `ϵ` could mean anything. So this PR ended up expanding massively to also documenting per algorithm where relevant what the returned error represents.

Details of the changes are mentioned in the changelog, and motivation for the errors in the docstrings or documentation. Importantly:

- `timestep`/`timestep!`/`time_evolve`/`time_evolve!` now return `(ψ, envs, ϵ)`. `time_evolve` also logs more correctly. Tests added for this.
- Documentation of what every reported error actually means.

Something I noticed along the way with `changebonds` is that the meaning of its truncations differ too strongly to unify and justify returning the error. There's an argument to returning it for `(VUMPS)SvdCut` as the error there is genuinely a truncation error, but I didn't do that.

## Checklist

- [x] Runic formatter is run


docs/src/man/algorithms.md
```diff
@@ -263,6 +275,138 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
 
+## Errors and accuracy
+
+Most algorithms in MPSKit report an error alongside their result, and the manual pages above refer to it as `ϵ` throughout.
+That single name covers genuinely different quantities, and the differences matter.
+This section clarifies the differences, and explains what they do and, just as important, what they don't measure.
+
+### The error convention
+
+Where `ϵ` is a *truncation* error, it is the 2-norm of the discarded singular values of a single factorisation, so that `ϵ²` is the discarded weight and the squared norm of the factorised tensor drops by exactly `ϵ²`.
```
lkdvos: this is a bit of a strange wording, I'm not really sure there's a lot of benefit referring to the squared norm dropping by the squared epsilon. Is it not easier to just replace this entire paragraph with truncation error is `|Aexact - Atrunc|`? I'm also not entirely sure we are applying this to properly normalized reduced density matrices everywhere so while it makes sense to mention this, it seems a bit overstated to me.

docs/src/man/algorithms.md
```diff
@@ -263,6 +275,138 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
+
+!!! warning
+    Not every `ϵ` is a truncation error.
+    The `ϵ` returned by [`find_groundstate`](@ref), [`leading_boundary`](@ref) and the iterative [`approximate`](@ref) algorithms is a *convergence* measure, with no truncation interpretation at all.
+    The two are unrelated quantities that happen to share a name; see the two sections below.
+
+### Ground-state accuracy
+
+[`find_groundstate`](@ref), [`leading_boundary`](@ref) and the iterative [`approximate`](@ref) algorithms return the quantity their `tol` is compared against.
+For the sweeping algorithms ([`DMRG`](@ref), [`DMRG2`](@ref), [`VUMPS`](@ref), [`IDMRG`](@ref), [`IDMRG2`](@ref)) this is the Galerkin error: the norm of the local gradient projected orthogonally to the current state.
```
lkdvos: I don't think you can really project a gradient orthogonally to a state, so maybe easier is: "the norm of the gradient projected onto the local MPS manifold" or something similar?

docs/src/man/algorithms.md
```diff
@@ -263,6 +275,138 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
+It vanishes exactly at a variational fixed point.
+[`GradientGrassmann`](@ref) instead reports the norm of the Riemannian gradient from its optimizer.
```
lkdvos: I might be wrong here, but is that not the same gradient, up to second order things?

docs/src/man/algorithms.md
```diff
@@ -263,6 +275,138 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
+A small `ϵ` certifies a fixed point, not an accurate state.
+Growing the bond dimension is the job of the two-site algorithms ([`DMRG2`](@ref), [`IDMRG2`](@ref)) or of a bond expansion ([`DMRG`](@ref) with an `alg_expand`, or an expanding `alg_gauge` such as [`DMRG3S`](@ref)); see also [`changebonds`](@ref).
+
+Once an algorithm does truncate, the two error notions interact.
+The Galerkin error cannot fall below the level set by the weight being discarded each sweep, so a truncating scheme converges once `ϵ` reaches the truncation error rather than the (unreachable) bare `tol`.
+[`DMRG`](@ref)/[`DMRG2`](@ref) account for this: their stopping test is `ϵ ≤ max(tol, maximum(ϵ_trunc))`, which reduces to the plain `ϵ ≤ tol` when nothing is truncated.
+
+Neither measure is an error bar on an observable.
+For that, the standard route is the energy variance ``\langle H^2 \rangle - \langle H \rangle^2`` and extrapolation of observables towards zero variance ([Hubig et al.](@cite hubig2018)).
```
lkdvos: While I agree energy variance is a good independent error measure, that actually only tells you how far the state is from an exact eigenstate, which is somewhat unrelated to error bars on observables.

docs/src/man/algorithms.md
```diff
@@ -263,6 +275,138 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
+  Plain single-site [`TDVP`](@ref) runs at fixed bond dimension and returns exactly `0`.
+
+* **Projection error.**
+  Single-site [`TDVP`](@ref) confines the evolution to the tangent space of a fixed-bond-dimension manifold, ``\lVert (1 - P_{T_\psi}) H \psi \rVert``.
+  The component of the exact evolution pointing off that manifold is simply dropped, and this happens even with no truncation and exact local solves.
+  It is not reported as measuring it costs an extra effective-Hamiltonian application per site.
+  This is what a bond expansion (CBE) exists to reduce ([Li et al.](@cite li2024)).
+
+* **Time-discretization error.**
+  The projector splitting is globally ``O(dt^2)`` for the symmetric back-and-forth sweep ([Lubich et al.](@cite lubich2015), [Paeckel et al.](@cite paeckel2019)), so it is controlled by `dt` alone.
```
lkdvos: I'm not sure I fully agree with the explanation/title combination here, I'd say the thing you are describing is the trotterization error, rather than a time-discretization error, although I agree this might be potato-potato 🙃

docs/src/man/algorithms.md
```diff
@@ -263,6 +275,138 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
+  This can only be estimated by comparing one step of `dt` against two of `dt / 2`.
+
+A trustworthy run needs all three under control, not just a small `ϵ`.
```
lkdvos: It might also be nice to add the usual explanation here that there usually is a sweet spot between choosing `dt` too small and too large?

docs/src/man/algorithms.md
```diff
@@ -263,6 +275,138 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
+They are worth knowing about, because the dominant one is usually not the one the algorithm is working on.
+
+* **Inherited ground-state error.**
+  Every method here builds on a ground state you supply and treats it as exact.
+  Its error propagates straight into the gap, and since a gap is a difference of two large energies, it is typically the limiting factor.
+  A well-converged ground state (in the sense of `ϵ` and bond dimension) is necessary for a meaningful excitation calculation.
+
+* **Ansatz limitation.**
+  [`QuasiparticleAnsatz`](@ref) varies over the single-quasiparticle tangent space on top of a fixed ground state.
+  It is variational within that space and well suited to isolated quasiparticle branches, but multi-particle continua are not representable in it, so results there are not to be trusted.
```
lkdvos: Maybe want to quote here: there is a result somewhere that says you get convergenc based on the gap between ground state and excitation and between excitation and next excitation

docs/src/man/algorithms.md
```diff
@@ -263,6 +275,138 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
+  For infinite systems the momentum superposition itself is exact, so momentum is a good quantum number and no error enters through it.
```
lkdvos: I don't really know what this means?

src/algorithms/groundstate/dmrg.jl
```diff
@@ -171,7 +172,8 @@ $(TYPEDFIELDS)
 Used as the `algorithm` argument of [`find_groundstate`](@ref) and [`approximate`](@ref).
 """
 struct DMRG2{A, G, F, B} <: Algorithm
-    "tolerance for convergence criterium"
+    "convergence tolerance, compared against the Galerkin error (the tangent-space gradient
```
lkdvos: as well as the truncation error I think?

src/algorithms/timestep/bug.jl
```diff
@@ -171,23 +180,28 @@ function timestep!(
...
-        ψ = local_update!(
+        ψ, ϵ = local_update!(
             site, Val(:left), ψ, H, alg, envs, t + h, h, allocator;
             imaginary_evolution, normalize, timeroutput
         )
+        ϵ² += ϵ^2
     end
 
-    return ψ, envs
+    return ψ, envs, sqrt(ϵ²)
```
lkdvos: This is not very consistent with the other reports, which try to keep most of the quantities size-independent. (This is also why the convergence measures actually don't measure the total gradient norm, and rather take the maximum of the local gradient norms)

lkdvos (commented):
Left some comments throughout, but as a more general idea here, since this is breaking anyways:

I think the entire idea of `psi, envs, eps` is probably not great to begin with, precisely because it is hard (or important) to keep that to meaning the same in every part. If we are making breaking changes anyways, it might be a convenient time to just go to a more KrylovKit-related approach where we just return an `info` struct, where we are then actually free to return the different quantities, and name them appropriately. This both keeps the signature the same everywhere, without unwantedly promising meaning to the number.

Thanks for taking the time to properly document many of these things by the way, this is definitely a welcome addition. There are some subtleties about the prose not lining up with the theory or the implementation, since especially for the convergence measures being practical had higher priority than being rigorous, and it seems like the language kind of mixes between the two. I'm not sure if you wanted to describe the theory or the implementation?

docs/src/man/algorithms.md
borisdevos: Yeah the wording was sloppy on my part, but wouldn't it also be more correct to say you're projecting onto the complement of the manifold? I can also just adapt `calc_galerkin`s current docstring definition: "the overlap of the current state with the single-site derivative, projected onto the nullspace of the current state"

docs/src/man/algorithms.md
borisdevos: So it's better to say at first order it's the same gradient, but at different metrics, and the tolerance still doesn't transfer because they're summed up differently?

docs/src/man/algorithms.md
borisdevos: At least within TDVP I think it's potato-potato, but maybe it's worthwhile generalising this to a "splitting error" since the time MPO methods split the Hamiltonian itself, while TDVP splits the tangent space projector.

docs/src/man/algorithms.md
borisdevos: I tried explaining that QPA makes momentum eigenstates by construction, so there's no such thing as a momentum error in its quantisation. Finite-size you might expect an error of the order 2pi/L. But it's really more mentioning something you don't have to worry about, so I think I'll just remove it in this part.

src/algorithms/groundstate/dmrg.jl
borisdevos: Yes, I explained this correctly in the docs but not in this docstring

src/algorithms/timestep/bug.jl
borisdevos: Ah you're right, indeed how I'm adding squares here will make the error size-dependent. It's then more correct here to do something like in the DMRG code? 

Related, but I think this will also be fixed if I do some convergence info struct like what you suggested above, because then we don't have to worry about how things should or shouldn't be summed up

borisdevos:
Thanks for the review! I agree I was inconsistent in separating theory vs what's effectively done in the code. I think it's beneficial to give both, so I'll see how far I get in that. I'm also in favor of the info struct, so I'll try that out!

borisdevos:
Okay, a bunch has happened, and the goal of this PR has completely shifted, but the changes are better, and it's good that they're done at once (or at least shown here bunched up, there's an argument to splitting up some parts).

I introduced `AlgorithmInfo` as the info struct, done in such a way that every algorithm that can return an error does it via this one struct. Along the way I actually found more algorithms than TDVP2 which calculated truncation errors, but didn't return them. So I think I found them all now. I think the interface is clean, but do let me know if anything's weird about it.

I expanded on the docs a bunch more as well, correcting some of the false statements I made along the way. I think it reads more clearly now what a user could expect from these errors versus what they actually get. I think I'm less wrong than last time, but there might still be mistakes 🙃

src/utility/algorithminfo.jl
```diff
@@ -0,0 +1,139 @@
...
+  - `numtrunc`: how many local factorisations actually discarded anything.
+
+See [Aggregating truncation errors](@ref) for how the two relate to a `trunc` setting, and for the
+per-strategy caveats.
+
+These are `0` for an algorithm that never truncates, which does not mean the result is exact. Rather,
+it means this particular error channel is absent. See the manual on [Errors and accuracy](@ref)
+for what is *not* measured here.
+"""
+struct AlgorithmInfo{T <: Real}
```
lkdvos: Is there a benefit to using this struct vs say a `NamedTuple`? I'm a bit wary that not all algorithms have access to the same information, and therefore might not want to report the same information. I do like having some standardization with what fields/names mean what properties, but I am a bit scared about this not being very extensible.

For example, one other way to do this would be to have something like:

```julia
struct AlgorithmInfo
    fields::Dict{Symbol, Any}
end
```
(+ presumably some syntactic sugar to directly access fields etc)

This basically means that all algorithms are "type-stable" in the sense that they can return any information they want without altering the output type, but of course still needs a dynamic dispatch for actually accessing the fields itself, which I'm assuming here to never be performance-critical.

src/utility/algorithminfo.jl
```diff
@@ -0,0 +1,139 @@
...
+    converged::Union{Bool, Nothing}
+    normres::Union{T, Nothing}
+    ϵ_max::T
+    ϵ_total::T
```
lkdvos: In general I try to avoid unicode in propertynames, or basically anywhere where it is basically a forced choice, as there are still some parts where typing unicode is not the most convenient and having an ascii alternative is nice.

src/utility/algorithminfo.jl
borisdevos: I think what you're suggesting is to trade "same name, different meanings" with the opposite, right? The bad part with the former which I do here and tried to fix with elaborate docstrings is that e.g. `normres` does in fact mean something else depending on the algorithm. The opposite is having some return `:galerkin` and others `:residual` or something along those lines, which to me at least at the level of `AlgorithmInfo` makes it annoying to get to the information. I guess that's what you mean by sugar then accounting for this?

The struct as-is prints also nicely, which is to be considered, and it gives you the information immediately without having to call extra functions. All the fields are known à priori, the only thing is interpreting it, which again leads to the docstrings/docs being elaborate. It's indeed not extensible, but the current fields cover a lot of ground on what an algorithm would want to return, no?

lkdvos: I think indeed having some return `:galerkin` and others `:gradientnorm` or anything like that, where names are tied to meanings, is exactly what I would expect. This still means that the relevant docstrings have to document what these mean, but I do think that gives us a lot more flexibility to actually return diagnostic data.

I guess I don't really agree with "the fields are known a priori" in the sense that you specifically have to check which ones are `nothing`, which is literally the same as having to check which fields are there. The only thing I'm suggesting here is to just make this fully dynamic from the beginning, only filling in "fields" that are present, which can easily be queried with `keys`, and additionally being trivially extensible (e.g. we could if we wanted to return iteration count, function application count, ... in a non-breaking way).

Having it print nicely is not really a great argument, it really isn't that hard to write the show method such that it prints the exact same information here as well 😉.

borisdevos: If you put it that way, then I must agree with you 😄 but I see your point now, and this flexibility is definitely sturdier to future changes. I'll get to overhauling the overhaul

(My print argument was more when thinking about named tuples, because that would be type piracy to pretty print that, no?)

borisdevos:
Alright, the new `AlgorithmInfo` with a `Dict` field is now available! The most important related change is that the old `info.normres` is now replaced by certain "convergence keys", which I've managed to fix to 4 that occur in the current algorithms. Of course, this can be expanded, and it should be such that the new `convergence_measure` function can call it (as well as for pretty printing). I think the names I chose there make sense, the only one I have some doubt for is `:localchange` for the finite VOMPS code.

Concerning the unicode, I decided to keep them as aliases, but the ASCII versions are promoted everywhere, and it's also what you see when you show/display the info. The aliases are mentioned briefly in the docs and docstring of `AlgorithmInfo`.

Just to give an idea of how this displays, an example for `DMRG2`:
```julia
julia> eps
AlgorithmInfo:
  converged              = true after 6 iterations
  galerkin               = 6.417388341837349e-7
  max_truncation_error   = 1.2567011712421484e-6        (largest single factorisation)
  total_truncation_error = 1.6092084215067107e-6        (quadrature over 32 truncations)
```

I forgot to mention last time, but I haven't regenerated the docs. This should be done before merging.

docs/src/man/algorithms.md
```diff
@@ -263,6 +275,251 @@ Es, ϕs = excitations(H, ChepigaAnsatz2(), ψ, envs; num=1)
...
+[`excitations`](@ref) and [`changebonds`](@ref) report nothing.
+What limits their accuracy is covered below all the same.
+A single bare number could not do this job, because the quantities involved are genuinely different.
+A convergence measure and a truncation error answer different questions, are not comparable, and not every algorithm produces both.
+Conversely, two algorithms that both report "a convergence measure" do not necessarily report the same measure.
+[`AlgorithmInfo`](@ref) therefore carries a set of named entries, and each algorithm fills in only the ones it actually computes.
+Entries are keyed by `Symbol`, and read either as a property (`info.max_truncation_error`) or by indexing (`info[:max_truncation_error]`).
+The usual dictionary interface is supported, in particular `haskey`, `keys`, `values` and `get`.
+Nothing is promised that was never measured: asking for an entry an algorithm did not report is an error naming what it did report, rather than a value that was never computed, and the name of the entry says which quantity it is rather than leaving that to prose.
```
lkdvos: This sentence is a bit hard to parse, and this paragraph reads a bit hard as well. I think this is really just saying that additional information can be retrieved in the `info` struct?

PR #514, opened 2026-08-17
Fix `changebonds` for `FiniteMPS` with composite physical spaces

### LINKED ISSUE #494: SketchedExpand doesn't work on FiniteMPS with a composite physical space (multiple site indices)
Hi all, thanks for all the great work on this package!

I'm building an LPTN library on top of MPSKit, and while testing time evolution with `SketchedExpand` I ran into an error. Here's a minimal example:

```julia
using MPSKit, TensorKit
include("test/setup/testsetup.jl")
using .TestSetup

L = 10
P = ℂ^2
Dummy = ℂ^2
Dstart = 2
Dcap = 16
dt = 0.05

H = heisenberg_XXX(ComplexF64, Trivial; spin = 1 // 2, L)

# site tensors are `GenericMPSTensor{ComplexSpace,3}`: (Vl, P, Dummy) <- Vr
ψ₀ = FiniteMPS(rand, ComplexF64, L, P ⊗ Dummy, ℂ^Dstart)
alg = TDVP(;
    alg_expand = SketchedExpand(; trunc = truncrank(Dstart), oversampling = 4),
    trunc = truncrank(Dcap)
)
for _ in 1:6
    timestep!(ψ₀, H, 0.0, dt, alg)
end
```

Error message:
```
ERROR: SpaceMismatch: Direct sum of a vector space and its dual does not exist
Stacktrace:
 [1] ⊕
   @ ~/.julia/packages/TensorKit/zSHBV/src/spaces/complexspace.jl:57 [inlined]
 [2] changebond!(site::Int64, ::Val{…}, ψ::FiniteMPS{…}, H::FiniteMPOHamiltonian{…}, alg::SketchedExpand{…}, envs::MPSKit.FiniteEnvironments{…}; normalize::Bool)
   @ MPSKit ~/MPSKit.jl/src/algorithms/changebonds/sketchedexpand.jl:123
 [3] changebond!
   @ ~/MPSKit.jl/src/algorithms/changebonds/sketchedexpand.jl:92 [inlined]
 [4] timestep!(ψ::FiniteMPS{…}, H::FiniteMPOHamiltonian{…}, t::Float64, dt::Float64, alg::TDVP{…}, envs::MPSKit.FiniteEnvironments{…}; imaginary_evolution::Bool)
   @ MPSKit ~/MPSKit.jl/src/algorithms/timestep/tdvp.jl:156
 [5] timestep!
   @ ~/MPSKit.jl/src/algorithms/timestep/tdvp.jl:120 [inlined]
 [6] timestep!(ψ::FiniteMPS{…}, H::FiniteMPOHamiltonian{…}, t::Float64, dt::Float64, alg::TDVP{…})
   @ MPSKit ~/MPSKit.jl/src/algorithms/timestep/tdvp.jl:120
 [7] top-level scope
   @ ~/MPSKit.jl/trial.jl:22
Some type information was truncated. Use `show(err)` to see complete types.
```

I think the issue is here: https://github.com/QuantumKitHub/MPSKit.jl/blob/0ff11c35da63d33f177eb2846b84503e3cab2bd8/src/algorithms/changebonds/sketchedexpand.jl#L123

`space(Q, 3)'` looks like it assumes `Q` has exactly one physical-like leg, so the flat index `3` happens to land on its new-bond-direction space only when there's a single physical leg. For a state whose site tensors carry more than one physical-like leg (as in my case, `GenericMPSTensor{S,3}`), index `3` lands somewhere else instead. Since `Q` always has exactly one domain leg regardless of how many codomain legs it has, I believe `domain(Q)[1]` is the leg-count-agnostic fix.

I tried this change locally and it resolves the error for me. Happy to open a PR with it if that would help!

leburgel:
## Description

Fixes #494, using the fix @lkdvos suggested there.

I added tests using the one-sided MPO-MPS application that was added for finite temperature support. Unfortunately, these don't pass due to an issue with the corresponding contractions for space types with non-trivial braiding. Changing the space type to `ComplexSpace` fixes things, but it's probably better to keep using `PlanarTrivial` in the tests`.

I reported the issue in QuantumKitHub/BlockTensorKit.jl#68.

## Checklist

- [x] Runic formatter is run

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -37,7 +37,7 @@ MPSKitAdaptExt = "Adapt"
 [compat]
 Accessors = "0.1"
 Adapt = "4"
-BlockTensorKit = "0.3.15"
+BlockTensorKit = "0.3.16"
 Compat = "3.47, 4.10"
 DocStringExtensions = "0.9.3"
 HalfIntegers = "1.6.0"
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -97,6 +97,9 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
   tensor, so it returned `O * O` on twice the physical space instead of `O`.
   ([#484](https://github.com/QuantumKitHub/MPSKit.jl/pull/484))
 - `leading_boundary` with `IDMRG2` didn't update the left edge of the AC tensor, which could result in space mismatches depending on the truncation scheme. This is corrected for in ([#516](https://github.com/QuantumKitHub/MPSKit.jl/pull/516)).
+- Fix hardcoding of number of physical spaces in the `changebonds` implementations for
+  `FiniteMPS`, enabling its use for systems with composite physical spaces
+  ([#514](https://github.com/QuantumKitHub/MPSKit.jl/pull/514))
 
 ### Performance
 
--- a/src/algorithms/changebonds/optimalexpand.jl
+++ b/src/algorithms/changebonds/optimalexpand.jl
@@ -179,7 +179,7 @@ function changebond!(
     right_tail = _transpose_tail(right)
     # embed `_transpose_tail(right)` into the enlarged codomain (zero weight in the new
     # directions), leaving the state unchanged
-    nc_space = (codomain(right_tail)[1] ⊕ space(Q, 3)') ← domain(right_tail)
+    nc_space = (codomain(right_tail)[1] ⊕ _lastspace(Q)') ← domain(right_tail)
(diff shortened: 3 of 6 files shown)


lkdvos (approved):

leburgel:
It seems some inference checks in the excitation tests broke. These seemed to pass on the previous commit here, do you think the loss of inference could be related to the BlockTensorKit updates @lkdvos?

lkdvos:
It must be, I'll try and investigate that separately. In any case, this is not related here because I'm seeing these failures on other PRs as well, so will merge this one.

PR #516, opened 2026-08-18
[fix] Update edge tensors accordingly in leading boundary with IDMRG2

borisdevos:
## Description

Leading boundary with IDMRG2 forgot to update one of the tensors in the right-to-left sweep, which caused space mismatches.
The test I added could've been put in the old bugs section, but where I put it has most of the code already compiled so it's less straining for CI. 

This was not caught before I guess because in other cases the rank just saturates immediately, so pre-updated data in the sweeps don't space mismatch.

## Checklist

- [x] Runic formatter is run


lkdvos (approved):
Thanks!

